PR #10, opened 2020-07-31
style: typography and skeleton of the app

goldsteinsveta:
as per [design reference](https://www.notion.so/paritytechnologies/Design-Reference-WIP-8fdbc2e048f743fc9686a6a5129bdad5)


projects/burnr/src/pages/Home.tsx
```diff
@@ -1,40 +1,55 @@
...
-interface Props {
-  className?: string;
-}
+function Home ():  React.ReactElement {
 
-function Home ({ className }: Props):  React.ReactElement<Props> {
-	const blockHash = useChainInfo();
+	return (
+		<>
+			<Grid container alignItems='center'>
```
Stefie, line 12: I'd put this part (`<LogoApp>` & `<NodeSelector>`) in the `<ThemeHeader>`, because they're both relevant or all pages (even though we just have a single one for now).

It possibly makes sense, to import all 4 components in the `<ThemeHeader>` component

So in ThemeHeader.tsx:
```
<div className={classes.root}>	
	<LogoSubstrate theme={theme} />
        <LogoApp>
        <NodeSelector>
	<ThemeButton theme={theme} onClick={() => setTheme(!theme)} />
</div>
```

Also: The header also  looks a little broken on smaller screens right now, the elements are overlapping.
Screenshot 2020-08-03 at 11 00 08

projects/burnr/src/pages/Home.tsx
```diff
@@ -1,40 +1,55 @@
...
+							<IconButton>
+								<VisibilityIcon />
+							</IconButton>
+						</Grid>
+					</Grid>
+				</Box>
+			</Paper>
+			<Divider/>
+			<NavTabs />
+			<NavFooter />
```
Stefie, line 49: `<NavFooter>` is also a general App element that shouldn't be specific for single pages. It also belongs in the App.tsx. I'd suggest the following structure:
```
<ThemeProvider theme={appliedTheme}>
	<CssBaseline />
	<ThemeHeader />
	<main className={classes.main}>
		{api && (
			<ApiContext.Provider value={api}>
				<div className='toolbar' />
				<Switch>
					<Route exact path='/' component={Home} />
				</Switch>
			</ApiContext.Provider>
		)}
	</main>
        <NavFooter />
</ThemeProvider>
```

projects/burnr/src/components/ThemeHeader.tsx
```diff
@@ -4,12 +4,19 @@ import { makeStyles } from '@material-ui/core';
 
 const useStyles = makeStyles(theme => ({
 	root: {
-		alignItems: 'center',
+		position: 'fixed',
+		zIndex: -1,
```
Stefie, line 8: That moves the header below the `html`. `body` etc. elements and makes it unclickable. You can't switch the theme right now. It shouldn't be needed

projects/burnr/src/components/NavTabs.tsx
```diff
@@ -0,0 +1,82 @@
...
+import CallMadeSharpIcon from '@material-ui/icons/CallMadeSharp';
+import WhatshotSharpIcon from '@material-ui/icons/WhatshotSharp';
+
+interface TabPanelProps {
+  children?: React.ReactNode;
+  index: number;
+  value: number;
+}
+
+const TabPanel: React.FunctionComponent<TabPanelProps> = ({ children, value, index, ...rest }: TabPanelProps) => {
```
Stefie, line 14: Instead of `...rest`, just using  `...props` is common practice in React

PR #33, opened 2021-01-14
Resurrect burnr

raoulmillais:
This PR updates a bunch of dependencies and fixes a few minor issues causing the burnr wallet not to run when running `yarn run dev`.  The most noteworthy changes are:

* Upgrade to webpack 5 (and update associated webpack things) and manually shim node built-ins (buffer / stream / crypto) as these are no longer auto-shimmed by webpack.
* Upgrade @polkadot/rpc-provider and fix code to account for name changes
* Bugfix to make wallet render when your local storage is empty


Stefie (approved):
Awesome! Thanks for updating the dependencies!

PR #35, opened 2021-01-22
Smoldot provider

raoulmillais:
Here's my implementation of the PolkadotJS `ProviderInterface` for the [smoldot](https://github.com/paritytech/smoldot) WASM node.  Currently none of the live nodes listen to websockets with certificates so this won't work in the browser but it works in nodejs with no local node running.

### Implementation notes
 
* `ProviderInterface` abstraction is leaky - it includes connection / disconnection which doesn't really apply to a light client
* Very similar to the `WsProvider `without the websocket connection management (obviously).
* Much of the logic is duplicated with `WsProvider` (the internal types are generally the same but with the 'Ws' prefix removed) if we upstream we should factor out the common bits used by these 2 providers.
* types for smoldot should be upstreamed to smoldot.
* I didn't want to learn and bring in all the PolkadotJS test infrastructure until we're sure we want to upstream this.  I used [ava](https://github.com/avajs/ava) instead as it has good typescript support, runs the tests fast (in parallel) and I know it

### Unresolved questions

What should we do when the WASM client crashes?  Should / can we recover from it?


packages/smoldot-provider/src/SmoldotProvider.ts
```diff
@@ -0,0 +1,352 @@
...
+          l.debug('saving database', database_content);
+          this.#db.save(database_content);
+        }
+      })
+      .then((client: smoldot.SmoldotClient) => {
+        this.#client = client;
+        this.#isConnected = true;
+        this.emit('connected');
+      })
+      .catch((error: Error) => {
```
raoulmillais, line 203: oops this should reject the promise

packages/smoldot-provider/src/SmoldotProvider.ts
```diff
@@ -0,0 +1,352 @@
...
+  ): () => void {
+    this.#eventemitter.on(type, sub);
+
+    return (): void => {
+      this.#eventemitter.removeListener(type, sub);
+    };
+  }
+
+  /**
+   * @summary Send JSON data using WebSockets to the wasm node.
```
raoulmillais, line 247: copy - pasted ... shouldnt talk about websockets

packages/smoldot-provider/src/SmoldotProvider.ts
```diff
@@ -0,0 +1,352 @@
...
+   * @param method The RPC methods to execute
+   * @param params Encoded paramaters as appliucable for the method
```
raoulmillais, line 249: typo

tomaka:
> What should we do when the WASM client crashes? Should / can we recover from it?

You can theoretically call `smoldot.start()` multiple times. If a previous instance crashes, recreate one.

However `smoldot` is supposed to never crash, and I'd like to avoid situations where `smoldot` crashes and nobody notices the crashes because it transparently continues to work. If nobody notices crashes, they also won't get fixed.

packages/smoldot-provider/src/SmoldotProvider.ts
```diff
@@ -0,0 +1,352 @@
...
+          subscription
+        };
+
+      // Should we really catch errors here?  What can we do if the wasm client
+      // throws? Are we in an inconsistent state?
+      try {
+        this.#client.send_json_rpc(json);
+      } catch (error) {
+        reject(error);
+      }
```
tomaka, line 286: It's not supposed to throw except under catastrophic situations. I'd suggest to simply not deal with the error.

Stefie (commented):
NIce, that's awesome! 
Medium term it probably really makes sense to move it to or integrate it in to the Polkadot JS repo  as one possible Provider type, but as long as it's active development it's 100% fine here.

packages/smoldot-provider/src/SmoldotProvider.ts
```diff
@@ -0,0 +1,346 @@
...
+  // reference to the smoldot module so we can defer loading the wasm client
+  // until connect is called
+  #smoldot: smoldot.Smoldot;
+  readonly #handlers: Record<string, RpcStateAwaiting> = {};
+  #subscriptions: Record<string, StateSubscription> = {};
+  readonly #waitingForId: Record<string, JsonRpcResponse> = {};
+
+   /**
+   * @param {string}   chainSpec  The chainSpec for the WASM client
+   * @param {Database} db         An implementation of Database for saving the chain state
```
Stefie, line 80: Is this necessary for the Provider to work or just a node specific thing?

raoulmillais: It's necessary.  I've provided a `Database` implementation that just saves the JSON to disk.  When we're ready to test I'll write one that saves it to localStorage

tomaka: Databases are also something completely optional, for what it's worth.
We intend to make syncing very fast even if you don't have any existing database. Having a database might save some time, but it's clear how much. If it turns out that there's no difference between database vs no database, we could even remove databases entirely.

Stefie: If it stays it would probably just make sense to find a cross browser/ Node.js solution. 
@raoulmillais  Just out of curiosity: Have you tested in a browser environment or just in node so far?

Stefie (approved):

packages/smoldot-provider/src/SmoldotProvider.ts
raoulmillais: I haven't tried in the browser actually but I'm assuming that I will just get the same websocket security error you saw when trying the polkadot browser demo based provider you wrote.  I agree a cross platform solution would be nice and it'd be easy to do.  I introduced the interface to make typescript happy and so I could inject a dummy do-nothing test database in my unit tests.

PR #36, opened 2021-01-25
Prepare smoldot for browser

raoulmillais:
The biggest change here is the implementation of the database for the browser using local storage and autodetecting which to use.  I also:
* Changed where the fs database saves the json file to avoid collisions and be more sensible: `$PWD/${CHAIN_NAME:-smoldot-provider}.json`
* tidied up the unit test mocks 
* made the unit tests more consistent
* added a test for out of order messages
* added documentation for all the database types
* tweaked the typescript configuration and scripts for compatibility with the browser


packages/smoldot-provider/src/Database.ts
```diff
@@ -1,4 +1,53 @@
-export default interface Database {
+import * as pkg from '../package.json';
+
+// REM: Don't know how to make typescript happy. This doesn't work:
+// let create: (name: string) => Database | undefined = undefined;
+let create: any = undefined;
+
+if (typeof window === 'object') {
```
Stefie, line 7: Shouldn't this be wrapped in an async function? I'm surprised await is working without async here

packages/smoldot-provider/src/Database.ts
```diff
@@ -1,4 +1,53 @@
...
     save: (state: string) => void;
+  /**
+   * @description Delete any saved chain state
+   */
     delete: () => void;
-}
\ No newline at end of file
+}
+
+// REM: what if the database format changes?  Should we include a version?
```
Stefie, line 30: Yes, I was thinking the same! Maybe people even want to keep 2 versions of the DB of the same chain, for example one for development and one for live interactions and overwriting it every time would be a waste of resources.

Stefie (approved):

packages/smoldot-provider/src/Database.ts
raoulmillais: Yep I think we should discuss usage of databases with @tomaka a little more.  There's also no way to give the smoldot wasm client an existing database when you load it. So it doesn't seem to be providing any value at all at the moment (unless we provide an API for querying it - which might be a good idea?)

tomaka: Smoldot already has an embedded versions systems for its database. If the database format changes, it already handles maintaining backwards compatibility.

> There's also no way to give the smoldot wasm client an existing database when you load it.

[There is](https://github.com/paritytech/smoldot/blob/52169120e612b03cc7e9d7a25323bf214bab90bf/bin/wasm-node/javascript/index.js#L27) a way to pass the database content on startup. [Example](https://github.com/paritytech/smoldot/blob/52169120e612b03cc7e9d7a25323bf214bab90bf/bin/wasm-node/javascript/demo.js#L38).

> overwriting it every time would be a waste of resources

The database is really a 5 to 10kiB string, which is small enough that this shouldn't be a concern at all.
It's called "the database" but it's really a small JSON file. 50 to 200 times smaller than the chain specs, for instance.

Stefie: > It's called "the database" but it's really a small JSON file
Ah, ok, yeah, that really doesn't make a difference

tomaka: To add to this, it's not really a database but a snapshot.
Smoldot wants to save this JSON file so that it's restored when it is restarted.
It will never need to access the JSON file while it is running.

raoulmillais: >There is a way to pass the database content on startup. Example.

Thanks.  I missed that I will create a new pull request to load any existing database and pass in through to the client

>Smoldot already has an embedded versions systems for its database. If the database format changes, it already handles maintaining backwards compatibility.

Ok so I think it makes sense for the user to keep it so the user chooses a new unique name whenever they want to have a separate database I.E. multiple chain clients per page, but I haven't got as far as testing that, I just wrote the code to support it when we get there

PR #37, opened 2021-01-25
Smoldot browser demo

raoulmillais:
Simple browser demo using PolkadotJS and the new @substrate/smoldot-provider - as discussed let's make it look pretty and do a bit more before publishing it to IPFS and gh pages.


projects/smoldot-browser-demo/src/index.ts
```diff
@@ -0,0 +1,61 @@
...
+    displayMessage(messageHtml(message));
+  }
+
+  (async () => {
+    const response =  await fetch('./assets/westend.json')
+    if (!response.ok) {
+      error(new Error('Error downloading chain spec'));
+    }
+
+    const chainSpec =  await response.text();
```
Stefie, line 47: Minor nitpick: I don't think this needs an await, because `response` is already loaded async

Stefie (approved):

projects/smoldot-browser-demo/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "@substrate/smoldot-browser-demo",
+  "version": "1.0.0",
+  "main": "./dist/index.js",
+  "author": "Raoul Millais",
```
Stefie, line 5: And another small thing: Since it's open source, a Parity project and it will most likely have more contributors than one I'd generally prefer this syntax
```
  "contributors": [
    "Raoul Millais<raoul@parity.io>"
  ],
```

projects/smoldot-browser-demo/src/index.ts
raoulmillais: Nope, I thought that too: https://developer.mozilla.org/en-US/docs/Web/API/Body/text

projects/smoldot-browser-demo/package.json
raoulmillais: 100%. This is just the syntax that comes out of yarn init

PR #43, opened 2021-01-27
Check for existing state and pass it to the smoldot client

raoulmillais:
This adds functionality to check if an existing database exists (either saved to disk in node or in the local storage in the browser) and provides it to the smoldot client when `connect` is called on the `SmoldotProvider`.  There are no user facing changes, it should just now do the right thing.


Stefie (approved):
Looks good!

PR #50, opened 2021-01-29
Migrate from Webpack to Parcel & clean up dependencies

Stefie:


Stefie:
Damn, hang on, I deleted the assets... :|

PR #52, opened 2021-02-01
Add type checks to browser demo

Stefie:


Stefie:
I added type checks to the browser demo and there were 3 smaller linter errors. I fixed two, the third is more of an logical error where I don't know the intention. Could you have a look at this one yourself? 
Screenshot 2021-02-01 at 13 02 10

raoulmillais:
> I added type checks to the browser demo and there were 3 smaller linter errors. I fixed two, the third is more of a logical error where I don't know the intention. Could you have a look at this one yourself? You're declaring error as type `Error` in the class but passing a bool
> 
> Screenshot 2021-02-01 at 13 02 10

I swear I fixed this a while ago.   Not sure where that ended up, maybe it was on another branch

raoulmillais:
The validator config in `.parcelrc` https://github.com/paritytech/substrate-connect/blob/master/projects/smoldot-browser-demo/.parcelrc#L6-L8 was *supposed* to make parcel do the typescript validation so you don't have to call `tsc --noEmit`.  I think maybe that only works with parcel@next :disappointed:

Stefie:
> The validator config in `.parcelrc` https://github.com/paritytech/substrate-connect/blob/master/projects/smoldot-browser-demo/.parcelrc#L6-L8 was _supposed_ to make parcel do the typescript validation so you don't have to call `tsc --noEmit`. I think maybe that only works with parcel@next 😞

Ah, nice!
I need to look into all the config options for parcel. I didn't dig into that yet.

raoulmillais:
> > The validator config in `.parcelrc` https://github.com/paritytech/substrate-connect/blob/master/projects/smoldot-browser-demo/.parcelrc#L6-L8 was _supposed_ to make parcel do the typescript validation so you don't have to call `tsc --noEmit`. I think maybe that only works with parcel@next disappointed
> 
> Ah, nice!
> I need to look into all the config options for parcel. I didn't dig into that yet.

Yeah unfortunately it seems those config options only work for Parcel2 (@next) which we aren't using.  I tried out parcel@next in my exploration over the weekend and I got it all working but there is an open bug where the build fails unless you specify the `--no-cache`.  We could upgrade but I think it's better to wait for it to be a bit more stable.

Stefie:
> We could upgrade but I think it's better to wait for it to be a bit more stable

Yeah, agreed. Especially because we're working with enough unstable components from our side so it makes sense to reduce the potential number of bug-sources and stick to the more stable version of external dependencies!

Stefie:
Shall we merge this?

PR #61, opened 2021-02-03
add on the fly accounts to localStorage based on connected network

wirednkod:


projects/burnr/src/Home.tsx
```diff
@@ -30,11 +62,11 @@ function Home ():  React.ReactElement {
 					<Grid container alignItems='center' spacing={1}>
 						<Grid item xs={6}>
 							{
-								userInfo.address &&
+								userInfo?.address &&
 								<AccountCard
 									account={{
 										address: userInfo.address,
-										name: 'account name',
+										name: user.name
```
Stefie, line 69: I'm getting a TypeScript error here when I'm compiling it.

`src/Home.tsx:69:17 - error TS2532: Object is possibly 'undefined'.`

```suggestion
										name: user?.name
```

Stefie (approved):
Works like charm apart from the small TS linting error

goldsteinsveta (approved):

projects/burnr/src/Home.tsx
wirednkod: Yes - I want to have a better  look on this (as I really do not like how I did it so far) - its on my list

PR #70, opened 2021-02-03
Update @polkadot dependencies & yarn lock

Stefie:


Stefie:
@raoulmillais 
What exactly do you mean by "which will not result in a compile failure of smoldot-provider but will cause bundling to break." in your comment here? https://github.com/paritytech/substrate-connect/pull/47#issuecomment-771867491

It builds & runs without problems for me.

Issue #60, opened 2021-02-02
SmoldotProvider - Make events behave like other providers


raoulmillais:
Consumers of the `SmoldotProvider` have no way of knowing that we failed to start syncing from bootnodes because eg. we are offline.  We need to simulate as closely as sensible what the `WsProvider` does in this case  so that the consumer can chose to show an offline UI.  

These `connected` `disconnected` and `error` events are [propagated](https://github.com/polkadot-js/api/blob/master/packages/api/src/base/Init.ts#L65-L67) to the `Api` instance so we want the following to work as expected:

```
const provider = new SmoldotProvider(chainSpec());
const api = await ApiPromise.create({ provider });
api.on('error', () => /* do something */);
api.on('connected', () => /* do something */);
api.on('disconnected', () => /* do something */);
```
### Behaviour with WsProvider 

Consumers of `WsProvider` listen to the 3 events: `error`, `connected` and `disconnected` : 
* If `WsProvider` fails to connect to its remote node it only emits `error`.  Consumers can optionally `connectWithRetry`
* If it connects without error it emits `connected`. 
* If later the socket is closed it emits `disconnected`.  

(It also has logic for trying other remote nodes but that's not relevant for a smoldot client.)

### Approach

Set an interval pinger after instantiating the smoldot client to send an RPC `system_health` message to check the `peers` count. 

### Desired behaviour

* Don't emit `connected` until we see `peers > 0` from `system_health` pinger
* Emit `disconnected` if peer count from `system_health` drops back to zero after having previously been `> 0`

* Emit `error` if the `system_health` request responds with `peers == 0` right after after instantiating the smoldot client (analagous to error connecting on startup)
* Emit `error` if the `system_health` request responds with an error.
* Do something similar to `connectWithRetry` on the `WsProvider` https://github.com/polkadot-js/api/blob/master/packages/rpc-provider/src/ws/index.ts#L188-L201 but rather than acutally `reconnecting` we delay reporting the `error` event until a threshold time period is elapsed.  

**Aside**

The Api has its own pinger for `system_health` [used to keep the websocket alive](https://github.com/polkadot-js/api/blob/master/packages/api/src/base/Init.ts#L332-L342).  Unfortunately we cant get at the responses it receives from the provider so we have to redundantly have a second pinger in the provider.

 See https://github.com/paritytech/smoldot/issues/435 for a longer conversation.

tomaka:
Sorry to reopen the discussion :see_no_evil: 

[...]

PR #77, opened 2021-02-04
Improve SmoldotProvider event emitting

raoulmillais:
This is the implementation of lifecycle events outlined in #60.  I have deviated from the initially planned implementation.

After extensive testing against Westend and  comparing with the `WSProvider`, I discovered that the length of time to get peers is highly variable.  It can be almost instant but it can take over 10 seconds.  The client is still usable with no peers but it will not respond to certain queries.  Therefore it doesn't makes sense to simulate the "error if we don't connect behaviour" unless we get errors from the `system_health` request. 

Also after discussing with Gav and Pierre on element, there is another property on the health response that is relevant: `shouldHavePeers`. This is set to `false` for development chains which are not expected to have any peers and the `system_health` response is never expected to respond with `peers > 0`. (we haven't encountered such a scenario yet but will do later on).

So the final implementation is as follows:

1. When `connect` is called start an interval pinger asking for the `system_health`.
2. If we see `shouldHavePeers` is false emit the `connected` event and never emit `disconnected`
3. Otherwise .. 
  * if we're not connected and `peers > 0` emit the `connected` event
  * if we're connected and `peers === 0` emit the `disconnected` event
4. If we ever get an error response from `system_health` emit the `error` event

Previously we always emitted `connected` when successfully calling `connect` and only emitted `disconnected` if we called `disconnect`

I've kept the steps I took in the commit history experimenting with different event implementations so you can see what i tried if you're interested but if we're happy with the final outcome this PR should be squashed

### DIFF
--- a/packages/smoldot-provider/src/SmoldotProvider.test.ts
+++ b/packages/smoldot-provider/src/SmoldotProvider.test.ts
@@ -7,6 +7,11 @@ import { SmoldotClient, SmoldotOptions } from 'smoldot';
 type RpcResponder = (message: string) => string;
 
 const EMPTY_CHAIN_SPEC = '{}';
+
+const systemHealthReponse = peerCount => {
+  return `{\"jsonrpc\":\"2.0\",\"id\":42,\"result\":{\"isSyncing\":true,\"peers\":${peerCount},\"shouldHavePeers\":true}}"`;
+}
+
 // Mimics the behaviour of the WASM light client by deferring a call to 
 // `json_rpc_callback` after it is called which returns the response
 // returned by the supplied `responder`.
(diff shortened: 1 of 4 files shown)


raoulmillais:
Fixes #60

packages/smoldot-provider/src/SmoldotProvider.test.ts
```diff
@@ -196,6 +320,7 @@ test('subscribe', async t => {
       }
 
       t.deepEqual(result, { dummy: "state" });
+      provider.disconnect();
```
wirednkod, line 323: Shouldn't this be either await or disconnect().then(() => { resolve() })?

packages/smoldot-provider/src/SmoldotProvider.test.ts
```diff
@@ -283,4 +411,5 @@ test('unsubscribe removes subscriptions', async t => {
   const id = await provider.subscribe('test', 'test_subscribe', [], () => {});
   const reply =  await provider.unsubscribe('test', 'test_unsubscribe', id);
   t.true(reply);
+  provider.disconnect();
```
wirednkod, line 414: no need for await or resolve()?

Stefie (approved):
Looks good to me. I haven't looked into Smoldot or the provider any deeper so I can't give you a more specified review at this time. 
The behaviour makes sense and you discussed it in detail with Pierre , so I'm fine with merging if it's working.

packages/smoldot-provider/src/SmoldotProvider.test.ts
raoulmillais: It doesn't acutally matter because the tests run separately in parallel and ava cancels the pending timeout implicitly for us when the test ends.  We should wait for the provider to cancel its interval pinger to be consistent and faithful to how its intended to be used though :+1:

wirednkod (approved):

PR #84, opened 2021-02-08
UI Improvements for burning keys

goldsteinsveta:
Addressing #3. Changes:
- Hide recovery in context menu (keeping it for now to play with accounts while developing)
- Warning when burning keys from account with funds

### DIFF
--- a/projects/burnr/src/components/AccountMenu.tsx
+++ b/projects/burnr/src/components/AccountMenu.tsx
@@ -10,9 +10,7 @@ import { POLKA_ACCOUNT_ENDPOINTS } from '../utils/constants';
 import { useLocalStorage } from '../hooks';
 
 import ExpandMoreIcon from '@material-ui/icons/ExpandMore';
-import { Menu } from '@material-ui/core';
-import { ListItem } from '@material-ui/core';
-import { MenuItem } from '@material-ui/core';
+import { ListItem, Menu, MenuItem } from '@material-ui/core';
 
 const useStyles = makeStyles((theme: Theme) =>
 	createStyles({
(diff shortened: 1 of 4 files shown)


projects/burnr/src/components/AccountMenu.tsx
```diff
@@ -1,22 +1,31 @@
...
+import { Typography, makeStyles, Theme, createStyles, IconButton, Divider } from '@material-ui/core';
 
 import { AccountContext } from '../utils/contexts';
 
 import { openInNewTab, downloadFile, createLocalStorageAccount } from '../utils/utils';
 import { POLKA_ACCOUNT_ENDPOINTS } from '../utils/constants';
 import { useLocalStorage } from '../hooks';
 
+import ExpandMoreIcon from '@material-ui/icons/ExpandMore';
+import { Menu } from '@material-ui/core';
```
wirednkod, line 13: These 3 (Menu, ListItem, MenuItem) you can import them in 1 line:
import { Menu, ListItem, MenuItem } from '@material-ui/core';

projects/burnr/src/components/AccountMenu.tsx
```diff
@@ -42,68 +51,59 @@ const AccountMenu: React.FunctionComponent = () => {
...
+			<IconButton onClick={handleClick}>
+				<ExpandMoreIcon style={{color: grey[500]}}/>
+			</IconButton>
+
+			<Menu
+				transformOrigin={{vertical: -40, horizontal: 'left'}}
+				anchorEl={anchorEl}
+				keepMounted
+				open={Boolean(anchorEl)}
+				onClose={handleClose}
```
wirednkod, line 75: Instead of calling a function that will call the setter its better to do:
```
onClose={() => setAnchorEl(null)}
```
and remove function `handleClose` from above

wirednkod (approved):

PR #91, opened 2021-02-09
fix the files and building process for extension

wirednkod:
Solves  #90

### DIFF
--- a/projects/smoldot-extension/package.json
+++ b/projects/smoldot-extension/package.json
@@ -14,8 +14,8 @@
   },
   "scripts": {
     "clean": "rm -rf dist",
-    "build": "yarn run clean && parcel build src/*.html src/*.ts --no-source-maps --public-url ./",
-    "dev": "yarn run clean && parcel watch  src/*.html src/*.ts --public-url ./",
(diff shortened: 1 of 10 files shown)


raoulmillais (approved):

projects/smoldot-extension/package.json
```diff
@@ -9,11 +9,13 @@
...
   "staticFiles": {
-    "staticPath": "src/assets"
+    "staticPath": "src/assets",
+    "watcherGlob": "**"
   },
   "scripts": {
-    "build": "parcel build src/popup.html src/background.ts src/content.ts --no-source-maps --public-url ./",
-    "dev": "parcel watch src/popup.html src/background.ts src/content.ts --public-url ./",
+    "clean": "rm -rf dist",
+    "build": "yarn run clean && parcel build src/*.html src/*.ts --no-source-maps --public-url ./",
```
raoulmillais, line 17: Is this right?  `parcel` expects a list of *entry points* and creates one bundle for each entry point

raoulmillais (changes requested):
I think it's safer to be specifc about exactly which files are entry points to parcel in the build and dev scripts.  Otherwise if we add more html / ts files in the src folder they will automatically made into bundles, which is not what we want.

PR #107, opened 2021-02-10
Move smoldot mocking helpers into a separate package

raoulmillais:
We'll need to use these in the extension too so that we can fake smoldot in the tests

### DIFF
--- a/packages/smoldot-test-utils/.gitignore
+++ b/packages/smoldot-test-utils/.gitignore
@@ -1,0 +1,11 @@
+/dist
+
+# dependencies
+/node_modules
+
+# misc
+.DS_Store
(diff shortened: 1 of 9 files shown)


packages/smoldot-test-utils/.gitignore
```diff
@@ -0,0 +1,11 @@
+/dist
```
wirednkod, line 1: I think directories should not start with /

wirednkod (approved):

packages/smoldot-test-utils/.gitignore
raoulmillais: You're right. I copied this from another package!

PR #116, opened 2021-02-12
Fix race condition concerning smoldot-test-utils and add them as a de…

### LINKED ISSUE #114: yarn cmd fails on '@substrate/smoldot-test-utils' (error TS2307: Cannot find module '@substrate/smoldot-test-utils' or its corresponding type declarations.)
This creates an error upon yarn.

https://github.com/paritytech/substrate-connect/blob/29bcdea4f36b1b74a037f7d4797c1ad96511707a/packages/smoldot-provider/src/SmoldotProvider.test.ts#L13

Steps to reproduce:
- yarn run clean
- yarn

wirednkod:
…pendency only for smoldot-provider

Closes #114

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,8 @@
   "license": "Apache-2.0",
   "private": true,
   "workspaces": [
-    "packages/*",
+    "packages/module",
+    "packages/smoldot-provider",
     "projects/*"
   ],
   "scripts": {
--- a/packages/smoldot-provider/package.json
+++ b/packages/smoldot-provider/package.json
@@ -39,7 +39,8 @@
     "ava": "^3.15.0",
     "sinon": "^9.2.3",
     "ts-node": "^9.1.1",
-    "typescript": "^4.1.3"
+    "typescript": "^4.1.3",
+    "@substrate/smoldot-test-utils": "file:../smoldot-test-utils"
   },
   "peerDependencies": {
     "@polkadot/wasm-crypto": "^3.2.2"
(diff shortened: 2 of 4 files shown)


raoulmillais (commented):
Are there any implications of this change (making smoldot-test-utils a `file:` dependency) on publishing to npm? We'll want to publish both the provider and the test-utils eventually.

wirednkod:
Normally it should not as it is included on same repo but It may (i am not sure about). We should deal with it depending on how the npm packages will be packaged. 
As for now the smoldot-test-utils are needed only for smoldot-provider and installing/building through workspaces resulted to a race condition that made smoldot-provider to install first - and due to smoldot-test-utils being imported with no dependencies whatsoever and yarn was exiting with error, so this is resolving the race condition.

raoulmillais:
> Normally it should not as it is included on same repo but It may (i am not sure about). We should deal with it depending on how the npm packages will be packaged.
> As for now the smoldot-test-utils are needed only for smoldot-provider and installing/building through workspaces resulted to a race condition that made smoldot-provider to install first - and due to smoldot-test-utils being imported with no dependencies whatsoever and yarn was exiting with error, so this is resolving the race condition.

Cool.  It's obviously more important that everyone can run the applications and tests successfully  I'm also new to yarn workspaces, but we can cross this bridge when we come to thinking about publishing (if there even is a problem).

raoulmillais (approved):

Issue #123, opened 2021-02-16
Add Jest for UI code testing


wirednkod:
Add testing dependencies (ts-jest + react testing library) for Smoldot Extension

PR #124, opened 2021-02-16
Add ts-jest and React Testing Library on smoldot-extension

wirednkod:


wirednkod:
Closes #123

README.md
```diff
@@ -68,6 +68,15 @@ yarn dev:smoldot-browser-demo
 
 (Make sure to run `$ yarn install` before.)
 
+## Run local version of Burnr wallet
```
raoulmillais, line 71: ~Burnr wallet~ The Smoldot Extension

raoulmillais:
So you got it working without babel then?  That's awesome.  Maybe I just need to find a way to avoid using `import.meta.url` in the Smoldot Provider :thinking:

raoulmillais:
Ok this is potentially a contentious topic, and I realise you've just followed the jest default practice, but I think its highly preferable to have the test files sit side-by-side with the component they are testing instead of in an isolated folder.  (For the rest of this discusssion I'll use the word component but I don't mean it in the react sense but rather in the sense of the unit of functionality that you are testing, which more often than not is a single file).  Here are my reasons for this:

1. It encourages you to think about unit tests as part of your code, rather than in addition to. 
2. When you're browsing the codebase it's MUCH easier to see which components are missing tests - you don't have to correlate components with their tests in the separate folder.
3. When newcomers join the project they see immediately that there are tests associated with the code they're working on and don't have to go hunting elsewhere for them - so promotes the practice.
4. When you are review a pull request the files that change together sit together - so you're not jumping around the codebase.
5. If you extend this notion to helper utility files and styles, all the things that work together - sit together - you can look at the self contained bundle of functionality without jumping around the place.
6.  This is the same way things are done in Rust - so there is some synergy when diving in to the rust codebase.

This is not done uniformly across projects in parity / polkadotjs .. so maybe @stefie can add some history. It is for example the way tests are organised in [PolkadotJS API](https://github.com/polkadot-js/api/tree/master/packages/rpc-provider/src/ws) though.

raoulmillais:
Checked out the PR and everything runs nicely :+1: Interested to hear your thoughts on the above ^^ we should agree on this from the beginning as it is a real pain to reorganise the folder structures once the codebase starts growing.

goldsteinsveta (approved):

raoulmillais (approved):

PR #127, opened 2021-02-17
Add hooks and contexts. Add new sample mock Json. Refactor the code b…

wirednkod:
…ased on Tuesday's conversation


projects/smoldot-extension/src/components/IconWeb3.tsx
```diff
@@ -3,13 +3,21 @@ import styled from 'styled-components';
 
 // TODO: more icon-like props
 
-const Icon = styled.span`
+interface Props {
+    size?: string;
+    color?: string;
+}
+
+const Icon = styled('span')<Props>`
```
goldsteinsveta, line 11: I'm a bit meh about using MUI && styled components. Would stick to just MUI. This is not a blocker

projects/smoldot-extension/src/components/Tab.tsx
```diff
@@ -0,0 +1,97 @@
...
+import {
+    Grid,
+    Typography,
+    Box,
+    Accordion,
+    AccordionSummary,
+    AccordionDetails } from '@material-ui/core';
+import { Switch } from '.';
+import { TabInterface, uApp, Networks } from '../utils/types';
+import { IconWeb3 } from '../components';
```
goldsteinsveta, line 11: `{ Switch }` and `{ IconWeb3 }` are exported from the same place. Should it be done in one preferable go?

projects/smoldot-extension/src/utils/types.tsx
```diff
@@ -1 +1,20 @@
...
+    name: string;
+    enabled: boolean;
+}
+
+export type Networks = {
+    name: NetworkTypes;
+    status: NetworkStatus; 
+}
+
+export type NetworkCtx = TabInterface[];
```
goldsteinsveta, line 20: my obsessive–compulsive disorder goes brr when no new lines in the end of files (it's not you, it's me)

goldsteinsveta (approved):

projects/smoldot-extension/src/components/IconWeb3.tsx
wirednkod: Ok. I'll merge it for now and you can refactor that (and this way show me how you prefer it)

projects/smoldot-extension/src/components/Tab.tsx
wirednkod: Absolutely - that was probably an auto-correct with copy/pasting stuff

projects/smoldot-extension/src/utils/types.tsx
wirednkod: Its you - I must at some point setup an eslinter that can be used for all projects so there is consistency between react.node projects

projects/smoldot-extension/src/utils/types.tsx
```diff
@@ -1 +1,20 @@
-export type NetworkTypes = 'kusama' | 'polkadot' | 'westend' | 'kulupu'
\ No newline at end of file
+export type NetworkTypes = 'kusama' | 'polkadot' | 'westend' | 'kulupu'
```
raoulmillais, line 1: Can we move the types into the root please?  `./src/types.ts`  (also not `tsx`)  that's the same pattern as in polkadotjs  - local types in a `types.ts` file for the folder ... shared types in the shared context usually `../types.ts`

PR #143, opened 2021-02-24
Switch to using BN.js/ Polkadot JS Balance Types

### LINKED ISSUE #51: Switch to using BN.js/ Polkadot JS Balance Types instead of `number`
Every Balance, Block number etc should be declared as the corresponding types from Polkador JS types, not as `number`, because `number` is a 64 bit in JavaScript, but many integers in Polkadot/ Substrate can be larger than that (`u128` or `u256`)

```
import { Balance } from '@polkadot/types/interfaces';

export interface MyType {
	free: Balance;
	reserved: Balance;
	feeFrozen: Balance;
}

```

wirednkod:
Switch to using BN.js/ Polkadot JS Balance Types instead of `number` or `string`

Closes #51

### DIFF
--- a/projects/burnr/src/components/HistoryTable.tsx
+++ b/projects/burnr/src/components/HistoryTable.tsx
@@ -1,8 +1,8 @@
 import React, { useContext } from 'react';
 
 import BN from 'bn.js';
-import { TypeRegistry } from '@polkadot/types/create';
 import { AccountContext } from '../utils/contexts';
+import { Balance } from '@polkadot/types/interfaces';
 import { useBalance } from '../hooks';
 
 import { makeStyles } from '@material-ui/core/styles';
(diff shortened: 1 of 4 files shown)


projects/burnr/src/components/HistoryTable.tsx
```diff
@@ -138,9 +142,9 @@ const HistoryTable: React.FunctionComponent = () => {
 												{ // This may look overwhelming but is just for "dump" data until page is fixed
 												column.id === 'value'
 													&& typeof value === 'number'
-													&& <BalanceValue value={
-														new TypeRegistry().createType('Balance', new BN(value))
-													} />}
+													&& <BalanceValue
+														value={new TypeRegistry().createType('Balance', new BN(value))}
```
Stefie, line 146: The Type `Balance` is already registered and known by the Polkadot JS API, you only need to register new, unknown types. You can just pass the `value` here as it is already defined as Type `Balance` in the `BalanceValue` component. https://github.com/paritytech/substrate-connect/blob/3d5eb9be27b18a920a391620e74f894d22a25880/projects/burnr/src/components/BalanceValue.tsx#L10

If you want to use the Type, you can just import the type from the API:
```
import { Balance } from '@polkadot/types/interfaces';
```

wirednkod:
Closes #51

Issue #96, opened 2021-02-09
Content script that creates the port to communicate with the background page


raoulmillais:
The extension-side of the bridge between the SubstrateConnectProvider and the extension. It should 

* [x] expose an APi for ending and receiving JSON RPC messages. 
* [x] establish a port for communicating with the background script

PR #150, opened 2021-03-01
create the 'connections' between page <-> content <-> background. Integrate burnr (westend only) with SmoldotProvider.Identify existence of extension

wirednkod:


wirednkod:
Closes #96

wirednkod:
Closes #96 
Closes #144 

**For now** IF extension exists on browser and only for westend, the burnr wallet will connect through Smoldot-Provider.
In other scenarios it will connect through Polkadot-api (as it used to).

Issue #94, opened 2021-02-09
Background page and SmoldotClientManager


raoulmillais:
## Tasks
* [x] Upstream the typescript typings to smoldot
* [x] Extract the smoldot client test helpers into a separate package
* [x] Implement request and subscription proxying
* [x] Capture the tabId and URL when a Uapp connects
* [x] Implement loading and saving database
* [x] Add AppMediator tests
* [x] Add ConnectionManager tests
* [x] Clean up subscriptions on disconnect

#### What should the SmoldotClientManager do

The `SmoldotClientManager` inside the extension must broker RPC requests and responses from multiple clients residing on browser pages (aka content pages) to multiple smoldot blockchain clients in the extension (for now we have one but there will be many).   This entails:
* The `SmoldotProvider` in each webpage will issue RPC requests with ids in its own namespace (starting from 1). The `SmoldotClientManager` must maintain a translation table of client request ids to the actual request ids and send the correct responses to the correct clients. The translation table will be a map of clients to ID mappings where an ID mapping is the originating request id and the actual request id sent to smoldot. The `SmoldotClientManager` will change the request and response IDs in the JSON on the fly and maintain the list of live mappings.
  * Remove the ID mapping when the response has been received for non-subscriptions
  * Remove the ID mapping when an  unsubscribe request / response has happened
  * Remove all mappings when a page disconnects from the extension
* The `SmoldotClientManager` must detect when clients disconnect and send unsubscribe messages for all that clients' subscriptions.
* The `SmoldotClientManager` could also intelligently cache queries in the future (out of scope for now).

raoulmillais:
Upstream PR for typings in smoldot is here: https://github.com/paritytech/smoldot/pull/495

PR #160, opened 2021-03-09
Extension-provider && Providers returning Api

wirednkod:

PR #163, opened 2021-03-16
Unsubscribe from all subscriptions when a UApp disconnects

raoulmillais:
When the port for a UApp disconnects we put its `AppMediator` into the `disconnecting` state and send an unsubscribe message for all the active susbcriptions.  There is an additional check in the smoldot message processing to see if we're disconnecting and to change the state to `disconnected` once all those unsubscribe messages have been responded to by the smoldot.

The test for disconnecting adds some subscriptions; calls disconnect; simulates the unsub repsonses and checks the state of the UApp at all stages.

I also refactored the unit tests to make them more readable and reusable and to add better assertion coverage of all the behaviours.

#### Still needed

* Handling unsubscribe messages from the UApp - I need to coordinate with you after we've merged this and your work to get the `ExtensionProvider` to tell the background about unsubscriptions.
* I also added a note about sending an info message back to the UApp to tell it that the associate request was successful... you had this logic in your branch but it does not exist on master.
* ~Remove the `AppMediator` from the `ConnectionManager` when the state has become `disconnected`~ :heavy_check_mark:

### DIFF
--- a/projects/extension/src/background/AppMediator.ts
+++ b/projects/extension/src/background/AppMediator.ts
@@ -92,8 +92,7 @@ export class AppMediator {
         if (this.requests.length === 0) {
           // All our unsubscription messages have been replied to
           this.#state = 'disconnected';
-          // TODO: remove this AppMediator from ConnectionManager and SmoldotMediator
-          // TODO: send notifications
+          this.#manager.unregisterApp(this, this.#smoldotName as string);
         }
 
         return true;
--- a/projects/extension/src/background/AppMediator.ts
+++ b/projects/extension/src/background/AppMediator.ts
@@ -187,7 +187,7 @@ export class AppMediator {
       this.#sendError(`Extension does not have client for ${name}`);
       return;
     }
-    this.#manager.registerAppWithSmoldot(this, name);
+    this.#manager.registerApp(this, name);
     this.#smoldotName = name;
     this.#state = 'ready';
     return;
--- a/projects/extension/src/background/AppMediator.ts
+++ b/projects/extension/src/background/AppMediator.ts
@@ -78,7 +78,7 @@ export class AppMediator {
     if (this.#state === 'disconnected') {
       // Shouldn't happen - we remove the AppMediator from the smoldot's apps
       // when we disconnect (below).
-      console.warn(`Asked a disconnected UApp (${this.name}) to process a message from ${this.#smoldotName}`);
+      console.warn(`Asked a disconnected UApp (${this.name}) to process a message from ${this.#smoldotName as string}`);
       return false;
     }
 
@@ -210,7 +210,7 @@ export class AppMediator {
     this.#handleDisconnect(true);
   }
 
-  #sendUnsubscribe = (sub: SubscriptionMapping, subIndex: number) => {
+  #sendUnsubscribe = (sub: SubscriptionMapping): void => {
     // use one higher than we've seen before from the UApp.  The UApp is now
     // disconnnecting so this won't ever be reused as we no longer
     // accept incoming RPC send requests
--- a/projects/extension/src/background/AppMediator.ts
+++ b/projects/extension/src/background/AppMediator.ts
@@ -23,6 +23,8 @@ export class AppMediator {
   #state: AppState = 'connected';
   readonly subscriptions: SubscriptionMapping[];
   readonly requests: MessageIDMapping[];
+  highestUAppRequestId = 0;
+  #notifyOnDisconnected = false;
 
   constructor(name: string, port: chrome.runtime.Port, manager: ConnectionManagerInterface) {
     this.subscriptions = [];
@@ -40,6 +42,10 @@ export class AppMediator {
 
   get name(): string {
     return this.#name;
+  }
+
+  get smoldotName(): string | undefined {
+    return this.#smoldotName;
   }
 
   get tabId(): number | undefined {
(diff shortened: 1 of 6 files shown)


raoulmillais:
Coverted to a WIP PR - I'll rebase this off master when #160 is merged and fix up any conflicts

wirednkod:
@raoulmillais We need to finalize the format of messages for uApp <-> background <-> smoldot in order for the uApp to be informed on what happened to the message sent. I write this as reply of the note:
```
I also added a note about sending an info message back to the UApp to tell it that the associate request was successful... you had this logic in your branch but it does not exist on master.
```

raoulmillais:
fixes #94

projects/extension/src/background/AppMediator.ts
```diff
@@ -69,6 +75,30 @@ export class AppMediator {
   }
 
   processSmoldotMessage(message: JsonRpcResponse): boolean {
+    if (this.#state === 'disconnected') {
```
wirednkod, line 78: add these in if () { } else if { } else.
This way you "save" some time for addition if checks (once 1 if succeeds rest checks wont follow)

projects/extension/src/background/AppMediator.ts
```diff
@@ -69,6 +75,30 @@ export class AppMediator {
...
+      // Shouldn't happen - we remove the AppMediator from the smoldot's apps
+      // when we disconnect (below).
+      console.warn(`Asked a disconnected UApp (${this.name}) to process a message from ${this.#smoldotName as string}`);
```
wirednkod, line 81: respective info message should return

projects/extension/src/background/AppMediator.ts
```diff
@@ -174,10 +209,33 @@ export class AppMediator {
...
+    // send the unsubscribe message
+    const smoldotID = this.#manager.sendRpcMessageTo(this.#smoldotName as string,  unsubRequest);
+    // track the request so we know when its completed
+    this.requests.push({ appID, smoldotID });
+  };
+
   #handleDisconnect = (notify: boolean): void => {
-    console.log(`Disconnecting and notify is set to ${notify.toString()}`);
+    if (this.#state === 'disconnecting' || this.#state === 'disconnected') {
+      throw new Error('Cannot disconnect - already disconnecting / disconnected');
```
wirednkod, line 232: Im not sure if this should throw an error or return an error message in a specific format for uApp to handle (This is not an exception error but a "flow/time" error)

projects/extension/src/background/mocks.ts
```diff
@@ -40,21 +40,27 @@ export class MockPort implements chrome.runtime.Port {
...
   readonly #willFindClient: boolean;
+  lastId = 0;
+
 
   constructor(willFindClient: boolean) {
     this.#willFindClient = willFindClient;
   }
 
-  registerAppWithSmoldot(app: AppMediator, name: string) {
+  registerAppWithSmoldot(app: AppMediator, name: string): void {
```
wirednkod, line 50: what `registerAppWithSmoldot`and `unregisterApp`do?

wirednkod (changes requested):

projects/extension/src/background/AppMediator.ts
raoulmillais: Could you elaborate please?  I don't understand.  This is a guard clause.. it returns early if the state is disconnected because it shouldn't be processing the message

projects/extension/src/background/AppMediator.ts
raoulmillais: what do you mean?  I should throw it as an error?  Even though this state shouldn't happen I was being cautious that it might in an unconsidered scenario / race-condition hence logging a warning.

projects/extension/src/background/AppMediator.ts
raoulmillais: I did it this way because it's "programmer error"  ... this happens if a the port disconnects twice I.E. it should not ever happen.. a chrome.runtime.Port should never trigger a disconnect event twice.  If it does we really want to know about it because it means my assumptions about ports are all wrong! Hence an error.

projects/extension/src/background/mocks.ts
raoulmillais: They are book-keeping methods.. `registerAppWithSmoldot` tells the specified `SmoldotMediator` about a new app (it adds the app to its internal apps list) ... when incoming messages arrive from that smoldot we only ask the apps that have been registered with that smoldot to process that message (not all the apps as the others arent interested).  `unregisterApp` does the opposite, it tells the smoldot that that app is now disconnected so that we stop asking it to process messages that come from that smoldot.

Now I reread this I think that mnames should be symetrical :thinking:  `registerApp` and `unregisterApp`

projects/extension/src/background/AppMediator.ts
wirednkod: Misread the code here a bit - I was mentioning the way if is written but please disreagard this. Its my mistake

projects/extension/src/background/AppMediator.ts
raoulmillais: We can't send the message back at this point .. the port is disconnected.  The scenario i have in mind is that the UApp sends an RPC message immediately before disconnecting... we might get the response back for that message and already be disconnected (although we shouldn't because unregistered the app with the smoldot).  If we start seeing these warnings in the dev console there is probably a bug in the AppMediator code

projects/extension/src/background/mocks.ts
raoulmillais: Ive renamed it for consistency

projects/extension/src/background/AppMediator.ts
raoulmillais: So maybe `console.error` is better?  It shows up as an error but doesn't cause the extension to crash (the extension may be servicing otherwise working UApps)

wirednkod: I agree. Its better to "save" the flow at this point and just inform through console

raoulmillais: OK done

raoulmillais:
Is this good to merge now?  I think I've addressed all your comments

wirednkod (approved):

PR #172, opened 2021-03-17
Fetch chainspec json with cache busting headers

### LINKED ISSUE #113: Force browser to get most recent version of westend.json
If you've already loaded the westend.json and its in your cache the browser will return the cached version.

raoulmillais:
Add headers to ask browser and proxies to return uncached versions of the westend.json in the smoldot-browser-demo.  This should solve issues where we update the chainspec but when we refresh the smoldot-browser-demo we still get the old version.

Fixes #113

### DIFF
--- a/projects/smoldot-browser-demo/src/index.ts
+++ b/projects/smoldot-browser-demo/src/index.ts
@@ -12,7 +12,12 @@ window.onload = () => {
 
   (async () => {
     try {
-      const response =  await fetch('./assets/westend.json')
+      const headers = new Headers();
+      headers.append('pragma', 'no-cache');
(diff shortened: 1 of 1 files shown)


wirednkod (changes requested):
I am not sure that this is needed...As discussed no uApp should add their chainSpecs. What we need to do here is to put this implementation inside the Substrate Connect in order to receive latest JSONs. I would leave this as is, and alter the substrate/connect -> smoldot provider to just receive a name (e.g. 'westend') and "download" the chainspecs if needed.

raoulmillais:
Ah yeah of course! :facepalm: Make sure you use the same technique where you fetch the chainspec though :pray:

PR #173, opened 2021-03-17
Enable linting and fix issues for SmoldotProvider

### LINKED ISSUE #136: Fix lint issues on packages
ESLint (TS) is setup up but bypassed for packages as issues needs to be fixed on:

- [x] smoldot-provider
- [x] smoldot-test-util

raoulmillais:
I ignored quite a few rules because we have intentionally copied from the PolkadotJS api package which uses `any` in its provider in quite a few ways.  Ideally we should resolve this at some point but at least for now this means we actually have linting in this project and we should catch any future linting problems if we make changes.

This fixes #136 although I have not added linting to the test-utils, I dont think that project should be linted ... we actually want to be loose on our definitions of things for test helpers. We're intentionally faking / hidding / not providing implementations of things to test functionality.

### DIFF
--- a/packages/smoldot-provider/src/FsDatabase.ts
+++ b/packages/smoldot-provider/src/FsDatabase.ts
@@ -25,9 +25,9 @@ export class FsDatabase implements Database {
   load(): string {
     try {
       statSync(this.#path);
-    } catch (error: any) { 
+    } catch (error: any) { // eslint-disable-line @typescript-eslint/no-explicit-any
       // Typescript does not allow type annotations on catch blocks :(
-      if (error.code === 'ENOENT') {
+      if ((error as NodeJS.ErrnoException).code === 'ENOENT') {
         return '';
       }
 
@@ -37,11 +37,11 @@ export class FsDatabase implements Database {
     return readFileSync(this.#path, { encoding: 'utf-8' });
   }
 
-  save(state: string) {
+  save(state: string): void {
     writeFileSync(this.#path, state);
   }
 
-  delete() {
+  delete(): void {
       unlinkSync(this.#path);
   }
 }
--- a/packages/smoldot-provider/src/FsDatabase.ts
+++ b/packages/smoldot-provider/src/FsDatabase.ts
@@ -25,7 +25,7 @@ export class FsDatabase implements Database {
   load(): string {
     try {
       statSync(this.#path);
-    } catch (error: any) { // eslint-disable-line @typescript-eslint/no-explicit-any
+    } catch (error: unknown) {
       // Typescript does not allow type annotations on catch blocks :(
       if ((error as NodeJS.ErrnoException).code === 'ENOENT') {
         return '';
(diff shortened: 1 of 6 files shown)


packages/smoldot-provider/src/FsDatabase.ts
```diff
@@ -25,9 +25,9 @@ export class FsDatabase implements Database {
   load(): string {
     try {
       statSync(this.#path);
-    } catch (error: any) { 
+    } catch (error: any) { // eslint-disable-line @typescript-eslint/no-explicit-any
```
wirednkod, line 28: You should not disable linter. There are error types that can be used. Any should not be used

packages/smoldot-provider/src/SmoldotProvider.ts
```diff
@@ -18,10 +18,13 @@ import { HealthCheckError } from './errors';
...
 const l = logger('smoldot-provider');
 
+// Note types are copied from polkadotjs for consistency - `any` types are used
+// there
+
 interface RpcStateAwaiting {
   callback: ProviderInterfaceCallback;
   method: string;
-  params: any[];
+  params: any[]; // eslint-disable-line @typescript-eslint/no-explicit-any
```
wirednkod, line 27: You should not disable linter. Any should not be used

packages/smoldot-provider/src/SmoldotProvider.ts
```diff
@@ -34,7 +37,7 @@ interface SubscriptionHandler {
 
 interface StateSubscription extends SubscriptionHandler {
     method: string;
-    params: any[];
+    params: any[]; // eslint-disable-line @typescript-eslint/no-explicit-any
```
wirednkod, line 40: You should not disable linter. Any should not be used

wirednkod (changes requested):

packages/smoldot-provider/src/FsDatabase.ts
raoulmillais: Yeah missed this one I can use `NodeJS.ErrnoException`

packages/smoldot-provider/src/SmoldotProvider.ts
raoulmillais: This came from PolkadotJS ... I wanted to keep it the same as the original code intentionally

packages/smoldot-provider/src/SmoldotProvider.ts
raoulmillais: This came from PolkadotJS ... I wanted to keep it the same as the original code intentionally

packages/smoldot-provider/src/FsDatabase.ts
raoulmillais: oh no it cant .. you aren't allowed to specify error types in catch blocks.  Typescript won't allow it.  I've changed it to unkown instead, that's better. 

see https://stackoverflow.com/questions/42618089/how-do-you-use-typed-errors-in-async-catch#42618403

raoulmillais:
The provider doesn't actually touch the `params` it stores for subscriptions or handlers it just stores them.  I copied the state types from the websocket provider which is using them for resubscribing after a disconnection ... this makes sense for a websocket provider but not for our usecases.  I can avoid having `any` types by removing those properties from the types.

I'm putting this back to draft (on hold) until we've finalised the move of the providers #174

Issue #177, opened 2021-03-18
Remove packages @substrate/Smoldot-provider and @substrate/extension-provider 


wirednkod:
I tried to completely remove packages @substrate/Smoldot-provider and @substrate/extension-provider but there is a relation with the projects (mainly burnr) that makes the overall yarn install to break - so for now I let them exist inside the package

Issue #176, opened 2021-03-18
Migrate api.examples.ts from Smoldot-provider to @substrate/connect


wirednkod:
if we need those then proceed with ticket.
If not we can just close this ticket

wirednkod:
This is partially fixed - There are some bugs that appear probably due to smoldot

Issue #171, opened 2021-03-17
Merge SmoldotProvider and ExtensionProvider into @substrate/connect


raoulmillais:

wirednkod:
Fixed in #179

Issue #170, opened 2021-03-17
Convert burnr wallet to use @substrate/connect


raoulmillais:

wirednkod:
Fixed in #179

PR #179, opened 2021-03-18
Move Smoldot and Extension providers inside  @substrate/connect

wirednkod:
Tasks remaining:

- [x] Integrate burnr wallet with @substrate/connect (for now the project has been removed from package.json as build was failing
- [x] Revert change in main package.json that will add burnr wallet back to proejcts list
- [ ] Fix tests issue with disconnection of smoldot client (Note: all tests are fixed to work with JEST but there is sitll issue with Worker. Relevant ticket opened: #185 )


wirednkod:
@raoulmillais  Please have a look when you have some time on tests of Smoldot Provider. as I have altered them in order to make them work. I just need a confirmation that the logic is still there correctly

packages/connect/src/SmoldotProvider/SmoldotProvider.test.ts
```diff
@@ -204,78 +185,92 @@ test('subscribe', async t => {
...
+//         reject();
+//       }
+
+//       expect(result).toEqual({ dummy: "state" });
+//       return provider.disconnect().then(() => resolve());
+//     }).then(reply => {
+//       expect(reply).toBe("SUBSCRIPTIONID");
+//     });
+//   });
+// });
```
raoulmillais, line 271: There are still several commented out tests.  I'm happy to help getting these to work if you want me too.

raoulmillais (changes requested):
The tests you've converted look good to me - I'd like to get the subscription tests fully working though, they test the subscribe happy path, what happens when we receive subscription methods out of order and that we translate american spellled RPC method names to british spelling.

packages/connect/src/SmoldotProvider/SmoldotProvider.test.ts
wirednkod: I just fixed those 😄

wirednkod:
Tasks completed:
- Move @substrate/SmoldotProvider and @substrate/Extension provider inside @substrate/connect - along with the log history;
- Alter all existing AVA tests to make them work with Jest;  Comment in tests that were commented out in previous PR
- Migrate api.tests.js from AVA to Jest **(There is still an issue that seems to be coming from smoldot)**;
- Add extra packages in extension and burnr in order to be able to install independently from other projects;
- Integrate burnr wallet with @substrate/connect by replacing smoldotProvider from it;

Fixes #177 #176 #125 #171 #170

packages/connect/package.json
```diff
@@ -15,25 +18,62 @@
     "url": "https://github.com/paritytech/substrate-connect/issues"
   },
   "homepage": "https://github.com/paritytech/substrate-connect#readme",
+  "scripts": {
+    "prebuild": "yarn clean",
+    "test-trace": "node --experimental-vm-modules --trace-warnings node_modules/.bin/jest dist",
+    "test": "node --experimental-vm-modules node_modules/.bin/jest dist",
```
raoulmillais, line 24: We need a `pretest` script to run the build or this will fail on a fresh clone (and run the tests against out-of-date built files potentially)

packages/connect/src/Detector.test.ts
```diff
@@ -0,0 +1,8 @@
+import * as smoldot from 'smoldot';
+
+describe('Test Detector (Fake)', () => {
+    it('works', () => {
+        console.log(smoldot);
```
raoulmillais, line 5: Remove the `cosnole.log`  and can we make this test do something a little more sensible at least?  Maybe check that `Detector.connect` returns something and doesn't error that's a reasonable happy path case

raoulmillais:
The changes look good to me. However when I run the tests locally I get "ERR: The Client is not initialised" messages for all the api tests.   I think you can fix the RPC query tests by making them purely promise based and not async/await, I started trying to verify this myself but I got stimied by the error messages about smoldot not being ready

raoulmillais (changes requested):

packages/connect/src/Detector.test.ts
wirednkod: This test was in the scope of "fixing jest to work with substrate connect".
I will remove it for now. Tests will be implemented on different PR as this one is already too large.

wirednkod:
> The changes look good to me. However when I run the tests locally I get "ERR: The Client is not initialised" messages for all the api tests. I think you can fix the RPC query tests by making them purely promise based and not async/await, I started trying to verify this myself but I got stimied by the error messages about smoldot not being ready

RPC needs no fixing as async/await works fine (and tests). Issue (as you mentioned before) is that smoldot never stops/disconnects. and it keeps logging when tests are done. All tests run successfully (besides the aforementioned error) but as soon as we get this error we cannot proceed with merging. 
I am not sure on how we should proceed here

raoulmillais:
You only git ignored coverage in the `packages/connect` gitignore not in the root.  That's why we still see the coverage files in the repo

packages/connect/src/SmoldotProvider/SmoldotProvider.ts
```diff
@@ -280,34 +281,33 @@ export class SmoldotProvider implements ProviderInterface {
...
-      this.#client = undefined;
-    }
-
-    if (this.#connectionStatePingerId !== null) {
-      clearInterval(this.#connectionStatePingerId);
-    }
+    try {
+      if (this.#client) {
+        // eslint-disable-next-line @typescript-eslint/unbound-method
+        this.#client.terminate;
```
raoulmillais, line 299: This should be a function call `this.#client.terminate()`

packages/connect/src/SmoldotProvider/SmoldotProvider.ts
```diff
@@ -280,34 +281,33 @@ export class SmoldotProvider implements ProviderInterface {
...
   // eslint-disable-next-line @typescript-eslint/require-await
   public async disconnect(): Promise<void> {
-    if (this.#client) {
-      this.#client = undefined;
-    }
-
-    if (this.#connectionStatePingerId !== null) {
-      clearInterval(this.#connectionStatePingerId);
-    }
+    try {
```
raoulmillais, line 296: This `try` / `catch` covers more than it needs to and could accidentally mask unreasoned about errors.  I think you meant only to wrap the `disconnect()` in a `try` / `catch`

packages/connect/src/SmoldotProvider/SmoldotProvider.ts
```diff
@@ -280,34 +281,33 @@ export class SmoldotProvider implements ProviderInterface {
...
 
-    this.#isConnected = false;
-    this.emit('disconnected');
+      if (this.#connectionStatePingerId !== null) {
+        clearInterval(this.#connectionStatePingerId);
+      }
 
-    return Promise.resolve();
+      this.#isConnected = false;
+      this.emit('disconnected');
```
raoulmillais, line 307: If you are using a `try`/`catch` these lines ought to be in a `finally` block ... we want to clear the interval / set the isconnected state and emit disconnected regardles of whether there is an error or not.  This provider is no longer usable without trying to reconnect

projects/burnr/src/hooks/api/useApiCreate.ts
```diff
@@ -35,13 +35,15 @@ export default function useApiCreate (): ApiPromise {
...
-        await provider.connect();
-        await provider.send('system_health', []);
+        const detect = new Detector('westend', chainSpec);
+        const api = await detect.connect();
+        // const provider = new SmoldotProvider(chainSpec);
+        // await provider.connect();
+        // await provider.send('system_health', []);
         // TODO:  API should be included inside the substrate connect (Both SmoldotProvider and extension should return an API most probably)
-        const api = await ApiPromise.create({ provider });
+        // const api = await ApiPromise.create({ provider });
```
raoulmillais, line 46: Remove the TODO and commented out code

projects/burnr/src/hooks/api/useApiCreate.ts
```diff
@@ -35,13 +35,15 @@ export default function useApiCreate (): ApiPromise {
...
         mountedRef.current && setApi(api);
       } catch (err) {
         console.log('A wild error appeared:', err);
```
raoulmillais, line 49: Remove the `try` / `catch` and let unexpected errors be thrown in the dev console.. that way they will have better context information for debugging and be easier to spot and fix

raoulmillais (changes requested):
See my comments inline and the comment about git ignore

packages/connect/src/SmoldotProvider/SmoldotProvider.ts
wirednkod: Tried that and for some reason it wont work (Will get error .terminate() is not a function).
You can give it a try - maybe I am missing something

projects/burnr/src/hooks/api/useApiCreate.ts
wirednkod: This is not correct. Unexpected error will make the app to break completely. We do not want that

packages/connect/src/SmoldotProvider/SmoldotProvider.ts
raoulmillais: This is a new bug.  All the tests  pass calling `terminate()`  except  one which fails with this:

```
Expected: "Got error response asking for system health"
Received: "__classPrivateFieldGet(...).terminate is not a function"".] {
  code: 'ERR_UNHANDLED_REJECTION'
}
```
An unhandled rejection error has been assigned to `this.#client` .. so something has gone wrong in the refactoring from Promise control flow to `async` / `await` with `try` / `catch` in the `connect()` function

raoulmillais: It's also possible this is a new bug in `smoldot.start`  try reverting to using a `Promise` with `then` and `catch` and see what happens if we still end up with an error being assigned to `this.#client` then the problem is in smoldot

projects/burnr/src/hooks/api/useApiCreate.ts
raoulmillais: Fair enough ... I don't know about the error handling strategy in burnr.  
Not for this PR but shouldn't we be showing some sort of error UI instead of ignoring the error?  Presumably burnr is unusable with a broken API client

.gitignore
```diff
@@ -1,4 +1,7 @@
 node_modules
 .DS_Store
 yarn-error.log
-dist
\ No newline at end of file
+dist
+
+# testing
+/coverage
```
raoulmillais, line 7: I think you want a trailing slash not a leading slash.  Also don't forget to `git rm -rf coverage/` in the root

packages/connect/src/SmoldotProvider/SmoldotProvider.ts
wirednkod: We should see that on scope of #185 and not on this  ticket.  Lets keep the PR on single scope

projects/burnr/src/hooks/api/useApiCreate.ts
wirednkod: There is already an open issue for that #86

packages/connect/src/SmoldotProvider/SmoldotProvider.ts
raoulmillais: I agree that we should try to get this merged asap and move onto fixing the outstanding issues to do with jest thinking there is a leak separately.  This thing however is an unrelated and new bug that has been introduced (probably by the refactoring to async/await in connect) and is definitely not the right logic (1. terminate is a function and 2, we weren't assigning errors to `this/#client` before). This means we are neither calling `terminate` (we just have a reference to the property) and we are guaranteed to throw errors to users in some cases.

I recommend we reinstate the old body of `connect()` in `SmoldotProvider.ts` and change `this.#client.terminate;` to `this.#client.terminate();` in `disconnect()` and then move on with merging .. that way disconnect in master isn't broken while we investigate the jest memory leaking issues

raoulmillais (approved):

PR #188, opened 2021-03-25
Lint connect package

raoulmillais:
This enables linting for @substrate/connect  I have removed the `params` from the subscription state in both providers.  This was copy/pasted over from the `WsProvider` where it keeps track of that so that it can resubscribe to any subscriptions if the websocket connection drops.  This doesn't make sense for our providers as we shouldn't have this connectivity loss problem

### DIFF
--- a/packages/connect/src/ExtensionProvider/ExtensionProvider.ts
+++ b/packages/connect/src/ExtensionProvider/ExtensionProvider.ts
@@ -22,7 +22,6 @@ const l = logger(EXTENSION_ORIGIN);
 interface RpcStateAwaiting {
   callback: ProviderInterfaceCallback;
   method: string;
-  params: any[];
   subscription?: SubscriptionHandler;
 }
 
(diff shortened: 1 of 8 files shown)


packages/connect/src/ExtensionProvider/ExtensionProvider.ts
```diff
@@ -165,6 +156,7 @@ export class ExtensionProvider implements ProviderInterface {
     }
     window.postMessage(initData, '*');
     window.addEventListener('message', ({data}) => {
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
```
wirednkod, line 159: You should follow the guidelines on how to correctly lint this instead of disabling linter.
For no-unsafe-member-access rule there are some correct guidelines here:
https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-unsafe-member-access.md

Same for across PR where disables take place

wirednkod (changes requested):

packages/connect/src/ExtensionProvider/ExtensionProvider.ts
raoulmillais: I think I cleared this up in chat yesterday. Can you approve this now?

raoulmillais: I opened #189

wirednkod: @raoulmillais - Yes we cleared this up. Comment was before our talk

wirednkod (approved):

Issue #169, opened 2021-03-17
UApps should provide a name to the background script for display purposes


raoulmillais:
Send a message (or use the existing `associate` message) supplying a display name for the UApp.  As noted [here](https://github.com/paritytech/substrate-connect/pull/163#issuecomment-800882299) we should have a design discussion to review and formalise the messages passed.

raoulmillais:
This is partially implemented by #194 and will probably be completed by implementing #126 and #95

Issue #126, opened 2021-02-17
State management for the UI


raoulmillais:
See https://github.com/paritytech/substrate-connect/issues/93 for the specification. Create the API that the UI will interact with for getting the shared state.

PR #195, opened 2021-03-29
Allow connecting to multiple networks

raoulmillais:
While doing #126 I realised that you can only ever create a single port from a tab.  This restricts an unstoppable app to a single network. 
This PR makes it possible to connect to multiple networks from the same tab.  In the process I have changed the Detector to accept a display name (which partially solves #169 but it still needs to parsed out of the port name in the background).

The new API is

```javascript
const detector = new Detector('my-cool-uapp');
const kusama = detector.connect('kusama');
const polkadot = detector.connect('polkadot');

/// use the 2 API clients

await detector.disconnect('kusama');
await detector.disconnect('polkadot');
```

Unfortunately I couldn't properly verify the changes because the upgrade to smoldot does not work with parcel.


packages/connect/src/Detector.ts
```diff
@@ -1,37 +1,40 @@
...
+  get name(): string {
+    return this.#name;
+  }
+
+  public constructor (name: string) {
+    this.#isExtension = !!document.getElementById('substrateExtension');
+    this.#name = name;
+  }
+
+  public connect = async (chainName: string, chainSpec?: string): Promise<ApiPromise> => {
```
wirednkod, line 20: I thought chainspec will not be provided from Apps

packages/connect/src/ExtensionProvider/ExtensionProvider.test.ts
```diff
@@ -0,0 +1,38 @@
+import {jest} from '@jest/globals'
+import { ExtensionProvider } from './ExtensionProvider';
+
+describe('ExtensionProvider', () => {
+
+  beforeEach(() => {
+    window.postMessage = jest.fn();
+  });
+
+  it('constructor sets properties', async () => {
```
wirednkod, line 10: use either test or it (I think we ended up on using solely test)

packages/connect/src/ExtensionProvider/ExtensionProvider.ts
```diff
@@ -15,9 +15,19 @@ import { logger } from '@polkadot/util';
 import EventEmitter from 'eventemitter3';
 import { isUndefined } from '../utils';
 
-const EXTENSION_ORIGIN = 'extension-provider';
+const origins = {
```
wirednkod, line 18: For constants use variables instead of object.

packages/connect/src/SmoldotProvider/SmoldotProvider.test.ts
```diff
@@ -168,9 +168,9 @@ test('throws when got error JSON response', async () => {
 
   await provider.connect();
 
-   await expect(() => provider.send('hello', [ 'world' ]))
-    .rejects
-    .toThrow('666: boom!');
+  await expect(() => provider.send('hello', [ 'world' ]))
```
wirednkod, line 171: This should be removed - right?

projects/extension/src/content.ts
```diff
@@ -1,43 +1,63 @@
...
-//This line opens up a long-lived connection FROM the page TO your background page.
 
-const port = chrome.runtime.connect({ name: PORT_CONTENT });
+const ports: Record<string, chrome.runtime.Port> = {};
 
-export type AppMessageType = 'associate' | 'rpc';
-export interface AppMessage {
-  type: AppMessageType;
-  payload: string; // name of the network or an rpc string (json stringified RPC message)
+const origins = {
```
wirednkod, line 7: use variables for constants

projects/multiple-network-demo/src/index.ts
```diff
@@ -0,0 +1,37 @@
+/* eslint-disable */
```
wirednkod, line 1: There is no point to install linter if everything is disabled in files. Its better to remove it completely

wirednkod (changes requested):

packages/connect/src/Detector.ts
raoulmillais: Yup but we hadn't fully decided what to do while I was working on this. Also the goal of this PR was not to deal with that .. so I just did the smallest change to satisfy what I was doing.  We can discuss and change the public API for chainspecs in a future PR

packages/connect/src/ExtensionProvider/ExtensionProvider.ts
raoulmillais: Fine by me but out of interest .. why?

projects/multiple-network-demo/src/index.ts
raoulmillais: Oops yeah the linter insnt enabled for this yet .. I think I did this to remove all the warnings in my editor so I could see the typescript errors more easily :smile: i'll enable linting and fix the errors.

packages/connect/src/ExtensionProvider/ExtensionProvider.ts
wirednkod: Because it is a constant.. it should not change later in the code. When you add it in an object key, you can alter it and that makes it lose the value of being a constant.
e.g. 
```
const a = '1';
a = '2';
->> Uncaught TypeError: Assignment to constant variable.
-----------
const a = { b: '1' }; // console.log(a) -> {b: '1'}
a.b = '2'; // console.log(a) -> {b: '2'}
```

packages/connect/src/SmoldotProvider/SmoldotProvider.test.ts
raoulmillais: The only changes in this file are whitespace fixes.  I havent touched the functionality.  That said .. this test still looks valid to me: We mock an error response and expect a rejected promise - that's an inteded behaviour that we want to test for.

packages/connect/src/Detector.ts
wirednkod: There was a question about that also in CI channel. We should consider leaving that option - but lets discuss in later stage

packages/connect/src/ExtensionProvider/ExtensionProvider.ts
raoulmillais: Ah yeah good point!

wirednkod (approved):

projects/multiple-network-demo/src/index.ts
```diff
@@ -0,0 +1,32 @@
+// hack to make poladot-js work without bringing in webpack and babel
+import "regenerator-runtime/runtime"
+import { Detector }  from '@substrate/connect';
+
+window.onload = () => {
+  void (async () => {
+    try {
+      const detect = new Detector('Multiple Network Demo');
+
+
```
wirednkod, line 10: remove double blank line

projects/extension/src/content.ts
```diff
@@ -1,43 +1,61 @@
...
-      }
-    } else {
-      appData = JSON.parse(data?.message || '');
-    }
-    port.postMessage({ ...appData, origin: EXTENSION_ORIGIN});
+  if (!data.origin || data.origin !== EXTENSION_PROVIDER_ORIGIN) {
+    return;
   }
-});
+  debug(`RECEIEVED MESSAGE FROM ${EXTENSION_PROVIDER_ORIGIN}`, data);
```
wirednkod, line 21: typo on RECEIEVED

wirednkod (approved):

PR #205, opened 2021-04-06
Add ErrorBoundary component in Burnr wallet

wirednkod:


projects/burnr/src/hooks/api/useApiCreate.ts
```diff
@@ -42,8 +42,8 @@ export default function useApiCreate (): ApiPromise {
         console.log("API api", api);
         mountedRef.current && setApi(api);
       })
-      .catch((): void => {
-        console.error
+      .catch((err): void => {
+        console.log(err)
```
raoulmillais, line 46: Better to use `console.error(err)`

projects/burnr/src/hooks/api/useApiCreate.ts
```diff
@@ -11,7 +11,7 @@ import { LazyProvider } from '../../utils/types';
 import { useIsMountedRef, useLocalStorage } from '..';
 import westend from '../../assets/westend.json';
 
-console.log('ALL_PROVIDERS', ALL_PROVIDERS)
+console.log('ALL_PROVIDERS: ', ALL_PROVIDERS)
```
raoulmillais, line 14: leftover `console.log` ?

raoulmillais (approved):

projects/burnr/src/hooks/api/useApiCreate.ts
wirednkod: This is there since beginning of project. For now I would leave it as its info that can be valuable.

raoulmillais: yeah we need a shared debugging library for genuinely useful debug messages that only appear when in local dev.  Ive added a `debug()` function to the extension for this

PR #223, opened 2021-04-13
ConnectionManager State management

### LINKED ISSUE #126: State management for the UI

### LINKED ISSUE #213: Refreshing the page does not disconnect and reconnect the port
This results in an error being sent back to the content script saying the App already exists.

raoulmillais:
This pull request implements emitting events when the state of the background multiplexing changes for the UI to update accordingly (fixes #126).  It is slightly simpler than what was initially proposed as discussed:

```javascript
const manager = chrome.runtime.getBackgroundPage().manager;
const state = manager.getState();

manager.on('stateChanged', () => {
  const newState = manager.getState();
  // update the UI with the new state
});
```

I have added tests for the connection manager ... these are more like integation tests as they add real smoldots and then simulate apps connecting (to multiple networks) and disconnecting and then testing that the state changes as expected.

In the process I fixed #213 - the tests demonstrated that when an app has no in flight requests or subscriptions and disconnects then it gets stuck in a `disconnecting` state and doesn't clean up its state.

I also slightly tidied up the message types because I was using them .. the `Message` types in the extension are now shared between the content and background scripts (moved all into the top level types folder).


projects/extension/src/background/SmoldotMediator.ts
```diff
@@ -32,7 +32,7 @@ export class SmoldotMediator {
     const nextID = ++this.#id;
     /* eslint-disable @typescript-eslint/no-unsafe-member-access */
     message.id = nextID;
-    this.#smoldotClient.send_json_rpc(JSON.stringify(message));
+    this.smoldotClient.send_json_rpc(JSON.stringify(message));
```
wirednkod, line 35: if smoldotClient is not a private var then there is no reason of existance of  creating and calling sendRpcMessage(message: JsonRpcRequest): number. It can be called straight

projects/extension/src/background/index.ts
```diff
@@ -21,9 +27,15 @@ chrome.runtime.onInstalled.addListener(() => {
   const westendChainSpec = JSON.stringify(westend);
   const kusamaChainSpec = JSON.stringify(kusama);
   const polkadotChainSpec = JSON.stringify(polkadot);
-  manager.addSmoldot('westend', westendChainSpec).catch((e) => { console.error('Error creating westend smoldot: ', e); });
-  manager.addSmoldot('kusama', kusamaChainSpec).catch((e) => { console.error('Error Creating kusama smoldot', e); });
-  manager.addSmoldot('polkadot', polkadotChainSpec).catch((e) => { console.error('Error Creating polkadot smoldot', e); });
+  manager.addSmoldot('westend', westendChainSpec).catch((e) => {
```
wirednkod, line 30: Its probably better to create a map of name/specs and loop through it

wirednkod (changes requested):

projects/extension/src/background/AppMediator.test.ts
```diff
@@ -1,14 +1,16 @@
+import { ExtensionMessage } from '../types';
 import { AppMediator } from './AppMediator';
 import { MockPort, MockConnectionManager } from './mocks';
 import { JsonRpcResponse } from './types';
 
 describe('AppMediator setup', () => {
 
   it('initialises correctly', () => {
```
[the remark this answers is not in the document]
raoulmillais: Sure ... These aren't new tests, that's why I didn't touch them in this pr

projects/extension/src/background/SmoldotMediator.ts
raoulmillais: No because it depends on the internal id counter. I made it public so I could call terminate but in hindsight I think what would have been better would be to expose a wrapping method that calls terminate on the private client. Users of this class shouldn't be touching the inner client

projects/extension/src/background/index.ts
raoulmillais: I think this is nice and clear even if it's a bit repetitive and we don't intend to add more networks to this list

wirednkod: I still think that its better to loop through it

raoulmillais: It's s stylistic argument and it's got nothing to do with this pull request: the only change here was reformatting. I think this pull request is big enough already

wirednkod (approved):

PR #230, opened 2021-04-14
Implement disconnect on ExtensionProvider and test content script message routing

### LINKED ISSUE #168: Disconnect the port when `disconnect()` is called on the ExtensionProvider
Send a message to the content script telling it to disconnect the port to the background when `disconnect()` is called on the provider.  This allows the user to call `disconnect()` on the `ApiPromiise` they have and have it actually disconnect from the extension.  We should also set `isConnected` to false in the provider and emit the `disconected` event to fulfill the expectations of PolkadotJS concerning disconnection.

raoulmillais:
This implements disconnect in the extension provider (fixes #168) which will now send a message to the content script telling it to disconnect the port to the background and emits the `disconnected` event on the provider.

I took the opportunity to add a full suite of tests for the content script message routing behaviours.  I had to refactor the routing out to make it easier to test.

### DIFF
--- a/projects/extension/src/types/index.ts
+++ b/projects/extension/src/types/index.ts
@@ -59,7 +59,8 @@ export interface ExtensionProviderMessage {
     appName: string;
     chainName: string;
     origin: string;
-    message: AppMessage | 'disconnect';
+    action: 'forward' | 'disconnect';
+    message?: AppMessage;
   }
 }
 
--- a/projects/extension/src/types/index.ts
+++ b/projects/extension/src/types/index.ts
@@ -32,24 +32,34 @@ export type ExtensionMessageType = 'error' | 'rpc';
 
 export interface ExtensionMessage {
   type: ExtensionMessageType;
-  payload: string;
+  payload: string; // JSON encoded RPC response or an error message
 }
 
 // Messages that come from the app
 export type AppMessageType = 'associate' | 'rpc';
 
+/* The inner message that was received by the content script to be sent on to
+ * the extension background.
+ */
 export interface AppMessage {
   type: AppMessageType;
-  payload: string; // smoldot name / json / message_id / subscription_id
+  payload: string; // smoldot name or JSON encoded RPC message
   subscription?: boolean;
 }
 
-export interface Message {
+/* A message that is sent by the `ExtenionProvider` calling `window.postMessage`
+ * that is received by the content script.
+ *
+ * The `message` property will either be an `AppMessage` for sending through the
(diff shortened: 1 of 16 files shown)


projects/extension/src/types/index.ts
```diff
@@ -32,24 +32,34 @@ export type ExtensionMessageType = 'error' | 'rpc';
...
+ * port to the background script or the string 'disconnect' to tell the content
+ * script to disconnect the port.
+ */
+export interface ExtensionProviderMessage {
   data: {
     appName: string;
     chainName: string;
     origin: string;
-    message: AppMessage
+    message: AppMessage | 'disconnect';
```
wirednkod, line 62: This confuses the types a bit. You assign to message either a json or a string. Its best if 'disconnect' is part of AppMessageType

wirednkod (changes requested):

projects/extension/src/types/index.ts
raoulmillais: The AppMessage type is what gets passed unchanged to the extension.  Putting it there would make the logic for forwarding messages more complicated.  This information is to be used by the router only and not by the background. I could put it on the `data` property of `ExtensionProviderMessage` as a sibling of `message`?

raoulmillais: I.E. I could add an `action` property which would be one of `'forward' | 'disconnect'`.  This would describe the action that the router should take when processing this message.  The `message` will be empty for 'disconnect' and it should disconnect the port and it will be an `AppMessage` for 'forward' when it is either an associate or rpc message and the router should take the action of forwarding that message.  Does that cover your concerns adequately?

wirednkod: Afaiu the 'disconnect' is used as action/status rather than message in this case. If that is it then  a "flag" or different type of structure should return. if this is supposed to be the message then previous comment applies

wirednkod (changes requested):

projects/extension/src/types/index.ts
raoulmillais: Yup I agree :+1:

raoulmillais: > I.E. I could add an `action` property which would be one of `'forward' | 'disconnect'`. This would describe the action that the router should take when processing this message. The `message` will be empty for 'disconnect' and it should disconnect the port and it will be an `AppMessage` for 'forward' when it is either an associate or rpc message and the router should take the action of forwarding that message. Does that cover your concerns adequately?

I implemented this in the latest commit

wirednkod (approved):

PR #231, opened 2021-04-15
Fix extension provider malformed message warnings and bump @substrate/connect version

raoulmillais:
I have made the warnings more explicit.  Added a specific warning for the case where there is no `action` and standardised the formatting.  

I also bumped the @substrate/connect version with a minor version number change as it affects the API between @substate/connect and the demos. (Although [the version numbers don't mean anything](https://semver.org/#spec-item-4) for a major version of "0")

### DIFF
--- a/packages/connect/package.json
+++ b/packages/connect/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@substrate/connect",
-  "version": "0.2.0",
+  "version": "0.3.0",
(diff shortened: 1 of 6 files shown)


wirednkod:
Even though 0 version does not matter we should get used to following this:
0. | MAJOR version when you make incompatible API changes,
0. | MINOR version when you add functionality in a backwards compatible manner, and
0  | PATCH version when you make backwards compatible bug fixes. <----

wirednkod (approved):

PR #235, opened 2021-04-16
Handle errors sent from the extension background to the app properly

### LINKED ISSUE #228: The ExtensionProvider does not handle "error" messages from the extension
The extension can send back an "error" message to the content script which gets forwarded to the ExtensionProvider.  At the moment it blindly tries to parse the payload assuming it contains JSON.  The extension provider should check the message "type" for "error", stop processing and emit an "error" event for these messages.

raoulmillais:
This fixes #228.  In the `ExtensionMessageRouter` when we received a message through the port from the background we were only sending its `payload` via `window.postMessage` on to the app.  The background could send either an `rpc` message where the payload was a JSON string or an `error` where the payload would be an error message.  I adjusted the `ExtensionMessageRouter` to send the whole message and not just the `payload`.  I also adjusted the `ExtensionProvider` to check for `error` message types and emit an error event instead of trying to process it as if it was an RPC message.

### DIFF
--- a/packages/connect/package.json
+++ b/packages/connect/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@substrate/connect",
-  "version": "0.3.0",
+  "version": "0.3.1",
   "description": "Substrate-connect to Smoldot clients. Using either substrate extension with predefined clients or an internal smoldot client based on chainSpecs provided.",
   "author": "Parity Team <admin@parity.io>",
   "license": "GPL-3.0-only",
--- a/packages/connect/src/ExtensionProvider/ExtensionProvider.test.ts
+++ b/packages/connect/src/ExtensionProvider/ExtensionProvider.test.ts
@@ -8,7 +8,7 @@ import {
   ExtensionMessageData
 } from './types';
 
-function waitForMessageToBePosted(): Promise<null> {
+const waitForMessageToBePosted = (): Promise<null> => {
   // window.postMessge is async so we must do a short setTimeout to yield to
   // the event loop
   return new Promise(resolve => setTimeout(resolve, 10, null));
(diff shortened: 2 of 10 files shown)


packages/connect/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@substrate/connect",
-  "version": "0.3.0",
+  "version": "0.4.0",
```
wirednkod, line 3: Even though 0 version does not matter we should get used to following this:
0. | MAJOR version when you make incompatible API changes,
0. | MINOR version when you add functionality in a backwards compatible manner, and
0 | PATCH version when you make backwards compatible bug fixes. <----

packages/connect/src/ExtensionProvider/ExtensionProvider.test.ts
```diff
@@ -1,8 +1,27 @@
...
+import { 
+  ExtensionProvider, 
+} from './ExtensionProvider';
+import {
+  ProviderMessage,
+  ProviderMessageData,
+  ExtensionMessageData
+} from './types';
 
+function waitForMessageToBePosted(): Promise<null> {
```
wirednkod, line 11: use arrow function for consistency

wirednkod (changes requested):

packages/connect/package.json
raoulmillais: This is not backwards compatible.  The interface of the message sent between the extension and @substrate/connect has changed.  It is a backwards incompatible fix.

raoulmillais: hmmm actually I see what you mean ... the **public** API has not changed .. I.E. the interface of `Detector` so it should be a revision change.  i'll change it to 0.3.1

raoulmillais: this is interesting .. we need to think give some thought about the implicit interface of messages passed between the `ExtensionProvider` in @substrate/connect and the `ExtensionMessageRouter` in the extension.  When we go public with this we'll basically be locking ourself into that interface because changing it will be **very** hard. (actually impossible with the current design)

wirednkod (approved):

PR #237, opened 2021-04-16
Integration popup and options pages with background script

### LINKED ISSUE #95: Popup - create integration with background and actions
Tasks:
- [ ] Create listener for events emitted on open/new tab creation
- [ ] Create listener for events emitted on change of tab contents (add/remove uApp, network/uApp status)
- [ ] Create listener for events emitted on destroy/close tab
- [ ] Create API call integration for the state of uApps/tabs

wirednkod:
Fixes #95


projects/extension/src/containers/Popup.tsx
```diff
@@ -20,27 +28,61 @@ const SearchBar: React.FC = (): React.ReactElement => (
...
-        <Tab current tab={tabCurrent}/>
+        {activeTab}
         <SearchBar />
         <Box marginY={1}>
-          {tabsRest.map(tab => <Tab key={tab.uApp.name} tab={tab}/>)}
+          {rTabs.map(r => r)}
         </Box>
         <Divider />
-        <MenuButton fullWidth>My Networks</MenuButton>
+        <MenuButton fullWidth onClick={() => openInNewTab('chrome-extension://npfgeeonalpabakkpdobjngodgljemkb/options.html')}>My Networks</MenuButton>
```
raoulmillais, line 85: This link is not stable  .. I think the correct way to do this is `chrome.runtime.openOptionsPage()`

projects/extension/src/background/ConnectionManager.ts
```diff
@@ -34,6 +34,10 @@ export class ConnectionManager extends (EventEmitter as { new(): StateEmitter })
     return this.#smoldots.map(s => s.name);
   }
 
+  get apps(): AppMediator[] {
```
raoulmillais, line 37: why did you create a new way of iterating the apps and networks instead of fixing the `getState()` method to add the extra data you needed?

projects/extension/src/background/index.ts
```diff
@@ -23,21 +27,46 @@ export interface RequestRpcSend {
...
-    console.error('Error Creating polkadot smoldot', e); 
-  });
+  void createSmoldot('polkadot', polkadot);
+  void createSmoldot('kusama', kusama);
+  void createSmoldot('westend', westend);
 });
 
 chrome.runtime.onConnect.addListener((port) => {
-  manager.addApp(port);
+  if (port.name === 'substrateExtension') {
```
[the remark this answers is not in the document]
wirednkod: Tried that - does not seem to return instance data of the manager. We can revisit on Monday to have a look

projects/extension/src/background/ConnectionManager.ts
wirednkod: Didn't realise that getState is there. I will give it a try - though as per previous comment I tried it but I did not get the apps that were registered (0 were always returned). We can actually have a look on this also on Monday

projects/extension/src/containers/Popup.tsx
```diff
@@ -20,27 +26,62 @@ const SearchBar: React.FC = (): React.ReactElement => (
...
-  const tabCurrent = tabs[0];
-  const tabsRest = tabs.slice(1);
+  const [apps, setApps] = React.useState<AppMediator[]>([]);
+
+  React.useEffect((): void => {
+    chrome.runtime.getBackgroundPage(backgroundPage => {
+      const bg = backgroundPage as Background;
+      bg.manager && setApps(bg.manager.apps);
+
+      // TODO: change the state when stateChanged occurs
```
raoulmillais, line 39: I think this is as simple as moving `bg.manager && setApps(bg.manager.apps);` inside the event handler

projects/extension/src/background/types.ts
```diff
@@ -10,6 +10,15 @@ export interface InitAppNameSpec {
   chainSpec?: string
 }
 
+export type AppType = {
```
raoulmillais, line 13: I don't think you are using this type any more

projects/extension/src/containers/Popup.tsx
wirednkod: Please elaborate

raoulmillais (approved):

projects/extension/src/containers/Popup.tsx
wirednkod: It is a bit more complicated as the state change should alter the state of the app in the existing apps array and reset - so I will leave that for a later PR.

PR #240, opened 2021-04-20
Move all message types into separate protocol pkg

raoulmillais:
This PR removes duplicated and inconsistently named message types and adds documentation for them.  Since the types are shared between `@substrate/connect` and the extension they need to be in their own package.

It also adds some strongly typed helpers that wrap `window.postMessage` / `window.addEventListener` to to get type checking on both sides of the communication. 

**This is a fairly mechanical cleanup / refactoring - I am happy to wait to merge your work first and to resolve any merge conflicts if you're worried about that**


wirednkod (changes requested):
I would split this PR to 2. First to create the new package - publish it to npm and then make another PR with the changes to the ExtensionProvider and whoever uses it. If this is merged now (before npm publish) will break.

raoulmillais:
> I would split this PR to 2. First to create the new package - publish it to npm and then make another PR with the changes to the ExtensionProvider and whoever uses it. If this is merged now (before npm publish) will break.

This is an interesting exercise to walk through imagining we were to merge this PR as-is.  Let's say a user is using `@substrate/connect` at the last recent published version (which does not have these changes)... They will be using the compiled dist code at that version so there are no types involved when running their app.  Since there are no changes to the shape of the objects passed to `window.postMessage` or received from the extension it will continue to work regardless of whether they are running the extension before this commit or after this commit.

When developing their app and compiling against `@substrate/connect` - none of the types which were moved out are exported at the package root and I don't think there's any good reason for an app developer to be consuming any of the message types. 

Thinking about the changes from the perspective of the extension is easier since we don't yet publish it. The only way to run it is from a clone of the repo. Also all the changes are intentionally in a single commit.  
* You will either have all of them or none-of-them 
* There are no changes to the actual shape of the messages.  
* The new package is a yarn workspace so the fact is not yet published does not matter - yarn will use the repo local version.

Regarding publishing `@substrate/connect` - it is important we publish this new package **first** and then publish `@substrate/connect` I think what's missing from this PR is:
1. Including publishing guidelines in a README.  I.E. If you want to make future changes to the protocol then you have to publish a new version of the protocol before publishing a new version of @substrate/connect.
2. A version bump to the package.json for @substrate/connect

Of course much of this only holds because the new package is (almost) only types which compile away

raoulmillais:
@wirednkod I think I addressed your comment above.  I have made the changes I said I would and have rebased and fixed the conflicts from master (since your last PR merge).

I've looked around and I think that there isn't a better solution to the `prelint` script I added.  I think it does also makes sense ... we have to generate the typing file (by running tsc) in order for eslint to be able to accurately lint connect and the extension. In a monorepo there is no way to avoid that.

I think what made it feel odd at first is the idea of 'compile before lint', but we are actually compiling a dependent project and that is more sane.

wirednkod (approved):

PR #244, opened 2021-04-23
introduce clean command for removing all node modules

wirednkod:


packages/connect-extension-protocol/package.json
```diff
@@ -5,6 +5,7 @@
   "main": "dist/index.js",
   "type": "module",
   "scripts": {
+    "deep-clean": "yarn clean && rm -rf node_modules",
```
raoulmillais, line 8: Would it not be better to have the `yarn clean` happen as a `predeep-clean` script?

wirednkod: Could be - at the end of the day its the same thing though

raoulmillais (approved):

PR #245, opened 2021-04-23
fix order of build workspaces

wirednkod:


raoulmillais (approved):

raoulmillais:
I think we ought to review build orderings and workspaces and using typescript project references in light of that information I discovered the other days.  I'm sure we've still not got the optimal setup. It's not super urgent though .. I'll create a separate tech debt ticket for that

PR #254, opened 2021-04-27
Remove associate message and ready state

### LINKED ISSUE #242: Remove the `associate` message
Since we now have the network name in the port name when an app connects the port, the associate message isn't really needed any more. We can associate the app with a smoldot client straight away.

### Why?

* It would simplify the types in the message passing - one less type of message.
* The chain (network) name is supplied only once - not in the payload of an `associate` message too
* It would clean up & make the protocol between `ExtensionProvider` and the `ExtensionMessageRouter` more consistent
* No more special case handling in the `AppMediator` for associate messages - it only has to handle connections / disconnections and RPC messages.
* The size of the setup in a lot of the unit tests will be much smaller making them easier to read.
* We no longer need the intermediate state where an `AppMediator` is connected but not ready.  When it is connected it is ready. All the logic to detect misuse in that state can be removed.
* There are no longer 2 stateChanged events one quickly after the other as an app connects and then becomes associated.

### How?

* [x] The message from the `ExtensionProvider` to the `ExtensionMessageRouter` should change with a new `connect` action (mirroring the `disconnect` action).  
* [x] The `ExtensionMessageRouter` changes to handle this action which initiates the connection (like the associate message did - but does not send the extra message).
* [x] The `MessageToManager` changes to no longer have the `associate` type.
* [x] The `AppMediator` associates the app with a network as soon as the port is connected by parsing the network name from the port name and it changes to `connected` (which is implicitly also ready) immediately.
* [x] Remove the handling of the associate message from the `AppMediator` - keep the association logic but do it on connection (see last point).
* [x] Remove the superflous `ready` state - a `connected` app is also ready.
* [ ] Bump @substrate/connect and @substrate/connect-extension-protocol

### Implications

It would be a breaking change in the protocol

raoulmillais:
This fixes #242 - significantly simplifying the message parsing and handling on the extension side.

When a user has the extension installed and an app connects the `ExtensionProvider` now sends a message with the type `connect` instead of an `associate` message.  The `ExtensionMessageRouter` in turn connects to the background as before but no longer sends an `associate` message.  The `AppMediator` parses the network name out of the port name and becomes `connected`. 

If the background does not have a client for the network named it sends an error back to the app (as before) and disconnects the app.  If the port name is in the wrong format the background also sends an error back (this shouldn't ever happen as we control the content script).

There is no need for a separate `ready` state as the app is implicitly ready if associating the app as part of a port connecting happens successfully.

### DIFF
--- a/projects/extension/src/background/ConnectionManager.ts
+++ b/projects/extension/src/background/ConnectionManager.ts
@@ -117,8 +117,7 @@ export class ConnectionManager extends (EventEmitter as { new(): StateEmitter })
       port.postMessage({ type: 'info', payload: `App ${port.name} already exists.` });
     } else {
       const newApp = new AppMediator(port, this as ConnectionManagerInterface)
-      const didAssociate = newApp.associate();
-      if (didAssociate) {
+      if (newApp.associate()) {
         this.#apps.push(newApp);
         newApp.on('stateChanged', () => this.emit('stateChanged'));
         this.emit('stateChanged');
--- a/projects/extension/src/background/ConnectionManager.ts
+++ b/projects/extension/src/background/ConnectionManager.ts
@@ -117,9 +117,12 @@ export class ConnectionManager extends (EventEmitter as { new(): StateEmitter })
       port.postMessage({ type: 'info', payload: `App ${port.name} already exists.` });
     } else {
       const newApp = new AppMediator(port, this as ConnectionManagerInterface)
-      this.#apps.push(newApp);
-      newApp.on('stateChanged', () => this.emit('stateChanged'));
-      this.emit('stateChanged');
+      const didAssociate = newApp.associate();
+      if (didAssociate) {
+        this.#apps.push(newApp);
+        newApp.on('stateChanged', () => this.emit('stateChanged'));
+        this.emit('stateChanged');
+      }
     }
   }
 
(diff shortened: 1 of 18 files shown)


projects/extension/src/background/ConnectionManager.ts
```diff
@@ -117,9 +117,12 @@ export class ConnectionManager extends (EventEmitter as { new(): StateEmitter })
       port.postMessage({ type: 'info', payload: `App ${port.name} already exists.` });
     } else {
       const newApp = new AppMediator(port, this as ConnectionManagerInterface)
-      this.#apps.push(newApp);
-      newApp.on('stateChanged', () => this.emit('stateChanged'));
-      this.emit('stateChanged');
+      const didAssociate = newApp.associate();
```
wirednkod, line 120: no need to assign variable and then check against it. you can just:
```
if (newApp?.associate()) {
```

wirednkod (changes requested):
There are also lint checks failing on the branch

projects/extension/src/background/ConnectionManager.ts
raoulmillais: True.. this is a remnant of refactoring

wirednkod: Yeap. Not a blocker for merging. I requested changed for the lint issues mainly

wirednkod (approved):

PR #257, opened 2021-04-28
Fix race condition that was not injecting the script in the page on time

wirednkod:


raoulmillais (approved):
We should add a unit test for this. It's quite easy to test

Issue #249, opened 2021-04-23
Review build dependencies and ordering


raoulmillais:
It feels like we are not using yarn workspaces + scripts + typescript quite right.  We are constantly tweaking / finding issues with the build or introducing susrprising workarounds.  I've created this issue to document the issues we've run into and to catalog investigations so far. Ultimately I'd like to feel confident we have a solid repeateable and idiomatic approach and implement any necessary changes so that we can stop monkeypatching things.

### Tweaks / Issues and Surprising workarounds

* [Adding smoldot-test-utils as a workspace](https://github.com/paritytech/substrate-connect/commit/6eb4e8091f8b493c1ae40b76e365eef5e38ad8b4)
* [Having to compile protocol to get types before linting](https://github.com/paritytech/substrate-connect/commit/07ff6d6524096efc41ba19cb932b00cec612da28)

### Upstream bugs

* [Yarn doesn't add local workspace dependency if version is 'latest'](https://github.com/yarnpkg/yarn/issues/4878)
* [Typescript eslint doesn't support typescript projects](https://github.com/typescript-eslint/typescript-eslint/issues/2094)

### Research and potential solutions to investigate further

* [Start using typescipt project references](https://stackoverflow.com/a/58185075)
* Upgrade to yarn 2?
* Switch to npm workspaces?

wirednkod:
@raoulmillais I believe we should stop using workspaces (as I think it does not fit in our needs) and split packages and projects to separate repos.

raoulmillais:
> @raoulmillais I believe we should stop using workspaces (as I think it does not fit in our needs) and split packages and projects to separate repos.

If I want to make a change or addition to how substrate-connect communicates with the extension: I must make changes in 3 npm packages:


`@substrate/connect` <--> `@substrate/connect-protocol-extension` <--> `extension`

I want to be able to change those 3 things *atomically*.  Having them in separate repos adds a lot of administrative burden:  
* I have to open 3 separate pull requests. 
* I have to explain in the pull request about which other specific pull requests are relevant for this changeset.
* I probably have ensure the pull requests are merged and the packages published in a specific order in order for the CI to remain green.

It also adds quite a bit of cognitive for the reviewer:
* The reviewer has to keep the 3 different pull requests in mind.
* The review no longer happens in a single "view" on github.
* The reviewer cant simply checkout the changeset anymore to run the changes for themselves before reviewing.

[...]

PR #264, opened 2021-05-06
Remove yarn [WIP]

wanderer:
This attempts to tackle #249 by removing yarn and use npm workspaces in place of it. Currently building works but tests are failing.


wirednkod (approved):
lgtm but lets have a chat if we want at this point to differentiate from every other project (afaik) of Parity that uses (for some reason) yarn

raoulmillais:
I think npm 7 which introduced workspaces first shipped with node 15

PR #272, opened 2021-05-13
Fix build of substrate connect and testing

wirednkod:
Fix the build structure to be more flat than before by removing unnecessary files that should not be included in the built.
Replaced the ts-jest with babel-jest but eventually that was removed as well from the package.
In addition paths where needed, altered to include extension in order to support both cjs and esm.


packages/connect/package.json
```diff
@@ -4,12 +4,12 @@
   "description": "Substrate-connect to Smoldot clients. Using either substrate extension with predefined clients or an internal smoldot client based on chainSpecs provided.",
   "author": "Parity Team <admin@parity.io>",
   "license": "GPL-3.0-only",
+  "type": "module",
```
wanderer, line 7: did the position of this change for a reason?

packages/connect/tsconfig.test.json
```diff
@@ -1,14 +1,16 @@
 {
     "compileOnSave": true,
     "compilerOptions": {
-        "target": "ES2018",
+        "target": "ESNEXT",
```
wanderer, line 4: why next and not 2020?

wanderer (approved):

packages/connect/package.json
wirednkod: Absolutely no reason :)

packages/connect/tsconfig.test.json
wirednkod: For now - ESNEXT and ES2020 are same - but to be on safe side I will change that to ES2020

PR #275, opened 2021-05-14
Add ApiOptions to be optional provided on Detector

wirednkod:


tomaka:
Random question: why not return the provider and let the user create the `ApiPromise` themselves?

wirednkod:
> Random question: why not return the provider and let the user create the `ApiPromise` themselves?

That was a discussion we had long time ago with @raoulmillais .
AFAI remember, we concluded that its better to kind of "ease" the work of the developer by providing him the API (no need to install Substrate-Connect AND polkadotApi in order to get things working. 
By installing just substrate-connect (which eventually returns the API) he can be up and running with just 2-3 lines of code

raoulmillais:
> > Random question: why not return the provider and let the user create the `ApiPromise` themselves?
> 
> That was a discussion we had long time ago with @raoulmillais .
> AFAI remember, we concluded that its better to kind of "ease" the work of the developer by providing him the API (no need to install Substrate-Connect AND polkadotApi in order to get things working.
> By installing just substrate-connect (which eventually returns the API) he can be up and running with just 2-3 lines of code

This was the brief from Gav..a one-liner to detect and go: https://www.notion.so/paritytechnologies/Briefings-b6dcec6e6d984d9a918952abe97a5ece

raoulmillais:
> > > Random question: why not return the provider and let the user create the `ApiPromise` themselves?
> > 
> > 
> > That was a discussion we had long time ago with @raoulmillais .
> > AFAI remember, we concluded that its better to kind of "ease" the work of the developer by providing him the API (no need to install Substrate-Connect AND polkadotApi in order to get things working.
> > By installing just substrate-connect (which eventually returns the API) he can be up and running with just 2-3 lines of code
> 
> This was the brief from Gav..a one-liner to detect and go: https://www.notion.so/paritytechnologies/Briefings-b6dcec6e6d984d9a918952abe97a5ece

I don't think it would be a bad idea to also export the providers to give users flexibility. It used to be like that at one point I'm not sure why it changed

tomaka:
> This was the brief from Gav..a one-liner to detect and go: https://www.notion.so/paritytechnologies/Briefings-b6dcec6e6d984d9a918952abe97a5ece

This doesn't necessarily mean that this one-liner should be in substrate-connect.
It is actually wrong to create an `ApiPromise` without any types. Half of the features of PolkadotJS will return errors if you pass `null` for the types. In other words this one-liner as it presently is in substrate-connect won't properly work.

raoulmillais:
But that's what this pull request addresses... You can now supply the options which includes the custom types. Or smi missing something?

packages/connect/src/Detector.ts
```diff
@@ -41,8 +42,20 @@ export class Detector {
       throw new Error(`No known Chain was detected and no chainSpec was provided. Either give a known chain name ('${Object.keys(this.#chainSpecs).join('\', \'')}') or provide valid chainSpecs.`)
     }
     await provider.connect();
+    let derives, initWasm, metadata, registry, rpc, signedExtensions, signer, source;
```
raoulmillais, line 45: Why explicitly name all these properties and copy them? Wouldn't it be neater to just use Object.assign?

tomaka:
> But that's what this pull request addresses... You can now supply the options which includes the custom types. Or am I missing something?

Well, my point is that it's no longer a one-liner anyway, so you might as well just return the `Provider` rather than the full `ApiPromise`.

packages/connect/src/Detector.ts
wirednkod: You mean something like this:
```
return await ApiPromise.create({ provider, ...Object.assign({}, options) });
```
?

raoulmillais: you want `provider` not to be overridden even if its supplied by the user so:

```javascript
return await ApiPromise.create(Object.assign(options, {provider}));
```

raoulmillais:
It is you're just also providing `options`.  I guess the point is do we really want the end user to care about providers?  It's not something that they normally have to deal with in polkadotjs.   I do like the idea of having them exported too so if they want to do more custom things they can.  

I.E. For the simple cases of "I want to connect to polkadot" .. it should be simple

```javascript
const api = await new Detector('my-app').detect('polkadot');
```

For a custom chain with custsom types, they do:

```javascript
const api = await new Detector('my-app').detect('mySubstrateChainName', chainSpec(), { types: { /* ... */ } });
```

For total flexibility they can use the providers and `ApiPromise` directly

packages/connect/src/Detector.ts
wirednkod: That is not correct. If options are not provided this will lead to error

raoulmillais (changes requested):
Can you change one of the the tests so that it takes some options please? Then we'll have that code path tested too

raoulmillais:
Just need to update the README and I think this is good to merge :+1:

raoulmillais (approved):

PR #277, opened 2021-05-19
Upgrade polkadot api

wirednkod:
Upgraded as needed and fix tests. Tests of extension (mock jest) are failing for now.
@raoulmillais please have a look


wirednkod:
Includes implementation of #278

raoulmillais:
The AppMediator tests are failing (not due to mock jest) ... after these changes we're now hitting a branch which should never be hit.

```bash
      ✓ ProcessSmoldotMessage: remaps the id to the apps id

  console.error
    Asked a disconnected UApp (test-app::westend) to process a message from undefined

      94 |       // Shouldn't happen - we remove the AppMediator from the smoldot's apps
      95 |       // when we disconnect (below).
    > 96 |       console.error(`Asked a disconnected UApp (${this.name}) to process a message from ${this.#smoldotName as string}`);
         |               ^
      97 |       return false;
      98 |     }
      99 |

      at AppMediator.processSmoldotMessage (src/background/AppMediator.ts:96:15)
      at Object.<anonymous> (src/background/AppMediator.test.ts:139:29)
```

packages/connect/src/Detector.test.ts
```diff
@@ -1,103 +1,59 @@
...
-      const options = {} as ApiOptions;
-      const api = await detect.connect(chainName, chainSpec, options);
-      expect(api).toBeTruthy();
-      done();
-    } catch (err: unknown) {
-      done(err);
-    } finally {
-      await detect.disconnect('westend2');
-    }
+  // DEACTIVATED for now due to chainSpecs - once Kusama specs are available this should be activated
```
raoulmillais, line 35: This says DEACTIVATED but the test is not ignored.

wirednkod:
> The AppMediator tests are failing (not due to mock jest) ... after these changes we're now hitting a branch which should never be hit.
> 
> ```shell
>       ✓ ProcessSmoldotMessage: remaps the id to the apps id
> 
>   console.error
>     Asked a disconnected UApp (test-app::westend) to process a message from undefined
> 
>       94 |       // Shouldn't happen - we remove the AppMediator from the smoldot's apps
>       95 |       // when we disconnect (below).
>     > 96 |       console.error(`Asked a disconnected UApp (${this.name}) to process a message from ${this.#smoldotName as string}`);
>          |               ^
>       97 |       return false;
>       98 |     }
>       99 |
> 
>       at AppMediator.processSmoldotMessage (src/background/AppMediator.ts:96:15)
>       at Object.<anonymous> (src/background/AppMediator.test.ts:139:29)
> ```

I am not sure I understand what you mean here. The test "ProcessSmoldotMessage: does nothing when it has sent no requests" is passing successfully.  Can you elaborate?

packages/connect/src/Detector.test.ts
wirednkod: afair kusama was fixed on the mean time - and test started to succeeded. Just did not remove the comment (which is now removed and test still pass)

raoulmillais:
```
       94 |       // Shouldn't happen - we remove the AppMediator from the smoldot's apps
       95 |       // when we disconnect (below).
     > 96 |       console.error(`Asked a disconnected UApp (${this.name}) to process a message from ${this.#smoldotName as string}`);
```

See the comment .. we've started hitting a codepath that is not supposed to ever be hit.  It's not actually resulting in a test failure but that's because we don't want to tank the entire extension in this scenario.  The fact that `console.error` is being run means that there is almost certainly a new bug somewhere.

wirednkod:
> ```
>        94 |       // Shouldn't happen - we remove the AppMediator from the smoldot's apps
>        95 |       // when we disconnect (below).
>      > 96 |       console.error(`Asked a disconnected UApp (${this.name}) to process a message from ${this.#smoldotName as string}`);
> ```
> 
> See the comment .. we've started hitting a codepath that is not supposed to ever be hit. It's not actually resulting in a test failure but that's because we don't want to tank the entire extension in this scenario. The fact that `console.error` is being run means that there is almost certainly a new bug somewhere.

There is no error on the code that you are pointing. Have a look at the unit tests (specifically the unit test: "ProcessSmoldotMessage: does nothing when it has sent no requests"). That test is actually testing the case that app is disconnected and a new message is tried to be processed. `expect(result).toBe(false);` <- that is why the test is passing

raoulmillais:
> > ```
> >        94 |       // Shouldn't happen - we remove the AppMediator from the smoldot's apps
> >        95 |       // when we disconnect (below).
> >      > 96 |       console.error(`Asked a disconnected UApp (${this.name}) to process a message from ${this.#smoldotName as string}`);
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > See the comment .. we've started hitting a codepath that is not supposed to ever be hit. It's not actually resulting in a test failure but that's because we don't want to tank the entire extension in this scenario. The fact that `console.error` is being run means that there is almost certainly a new bug somewhere.
> 
> There is no error on the code that you are pointing. Have a look at the unit tests (specifically the unit test: "ProcessSmoldotMessage: does nothing when it has sent no requests"). That test is actually testing the case that app is disconnected and a new message is tried to be processed. `expect(result).toBe(false);` <- that is why the test is passing

Aah, it's because you've changed that test so that it now calls `appMed.disconnect()` before trying to process the message.  So it's now printing an error because the test is asking a disconnected AppMediator to process a message, which is indeed a case that should never happen.  Can you remember why you added the call to `disconnect()`? ... it shouldn't be there - the test is checking that if we ask an AppMediator that is connected but has not sent any messages to process a message that it doesn't process the message.

wirednkod:
> Can you remember why you added the call to `disconnect()`? it shouldn't be there - the test is checking that if we ask an AppMediator that is connected but has not sent any messages to process a message that it doesn't process the message.

I have added a lot of unit test cases in order to increase the code test coverage. One of the scenarios that needs to be unit tested is also  this (app is disconnected but messages passes through). I have not altered or deleted any of the existing tests.

raoulmillais:
> I have added a lot of unit test cases in order to increase the code test coverage. 

:tada: 

> I have not altered or deleted any of the existing tests.

[This is the test that produces the error](https://github.com/paritytech/substrate-connect/pull/277/files#diff-656082cbc13d24ae7661691529bc0db322f5c27a4980b9c75c49ee4b495c818fR135) but the test name has no mention of the fact that it actually tests asking a disconnected `AppMediator` to process an incoming smoldot message.  [Previously that test did not call `disconnect`](https://github.com/paritytech/substrate-connect/blob/0c223f4a22e99af1cd48142590a641ef064366c7/projects/extension/src/background/AppMediator.test.ts#L58-L66) which is why we are now seeing the error stack trace in the test results.

If it is now the intention of the test to test disconnected apps being asked to process messages then the test name should reflect that.  Also since we are *expecting* this test to print an error to the console, we should be spying on `console.error` to 1) check the message is correct and 2) prevent a stack trace with an error being reported in the test results.

I think we should also keep around a test for asking an `AppMediator` to process an incoming message when it has not sent any requests.  (basically the same test but no call to `disconnect` and no spying on `console.error`)

These are 2 separate and valid behaviours to have test cases for.

projects/extension/src/background/AppMediator.test.ts
```diff
@@ -120,67 +42,174 @@ function setupAppMediatorWithSubscription(am: AppMediator, port: MockPort, appID
     .toEqual({ appIDForRequest, subID, method: 'system_health' });
 
   // should send the acknowledgement of the subscription request back to the UApp
-  const msgCalls = port.postMessage.mock.calls;
-  const lastMsg = msgCalls[msgCalls.length - 1][0] as MessageFromManager;
-  expect(lastMsg.payload).toEqual(`{"id":${appIDForRequest},"jsonrpc":"2.0","result":${subID}}`);
+  // const msgCalls = port.postMessage.mock.calls; // Replaced with the line below
```
raoulmillais, line 45: Remove the commented code

raoulmillais (approved):
This looks good to merge :+1:

Issue #286, opened 2021-05-28
Add extra checks and relevant tests in SmoldotProvider 


wirednkod:
As of now we are checking only for health peers to be > 0 in order to set the SmoldotProvider connected.

In order to avoid errors that appear on client extra checks should take place.
We need to also check for `health.shouldHavePeers` (in case peers.count === 0 then shouldHavePeers will be false in order to indicate that it's normal that peers is equal to 0, If shouldHavePeers is true and peers is 0, that means that you are not connected to the Internet or something. shouldHavePeers is always true for Westend, Kusama and Polkadot.

In addition a check must take place on health.isSyncing that should be false (denoting that Grandpa sync is finished).

Without these checks - there are errors appearing on UI when connected to smoldot, relevant to the description above

PR #290, opened 2021-06-07
Make mocking system_health more flexible

raoulmillais:
For this pull request I've only achieved some of what I'd hoped but the key change here is that you are now able to specify a sequence of health responses when testing your new 'connected' behaviour (previously you could only control the 'hasPeers' property of the health responses).

In the process I also reorganised and documented all of smoldot-test-utils and refactored some things to make them more readable.  I also enabled linting and fixed all the problems.  

You might find it easier to review this pull request by reviewing each commit one-by-one instead of viewing the "Files Changed" for the whole PR. I was careful to do each refactoring in a small step and to commit after (although I did retrospectively fix a couple of errors I'd missed as I went).  

### Future work

My vision with smoldot-test-utils is to eventually make it also provide a fake polkadotJS API instance that is pre-configured to respond with whatever responses you want so that you can integration test how an app behaves when the API gives known responses without having to mock the network.  I'm actually quite close to being able to do that, but I need to do some further refactoring to:

* unify the healthResponder and responders.  I.E. make it so that you can tell it what responses it should reply with for what request types (have different sequences of responses for different request methods) 
* fix the subscription mock responses so that they are more configurable (allow more control than sending a single subscription message)

In doing those 2 things I will also be able to make the code drastically more understandable than it currently is.  Unfortunately those changes require more invasive public API changes and I think it's better to wait until your test refactorings are merged before attempting that so as not to cause merge hell.

### DIFF
--- a/packages/connect/src/SmoldotProvider/SmoldotProvider.test.ts
+++ b/packages/connect/src/SmoldotProvider/SmoldotProvider.test.ts
@@ -5,7 +5,6 @@
 /* eslint-disable @typescript-eslint/no-unsafe-member-access */
 /* eslint-disable @typescript-eslint/no-unsafe-assignment */
 /* eslint-disable @typescript-eslint/no-unused-vars */
-import sinon from 'sinon';
 import { SmoldotProvider } from './SmoldotProvider';
 import { SmoldotClient, SmoldotOptions } from 'smoldot';
 import {
@@ -14,7 +13,6 @@ import {
   devChainHealthResponder,
   mockSmoldot,
   smoldotSpy,
-  SystemHealth,
   respondWith
 } from '@substrate/smoldot-test-utils';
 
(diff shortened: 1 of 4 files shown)


packages/smoldot-test-utils/src/index.ts
```diff
@@ -1,69 +1,152 @@
...
+    return systemHealthReponse(id, health);
   };
 }
 
-// dev chains never have peers
-const devChainHealthResponse = (id: number) => {
-  return `{"jsonrpc":"2.0","id":${id} ,"result":{"isSyncing":true,"peers":0,"shouldHavePeers":false}}`;
+// Orchestrates an `RpcResponder` with a queue of mock reponses to respond with.
+export const respondWith = (jsonResponses: string[]) => {
+  return (_: string): string => {
```
wirednkod, line 122: Following up our conversation yesterday,  I am not sure why there is a need for this `return (_: string): string` to exist at all. Couldn't that be: `(): string` ? 

Not a blocker ofc.

wirednkod (approved):
lgtm overall.  Just 1 non-blocker comment about the usage of _

wirednkod:
This also fixes #286 - right?

packages/connect/src/SmoldotProvider/SmoldotProvider.test.ts
```diff
@@ -126,15 +134,13 @@ test('emits connect and never emits disconnect for development chain', async ()
 test('send formats JSON RPC request correctly', async () => {
   // we don't really care what the reponse is
   const responses =  ['{ "id": 1, "jsonrpc": "2.0", "result": "success" }'];
-  const rpcSend = sinon.spy();
+  const rpcSend = jest.fn();
```
wirednkod, line 137: In order to use jest here you need to import it. There are 2 cases that `ReferenceError: jest is not defined` in this file

wirednkod (changes requested):

raoulmillais:
> This also fixes #286 - right?

No it makes it 

> This also fixes #286 - right?

No you closed the pr that implements that change. This PR makes it possible to test it easily if you reopen it

packages/smoldot-test-utils/src/index.ts
raoulmillais: I configured the rule to allow underscores as ignored param names

packages/connect/src/SmoldotProvider/SmoldotProvider.test.ts
raoulmillais: Fixed

PR #291, opened 2021-06-08
Add tests for debug.ts and utils.ts

wirednkod:
In order to increase coverage and have a start point for future utils that may be introduced (plus the fact that these tests were added in different PR that was solving another issue - thus they were ready), I add them as separate PR here

### DIFF
--- a/projects/extension/src/utils/debug.test.ts
+++ b/projects/extension/src/utils/debug.test.ts
@@ -1,9 +1,16 @@
 import { debug } from './debug';
 import { jest } from '@jest/globals';
 
+const PREV_ENV_VAL = process.env;
+
 beforeEach(() => {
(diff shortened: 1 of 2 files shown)


projects/extension/src/utils/debug.test.ts
```diff
@@ -0,0 +1,25 @@
+import { debug } from './debug';
+import { jest } from '@jest/globals';
+
+beforeEach(() => {
+  jest.resetModules();  
+  process.env = { NODE_ENV: 'development' };
```
raoulmillais, line 6: Should probably reset the environment after the tests to avoid surprises in other tests that depend on it

raoulmillais (changes requested):

PR #297, opened 2021-06-08
Add SmoldotMediator tests

wirednkod:
Add SmoldotMediator.test.ts file that covers the unit tests of SmoldotMediator

### DIFF
--- a/projects/extension/src/background/SmoldotMediator.test.ts
+++ b/projects/extension/src/background/SmoldotMediator.test.ts
@@ -5,7 +5,7 @@
 /* eslint-disable @typescript-eslint/no-unsafe-member-access */
 /* eslint-disable @typescript-eslint/no-unsafe-assignment */
 import { jest } from '@jest/globals';
-import * as smoldot from 'smoldot';
+import { SmoldotClient} from 'smoldot';
 
 import {
   customHealthResponder,
@@ -19,7 +19,7 @@ import westend from '../../public/assets/westend.json';
 import { AppMediator } from './AppMediator';
 import { MockPort, MockConnectionManager } from '../mocks';
 
-let sc: smoldot.SmoldotClient;
+let sc: SmoldotClient;
 let sm: SmoldotMediator;
 let appMed: AppMediator;
 
--- a/projects/extension/src/background/SmoldotMediator.test.ts
+++ b/projects/extension/src/background/SmoldotMediator.test.ts
@@ -6,6 +6,13 @@
 /* eslint-disable @typescript-eslint/no-unsafe-assignment */
 import { jest } from '@jest/globals';
 import * as smoldot from 'smoldot';
+
+import {
+  customHealthResponder,
+  smoldotSpy,
+  respondWith
+} from '@substrate/smoldot-test-utils';
+
 import { SmoldotMediator } from './SmoldotMediator';
 import { JsonRpcRequest, JsonRpcResponse } from './types';
 import westend from '../../public/assets/westend.json';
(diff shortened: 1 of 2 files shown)


projects/extension/src/background/SmoldotMediator.test.ts
```diff
@@ -0,0 +1,74 @@
...
+let sc: smoldot.SmoldotClient;
+let sm: SmoldotMediator;
+let appMed: AppMediator;
+
+let spyScSendJsonRpc: unknown;
+let spyScTerminate: unknown;
+
+beforeAll(async () => {
+  // init smoldot and SmoldotMediator
+  sc = await (smoldot as any).start({
```
raoulmillais, line 24: We don't have to use a real smoldot here. It would be better to use a fake from the test-utils as the tests will run faster

raoulmillais (changes requested):

projects/extension/src/background/SmoldotMediator.test.ts
```diff
@@ -0,0 +1,74 @@
...
+});
+
+test('Test removeApp', () => {
+  sm.removeApp(appMed);
+  expect(sm.apps).toHaveLength(0);
+});
+
+test('Test sendRpc', () => {
+  const message: JsonRpcRequest = {
+    id: 1,
```
raoulmillais, line 60: You should test with a number other than 1 to check the I'd remapping works correctly

projects/extension/src/background/SmoldotMediator.test.ts
```diff
@@ -0,0 +1,104 @@
...
+
+const createMessage = (id: number): JsonRpcRequest => ({
+  id: id,
+  jsonrpc: '2.0',
+  method: 'something',
+  params: []
+});
+
+beforeAll(async () => {
+  const healthResponses = [
```
raoulmillais, line 31: You don't need to use `customHealthResponder` and supply a sequence of healthy reponses when using `spoldotSpy`. If you omit the last argument it defaults to using an internal `healthyResponder` that takes care of responding with a healthy response each time it gets a "system_health" request.

projects/extension/src/background/SmoldotMediator.test.ts
```diff
@@ -0,0 +1,104 @@
...
+
+test('Test correctness of sendRpcMessage', () => {
+  // at this point this.#id is 0; create a message with id: 1 and expect to receive id = 1
+  expectedResponse = 1;
+  message = createMessage(1);
+  idReturned = sm.sendRpcMessage(message); // now on return it must be 1
+  expect(spyScSendJsonRpc).toHaveBeenCalledTimes(1);
+  expect(spyScSendJsonRpc).toHaveBeenCalledWith(`{"id":1,"jsonrpc":"2.0","method":"something","params":[]}`, 0);
+  expect(idReturned).toBe(1);
+  expect(spyScSendJsonRpc).toHaveBeenNthCalledWith(1, "{\"id\":1,\"jsonrpc\":\"2.0\",\"method\":\"something\",\"params\":[]}", 0);
```
raoulmillais, line 79: This is a duplicate assertion.  Line 77 checks the same thing

projects/extension/src/background/SmoldotMediator.test.ts
```diff
@@ -0,0 +1,104 @@
...
+});
+
+test('Test remapping of id on sendRpcMessage', () => {
+  // at this point this.#id is 1; create a message with id: 2 and expect to receive id = 2
```
raoulmillais, line 83: You've coupled the 2 tests together by setting up the fake smoldot `beforeAll`.  If you add or reorder tests they will break unexpectedly (order of calls will change as will the order of the ids in the tests).  Tests shouldn't interact / depend on one another ideally.  I think you should setup the fake smoldot afresh in a `beforeEach`, this will remove the annoying need to have to specifiy which "nth" the calls are on the spies because the tests can be self-contained with less reliance on the shared global variables and the order in which they are run.

raoulmillais (changes requested):

projects/extension/src/background/SmoldotMediator.test.ts
wirednkod: Even though I do not think that this invalidates the test - i have removed it as you suggested

raoulmillais: it doesn't but its a bit neater (less lines of code to distract from the essence of what we are testing - none of these tests have anything to do with health reponses).  Also I wasn't sure if you'd missed that it was possible to do this.

raoulmillais (approved):

PR #316, opened 2021-06-10
update landing texts

goldsteinsveta:


raoulmillais:
I'm happy to merge this, we can update it again if needed. It's considerably better than what we had before. Or would you rather get the ok from marcomms?

wirednkod (approved):

PR #324, opened 2021-06-14
Update smoldot 0.2.6 and pin parcel to 2.0.0-beta2

### LINKED ISSUE #322: Pass forbidWss to Smoldot from Smoldot and Extension Providers
Pass `forbidWss: true` when smoldot is directly embedded in a webpage and leave it to `false` when in the extension.

raoulmillais:
I tried to update parcel to remove the security vulnerabilities.  Unfortunately trying the following failed:
- parcel 2.0.0-beta3.1 is broken with "unreachable statement after
  return" errors in the Buffer shim (needs investigation)
- newest nightly crashes with a v8 stack trace

Instead I have pinned to 2.0.0-beta2 which is known good and working with our demos.

I updated smoldot to v0.2.6 and now pass `forbidWs` in the `SmoldotProvider` to suppress the console warnings when smoldot is embedded in a page.  I haven't added a test for this on purpose.  It's not really properly testable easily (check that no ws connections are made) and I don't think there's value in just checking that the flag is passed to `start` because it is only there to suppress these warnings.

This fixes #322


wirednkod (approved):
Maybe bump also substrate-connect version?

PR #330, opened 2021-06-14
Remove local network. Show wallet connected on light Client

### LINKED ISSUE #312: Remove everything except smoldot connection to westend from burnr
This might just be a text issue but it isn't clear to the user how to switch back to using @substrate/connect in the burnr wallet. See also #311

wirednkod:
Fixes #312


projects/burnr/src/utils/constants.ts
```diff
@@ -80,57 +74,18 @@ export const JS_WASM_PROVIDERS: Record<string, LazyProvider> = {
  * These fallback providers connect to a centralized remote RPC node.
  */
 export const REMOTE_PROVIDERS: Record<string, LazyProvider> = {
```
raoulmillais, line 76: Does it make sense for this to be a `Record<T, U>` at all anymore?  There's now only one provider and I think there will only ever be one provider.  If that's true why not go all the way and refactor to account for that?  It's a bit more work but it will tidy things up and make the code more obvious.

raoulmillais (changes requested):

projects/burnr/src/utils/constants.ts
wirednkod: Absolutely - I was thinking to keep that a bit more in case we decide to roll back but we can always dive back in git history and retrieve what we may need.

wirednkod: @raoulmillais I have gone through the refactor you mention, on removing extra things and leaving only one provider (and as you think there will only ever be one provider) then this also means we need some design changes (e.g. replace dropdown with a label, remove autocomplete/search of networks etc etc).

@goldsteinsveta can maybe share an opinion here.

I think though we should keep the functionality of multiple Networks (not local ones) as parachains can a potential pool that could be added here.

goldsteinsveta: I would not remove the selector just yet just because we will be adding parachains soon, as I understand. (If Westend's parachain capability is not planned to be maintained in burnr, then sure thing the selector can go)

wirednkod: @raoulmillais I kept the structure as is in order to verify if connection with parachains will be eventually need or not. If not then on later PR we can remove the specific structure and reduce Burnr wallet to usage of only 1 provider - which will reduce code and also redesign parts of the app.

raoulmillais: Thinking out loud... I guess but should we make burnr do transfers on parachain networks?  These demos are kind of a toy examples and that adds more complexity for us to support and for users to understand.  If we have a dedicated separate parachain demo then it probably covers the needs of demoing substrate/connect with parachains and allows us to show off something different than a balance transfer - something more parachain-specific.  I also feel like having a demo per use-case is a neater narrative for showcasing usage.

goldsteinsveta: I see you point. This is a healthy and sane approach for all tech demos at paritytech. Thinking out loud...if "switching between networks" is an interesting toy example, let's separate it into its own `projects/network-switch`, if it's not – let it go, I guess

wirednkod: I could say it makes sense what you say @raoulmillais .I will proceed with removing dropdown (and most likely keep it in a network-switch place)

raoulmillais (changes requested):
I thought we came to the conclusion we weren't going to add more networks and so it makes sense to simplify the code by not having a `Record<T,U>` of providers?

goldsteinsveta (approved):

raoulmillais (changes requested):
I can't build or run this PR for some reason:

```
λ ~/code/paritytech/substrate-connect/projects/landing-page/ git master  » g pr checkout 330
remote: Enumerating objects: 228, done.
remote: Counting objects: 100% (168/168), done.
remote: Compressing objects: 100% (49/49), done.
remote: Total 108 (delta 92), reused 72 (delta 58), pack-reused 0
Receiving objects: 100% (108/108), 15.90 KiB | 1.44 MiB/s, done.
Resolving deltas: 100% (92/92), completed with 36 local objects.
From github.com:paritytech/substrate-connect
   371d405..f6ed8f1  nik-remove-local-websocket-client -> origin/nik-remove-local-websocket-client
Switched to a new branch 'nik-remove-local-websocket-client'
λ ~/code/paritytech/substrate-connect/projects/landing-page/ git nik-remove-local-websocket-client  » cd ../burnr                                                                                [NORMAL]
λ ~/code/paritytech/substrate-connect/projects/burnr/ git nik-remove-local-websocket-client  » yarn clean
yarn run v1.22.10
$ rm -rf dist tsconfig.tsbuildinfo
Done in 0.07s.
λ ~/code/paritytech/substrate-connect/projects/burnr/ git nik-remove-local-websocket-client  » yarn dev
yarn run v1.22.10
$ yarn run clean && tsc -b & parcel ./public/index.html
$ rm -rf dist tsconfig.tsbuildinfo
ℹ️ Server running at http://localhost:1234
🚨 Build failed.
@parcel/transformer-json: JSON5: invalid end of input at 1:1
SyntaxError: JSON5: invalid end of input at 1:1
    at syntaxError (/home/raoul/code/paritytech/substrate-connect/node_modules/json5/lib/parse.js:1083:17)
    at invalidEOF (/home/raoul/code/paritytech/substrate-connect/node_modules/json5/lib/parse.js:1032:12)
    at Object.start (/home/raoul/code/paritytech/substrate-connect/node_modules/json5/lib/parse.js:821:19)
    at Object.parse (/home/raoul/code/paritytech/substrate-connect/node_modules/json5/lib/parse.js:32:32)
    at Object.transform (/home/raoul/code/paritytech/substrate-connect/node_modules/@parcel/transformer-json/lib/JSONTransformer.js:38:96)
    at async Transformation.runTransformer (/home/raoul/code/paritytech/substrate-connect/node_modules/@parcel/core/lib/Transformation.js:830:5)
    at async Transformation.runPipeline (/home/raoul/code/paritytech/substrate-connect/node_modules/@parcel/core/lib/Transformation.js:585:36)
    at async Transformation.runPipelines (/home/raoul/code/paritytech/substrate-connect/node_modules/@parcel/core/lib/Transformation.js:413:40)
    at async Transformation.run (/home/raoul/code/paritytech/substrate-connect/node_modules/@parcel/core/lib/Transformation.js:316:19)
    at async Child.handleRequest (/home/raoul/code/paritytech/substrate-connect/node_modules/@parcel/workers/lib/child.js:255:9)
```

raoulmillais:
> I can't build or run this PR for some reason

Same after a deep-clean reinstall and rebuild

raoulmillais (changes requested):
We don't need to have a `Record<T, U>` of providers ... there is only one.  I think this is the last simplifying step needed.  We should get rid of all the old config that isn't needed in the constants file (config for non-westend networks) and then [this config](https://github.com/paritytech/substrate-connect/blob/f6ed8f149a7a43d73cb81787817bf9c346902867/projects/burnr/src/utils/constants.ts#L30-L39) can just be something like:

```javascript
 export const PROVIDER = {
  id: 'Westend-WsProvider',  // Shouldn't be "WsProvider" 
  network: 'Westend',
  endpoint: endpoints.westend,  // Not needed either
  client: 'Light client',
 };
```

I don't think that provider config needs the `endpoint` either that looks like a hangover from when there were websocket connections and furthermore we don't need to worry about saving / retrieving the endpoint from local storage.  We know it will always be what's in the config constant.

Then you can fixup all the places that use the provider to use this simpler version instead of looking up the provider from the `Record<T, U>`

I'm pushing this because I'm trying to put myself in the shoes of someone who is looking at burnr code trying to understand how it's all put together (which hopefully is what our users will do).  All the leftover config and references to endpoints / multiple providers etc is confusing if you don't know the history of how this project got to where it is now.

raoulmillais (approved):

PR #334, opened 2021-06-16
fix yml tests

wirednkod:


wirednkod:
In order to avoid the "noise" that too many test commits will create when merged to master, this PR will close and a new one is opened ( #338 ) that contains all final changes and findings.

PR #335, opened 2021-06-16
Use project references to fix build

### LINKED ISSUE #292: Setup typescript project references
connect and extension depend on the source code in connect-extension-protocol and smoldot-test-utils.  We should configure typescript so that it knows this relationship.  This change means typescript will know to build the dependent projects when they have not been built or when there are changes in them.  It will also make code jumping and refactoring work across the projects.

See:

* https://www.typescriptlang.org/docs/handbook/project-references.html
* https://blog.logrocket.com/boost-your-productivity-with-typescript-project-references/

This is a good opportunity to standardise our tsconfigs and use a common base config across the packages.

The typescript eslint plugin [doesn't support project references yet](https://github.com/typescript-eslint/typescript-eslint/issues/2094) but according to [this comment](https://github.com/typescript-eslint/typescript-eslint/issues/2094#issuecomment-765142357) we can use [this option](https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/parser#parseroptionsproject) as a work around.   If so we can remove the `prelint` script that builds connect-extension-protocol before linting

raoulmillais:
* Create a root tsconfig to enable compiling everything from the root
* Extract a minimal tsconfig.base.json to share settings amongst projects
* Reference depedencies in project tsconfigs so tsc knows how to compile them if there are changes / no build outputs yet
* Fix typescript include paths for extension
* Correct usage of rootDir and baseUrl everywhere
* Remove superfluous tsconfig from SmoldotProvider directory
* Fix ConnectionManager imports in burnr
* Remove no-longer-needed postinstall script from smoldot-test-utils
* Use `tsc -b` everywhere to get incremental builds
* Upgrade typescript to latest and typescript eslint packages to latest stable

This fixes #292


build.out
```diff
@@ -0,0 +1,3 @@
+packages/connect/src/specs/kusama.json(235,590196): error TS1002: Unterminated string literal.
```
wirednkod: Should this file be pushed?

projects/burnr/package.json
```diff
@@ -14,8 +14,8 @@
     "prebuild": "yarn clean && mkdir dist && cp -r ./public/* ./dist/.",
     "build": "parcel build ./dist/index.html --public-url ./",
     "deep-clean": "yarn clean && rm -rf node_modules",
-    "clean": "rm -rf dist",
-    "dev": "yarn run clean && tsc --noEmit & parcel ./public/index.html",
+    "clean": "rm -rf dist tsconfig.tsbuildinfo",
+    "dev": "yarn run clean && tsc -b & parcel ./public/index.html",
```
wirednkod, line 18: Why remove the noEmit? tsc is typechecking while parcel takes care of compiling the code. Afaiu this change will typecheck, build and then parcel will compile the code to show it for dev.

projects/multiple-network-demo/package.json
```diff
@@ -12,9 +12,9 @@
   ],
   "scripts": {
     "deep-clean": "yarn clean && rm -rf node_modules",
-    "clean": "rm -rf dist",
-    "typecheck": "tsc --noEmit",
-    "typecheck:watch": "tsc --noEmit --watch",
+    "clean": "rm -rf dist tsconfig.tsbuildinfo",
+    "typecheck": "tsc -b",
```
wirednkod, line 16: by removing -noEmit then "typecheck" command also compiles the code. We are using parcel for that

projects/multiple-network-demo/tsconfig.json
```diff
@@ -1,18 +1,18 @@
 {
```
wirednkod, line 1: Im not sure which change does the following, but when building smoldot-browser-demo and multiple-network-demo, src directory is included in the dist directory which did not use to do before

projects/smoldot-browser-demo/tsconfig.json
```diff
@@ -1,18 +1,18 @@
 {
```
wirednkod, line 1: Im not sure which change does the following, but when building smoldot-browser-demo and multiple-network-demo, src directory is included in the dist directory which did not use to do before

projects/landing-page/package.json
```diff
@@ -13,8 +13,8 @@
     "prebuild": "yarn clean && mkdir dist && cp -r ./public/* ./dist/.",
     "build": "parcel build ./dist/index.html --public-url ./",
     "deep-clean": "yarn clean && rm -rf node_modules",
-    "clean": "rm -rf dist",
-    "dev": "yarn run clean && tsc --noEmit & parcel ./public/index.html",
+    "clean": "rm -rf dist tsconfig.tsbuildinfo",
+    "dev": "yarn run clean && tsc -b & parcel ./public/index.html",
```
wirednkod, line 17: by removing -noEmit then "tsc" command also compiles the code instead of just typechecking. We are using parcel for that

projects/smoldot-browser-demo/package.json
```diff
@@ -13,9 +13,9 @@
   ],
   "scripts": {
     "deep-clean": "yarn clean && rm -rf node_modules",
-    "clean": "rm -rf dist/ && mkdir -p dist/assets",
-    "typecheck": "tsc --noEmit",
-    "typecheck:watch": "tsc --noEmit --watch",
+    "clean": "rm -rf dist/ tsconfig.tsbuildinfo && mkdir -p dist/assets",
+    "typecheck": "tsc -b",
```
wirednkod, line 17: by removing -noEmit then "typecheck" command also compiles the code instead of just typechecking. We are using parcel for that

wirednkod (changes requested):
I added same comment more than one times in separate files just to point the issue where it repeats.

projects/burnr/package.json
raoulmillais: You can't use -b and --noEmit together. I decided it would be preferable to have incremental building and detection of changes in dependencies than suppressing the build output

projects/multiple-network-demo/package.json
raoulmillais: See last comment

projects/multiple-network-demo/tsconfig.json
raoulmillais: Oh fmdamn I must have missed a rootDir in those projects. They still work but it's not neat. Will fix.

projects/landing-page/package.json
raoulmillais: See above .. we can't detect changes I'd dependent projects (and therefore typecheck properly) with noEmit

projects/burnr/package.json
raoulmillais: Here is the justification from the typescript team on why noEmit wont build dependent references with `--noEmit` - https://github.com/microsoft/TypeScript/issues/40431

projects/smoldot-browser-demo/package.json
raoulmillais: As above

projects/smoldot-browser-demo/tsconfig.json
raoulmillais: Fixed

projects/multiple-network-demo/tsconfig.json
raoulmillais: Fixed

projects/burnr/package.json
wirednkod: I agree with the above. noEmit is used for not emitting compiler output files like JavaScript source code, source-maps or declarations. https://www.typescriptlang.org/tsconfig#noEmit . All I am saying is that so far we were using parcel to compile our project while keep tsc (with --noEmit) just for typechecking. By replacing --noEmit with -b we now use tsc to typecheck and compile, and then parcel to compile again.

raoulmillais: The point is a user cannot just clone then cd into burnr and run `yarn build` if we use `--noEmit` ... I think that is surprising.  By using `-b` instead of `--noEmit` its true that the compile is slightly slower than using `--noEmit` but its "bulletproof" .. it will always do what you expect no matter the state of the other projects (it will compile them if needed)

raoulmillais: Also if you make changes to `@substrate/connect` (and don't recompile it) then cd into one of the demos and then run `yarn dev` in the demo you will not have the changes in @substrate/connect since typescript with `--noEmit` won't check the dependency for changes - it will use the stale build output from `@substrate/connect`.  That's also surprising

raoulmillais: Regarding the slowdown by doing a full compile instead of using `--noEmit`.. that is probably only a concern for the first hit.  `-b` does incremental compilation so after the first build it will have build caches and so is able to optimise the build time depending on what has changed

wirednkod: I'm not sure I follow. I just replicated the steps you mentioned (after changing the `-b` to `--noEmit` ofc) and I was able to `yarn dev` from inside burnr dir.

wirednkod (approved):

PR #336, opened 2021-06-16
Share eslint config across projects

### LINKED ISSUE #293: Standardise and share our eslint configuration
Our eslint configuration is copy-pasted between the various packages with `"root": "true"` set in the `package.json`s.  We should introduce a shared eslint configuration to standaradise across the projects.

raoulmillais:
This PR fixes #293.  I have moved all the common eslint configuration into a root `.eslintrc.js`.  I favoured using a separate js file since it is easier to read when its not embedded in a `package.json` and we can use comments to explain why we configured / disabled rules if we want to.  Many of the projects no longer have any eslint config but those that do now have a minimal `.eslintrc.js` file specifying only the project-specific overrides.  I also removed the `no-misused-promises` rule as there were no errors and I moved the `restrict-plus-operands` rule config to the root.

### DIFF
--- a/packages/connect/src/__mocks__/client.js
+++ b/packages/connect/src/__mocks__/client.js
@@ -2,12 +2,6 @@
 /* eslint-disable @typescript-eslint/no-unsafe-call */
 /* eslint-disable @typescript-eslint/no-unsafe-member-access */
 /* eslint-disable @typescript-eslint/no-unsafe-assignment */
-import { JSDOM } from 'jsdom';
-const dom = new JSDOM();
-
-window.document = dom.window.document;
-window.window = dom.window;
-
 const localStorageMock = (() => {
     let store = {};
     return {
--- a/packages/connect/src/__mocks__/client.js
+++ b/packages/connect/src/__mocks__/client.js
@@ -5,8 +5,8 @@
 import { JSDOM } from 'jsdom';
 const dom = new JSDOM();
 
-global.document = dom.window.document;
-global.window = dom.window;
+window.document = dom.window.document;
+window.window = dom.window;
 
 const localStorageMock = (() => {
     let store = {};
@@ -26,4 +26,4 @@ const localStorageMock = (() => {
     };
 })();
 
-Object.defineProperty(global.window, 'localStorage', { value: localStorageMock });
+Object.defineProperty(window, 'localStorage', { value: localStorageMock });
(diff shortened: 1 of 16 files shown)


packages/connect/src/__mocks__/client.js
```diff
@@ -5,8 +5,8 @@
 import { JSDOM } from 'jsdom';
 const dom = new JSDOM();
 
-global.document = dom.window.document;
-global.window = dom.window;
+window.document = dom.window.document;
+window.window = dom.window;
```
wirednkod, line 9: This looks a little bit strange. Shouldn't this be `window = dom.window;`?
This is not a blocker (if this will create an  issue tests will show it)

wirednkod (approved):

packages/connect/src/__mocks__/client.js
raoulmillais: It's not just strange, it's wrong! I changed it hastily to fix linting errors, the linter is now configured for browser environment everywhere instead of a mixture of node and browser.  This is correct since jest is configured with the default jsdom (browser) env.  Interestingly I did run all the tests of connect after I made this change and everything passed.  I suspect that this is because the jest jsdom env automatically configures a window and document for you, so maybe this bit of the mock isn't required at all - I will experiment and see.

wirednkod: I say that this is strange cause I run the tests and they actually passed (not lint) - thats what surprised me a bit

raoulmillais: >  Shouldn't this be window = dom.window;

No window is the global object and replacing the global object is a really bad practice.  It might even screw with jest since jest is also modifying the global object - I'll have a play with various options and see what works.

raoulmillais: > It's not just strange, it's wrong! I changed it hastily to fix linting errors, the linter is now configured for browser environment everywhere instead of a mixture of node and browser. This is correct since jest is configured with the default jsdom (browser) env. Interestingly I did run all the tests of connect after I made this change and everything passed. I suspect that this is because the jest jsdom env automatically configures a window and document for you, so maybe this bit of the mock isn't required at all - I will experiment and see.

My suspicion was right, this JSDOM initialisation isn't needed as the jest jsdom environment already does this for us. Fixed by removing it.

wirednkod (approved):

Issue #185, opened 2021-03-24
Issue with smoldot Worker with CI tests (jest)


wirednkod:
There is an issue when Jest tests run that need smoldot client.
When smoldot client initialises (through smoldot Provider tested so far), during test coverage some logs coming form [smoldot-js] appear:

<img width="1176" alt="Screenshot 2021-03-24 at 9 53 43 AM" src="https://user-images.githubusercontent.com/5408605/112274216-e925b180-8c86-11eb-8459-2e0c31173de3.png">

At the end of the tests the following message appears:
```
A worker process has failed to exit gracefully and has been force exited. This is likely caused by tests leaking due to improper teardown. Try running with --detectOpenHandles to find leaks.
```
Even though the test pass fine the above issue do not allow CI to finish.

When jest runs with --detectOpenHandles to find leaks the following error appears:
<img width="947" alt="Screenshot 2021-03-24 at 9 55 25 AM" src="https://user-images.githubusercontent.com/5408605/112274456-2d18b680-8c87-11eb-9195-e9c20ff5c09b.png">

wirednkod:
Related comments:
https://github.com/paritytech/substrate-connect/pull/179#discussion_r600485429
https://github.com/paritytech/substrate-connect/pull/179#discussion_r600488990

tomaka:
I believe this was fixed by https://github.com/paritytech/smoldot/pull/1710 and https://github.com/paritytech/substrate-connect/pull/559

It turns our that on NodeJS, `Worker.terminate` is asynchronous (contrary to browsers where it's not) and we weren't waiting for the promise to be finished to end the test.

PR #338, opened 2021-06-16
Fix CI tests to run in github Actions

### LINKED ISSUE #325: CI should run all the tests
I can't remember why they were disabled but we should definitely have CI running all the tests as a priority

wirednkod:
After narrowing down one by one the tests the outcome is the following and actions have been taken in this PR for fixes:
_In SmoldotProvider.tests:_
a) `test('emits connect and never emits disconnect for development chain')`: as discussed this shall be skipped as is of dubious value;
b) `test('emits events when it connects / disconnects / reconnects')`: As (a) also this test was not allowing Jest to exit returning an error message: "Jest did not exit one second after the test run has completed. This usually means that there are asynchronous operations that weren't stopped in your tests.". This was fixed by moving specific test in a `describe` jest function in order to be able to add a "aterEach" function that will disconnect the provider once test is finished

_In Detector.tests.ts:_
a) tests with actual actual chains (polkadot, kusama, westend), during CI tests, ended up with timeouts as networks take longer to respond (more than 35 seconds each). Error returned was "Jest did not exit one second after the test run has completed. In other random times these would lead to the `terminate() has been called` error.

_In api.tests.ts:_
Tests were leading randomly to "Jest did not exit one second after the test run has completed"
> Note: Error FATAL ERROR: v8 appeared several times when having api.tests.ts. Never appeared again when they were removed Test run

As it was decided, `Detector.tests.ts` and `api.tests.ts` will be identified as integration tests. 2 new jest.conig files are introduced in this PR. First on (as of now default)  named `jest.config.unit.ts` that extends  the base `jest.config.ts` with excluding the integration tests and second named `jest.config.integration.ts` that runs the aforementioned tests and exclude the unit ones.

Fixes #325 (and apparently #185)


raoulmillais:
Excellent write-up! Thanks

packages/connect/src/SmoldotProvider/SmoldotProvider.test.ts
```diff
@@ -85,7 +112,8 @@ test('emits events when it connects then disconnects', async () => {
   });
 });
 
-test('emits events when it connects / disconnects / reconnects', async () => {
+/** Skip as a potential issue tests (see: https://github.com/paritytech/substrate-connect/issues/325#issuecomment-862585814) */
+test.skip('emits events when it connects / disconnects / reconnects', async () => {
```
raoulmillais, line 116: You skipped the wrong test.  'emits events when it connects / disconnects / reconnects' is definitely a valid useful test. We want to skip 'emits connect and never emits disconnect for development chain'  - although I actually think better to just remove it entirely since its not valuable. WE can get it back from git history if we think we want it later.  This

raoulmillais (changes requested):

packages/connect/src/SmoldotProvider/SmoldotProvider.test.ts
wirednkod: Yes my bad - the other test is fixed now. Will try to fix this and have no deleted or skipped tests

packages/connect/src/SmoldotProvider/SmoldotProvider.test.ts
```diff
@@ -19,9 +17,63 @@ import { jest } from '@jest/globals';
...
+          off(); // stop listening
+          provider.on('connected', () => {
+            provider.disconnect().then(() => resolve());
+          });
+        });
+      });
+    });
+  });
+
+  test('emits connect and never emits disconnect for development chain', async () => {
```
raoulmillais, line 52: I thought you were going to delete this test

raoulmillais (changes requested):

packages/connect/src/SmoldotProvider/SmoldotProvider.test.ts
wirednkod: As soon as there is a fix that makes it work - why to remove it? It was written to test some functionality - I was suggesting deletion if it was overkill to fix - but as soon as same solution applies for both tests, I don't find a reason to delete it

raoulmillais (approved):

Issue #220, opened 2021-04-12
Separate out deploying demos from deploying docs


raoulmillais:
We want to contionuously deploy the demos on every merge to master, but the docs should only be deployed when a version is published.  Currently everything is deployed continuosly.

* [ ] create separate deploy targets in root package.json for docs and demos
* [ ] only deploy demos in CI

wirednkod:
This should close. Demos, docs and extension are all deployed through CI now

PR #362, opened 2021-06-22
Autogenerate API docs on deployment

### LINKED ISSUE #320: Generate documentation site from tsdocs
Discuss integration with substrate.dev website.  Should code docs remain separate? (probably).  We have options:

* Use [typedoc](https://typedoc.org/) to generate static HTML
* Use [a plugin to docusaurus](https://github.com/tgreyuk/typedoc-plugin-markdown) which would allow us to also have freestanding markdown docs (e.g. for guides) to complement the code docs.
* Integrate with substrate.dev?

When we've decided our approach, we should try to style the code documentation to match the other documentation we have (create a separate ticket for this when we've decided which tools to use)

raoulmillais:
Note that this is the simplest possible implementation using the default theme and deploying the documentation everytime the deploy script runs.  Next PR will refactor the deploy script to allow publishing docs separately from deployment (issue #220)

This fixes #320 and pins parcel in the root to the known good beta


raoulmillais:
I need to add a link to the docs from the landing page

wirednkod (approved):
lgtm

PR #365, opened 2021-06-22
Tidy up build and test output

raoulmillais:
This contains various minor fixes to remove all warnings and errors from the build and test output as well as to reduce all superflous output from those commands. The unit test output now fits on a single screen.

* Fixed a leftover relative pacakge import of the test-utils to use the workspace name instead
* Stop using `process.nextTick` in test-utils - all tests now run in a browser env so dont use node APIs (removes a test warning)
* removes unused `test:watch` scripts from pacakages that dont have tests (we actually dont have `test:watch` scripts anywhere)
* Fix connect jest config to remove warning about duplicate mocks (tells jest that files are all in dist/)
* Fix incorrect connect jest config about ts and tsx extensions being ESM (we point at the compiled output)
* Make top-level scripts that run scripts in all workspaces less verbose
* Standardise how we run node with esm flags in a top-level helper script so we can change flags in one place
* Suppress the warnings about modules being experimental
* Only show failing tests when running tests
* Slimmer coverage output summary


packages/connect/package.json
```diff
@@ -42,14 +42,15 @@
...
-    "@substrate/smoldot-test-utils": "file:../smoldot-test-utils",
+    "@substrate/smoldot-test-utils": "^0.1.0",
     "@types/jest": "^26.0.20",
     "@types/jsdom": "^16.2.7",
     "@types/mkdirp": "^1.0.1",
     "@typescript-eslint/parser": "^4.27.0",
     "eslint": "^7.21.0",
     "eslint-plugin-tsdoc": "^0.2.14",
     "jest": "^27.0.0-next.9",
+    "jest-silent-reporter": "^0.5.0",
```
wirednkod, line 53: why is this added? I dont thing we need this for any reason

packages/connect/package.json
```diff
@@ -23,9 +23,9 @@
     "prebuild": "yarn clean && yarn downloadSpecs",
     "pretest": "yarn build",
     "test": "yarn test:unit",
-    "test:all": "node --experimental-vm-modules $(yarn bin)/jest --colors --silent --coverage dist/",
-    "test:unit": "node --experimental-vm-modules $(yarn bin)/jest --colors --silent --coverage dist/ --config=jest.config.unit.ts",
-    "test:integration": "node --experimental-vm-modules $(yarn bin)/jest --colors --silent --coverage dist/ --config=jest.config.integration.ts",
+    "test:all": "../../bin/node-esm $(yarn bin)/jest --colors --silent --coverage dist/",
```
wirednkod, line 26: I would prefer to avoid accessing bin directly and use the `node` command as it was before. Also I would avoid `--no-warnings`. It is better to see warnings when they occur (also its not like we will be altering the flags of this too often)

packages/connect/package.json
raoulmillais: Its the reporter used to minimse the test outputs so that it passes silently but fails on failing tests.

packages/connect/package.json
raoulmillais: we had experimental warnings for using modules >10 times in the testoutput before this.  This only hides the node warnings not any warnings from packages

packages/smoldot-test-utils/package.json
```diff
@@ -12,13 +12,15 @@
     "deep-clean": "rm -rf node_modules",
     "clean": "rm -rf dist/ tsconfig.tsbuildinfo",
     "build": "tsc -b --clean",
-    "test": "echo \"No tests for smoldot-test-utils\"",
+    "test": "exit 0; #This is a test support library",
     "lint": "eslint . --ext .js,.ts"
   },
   "dependencies": {
+    "asap": "^2.0.6",
```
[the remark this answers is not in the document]
raoulmillais: This is a very well-known battle tested library: It's written by a nodejs core contributor.  There's been no changes to it because its "done".  We can't use `setImmediate` because its non-standard and we shouldn't use `process.nextTick` because its a node specific API and we (rightly) get warnings about it.  It's the recommended (in MDN) way for tests to simulate yielding to the event loop to simulate asynchronous behaviour - this is exactly what we are doing here.

packages/connect/package.json
wirednkod: I did not initially realized that this `bin` is ours.. I got confused that it dives into node_modules or something..

wirednkod (approved):

packages/connect/package.json
raoulmillais: Yup it's a tiny wrapper script so we can change the flags we use in one place and avoid having long unreadable scripts in the package json

packages/connect/jest.config.ts
```diff
@@ -1,10 +1,12 @@
...
 const config: Config.InitialOptions = {
+  coverageReporters: ["text-summary"],
+  reporters: ["jest-silent-reporter"],
+  roots: ['<rootDir>/dist'],
   verbose: true,
   transformIgnorePatterns:["./src/__mocks__/client.js"],
   testURL: "http://localhost/",
-  transform: {},
-  extensionsToTreatAsEsm: [".ts", ".tsx"]
+  transform: {}
```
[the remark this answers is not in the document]
raoulmillais: Fixed

PR #367, opened 2021-06-23
Bump jest version to stable 27.0.5 instead of @next

wirednkod:
As soon as 27 is officially released I thought its a good idea to bump the version to a stable (^27.0.5) one instead of (^27.0.0-next.9). Run deep clean and test and all passed.


raoulmillais:
🎉

raoulmillais (approved):

PR #397, opened 2021-07-06
Fix bug that was not passing 2 values on the localStorage.setItem

wirednkod:


projects/burnr/src/utils/utils.ts
```diff
@@ -128,8 +128,8 @@ export const validateLocalstorage = (): void => {
     } else {
       // Check if the values of existing keys are among the accepted ones
       // if not then set the default value of expectedValies (index 0)
-      if (expectedValues[key] && !(expectedValues[key] as string[]).includes(localStorage[key])) {
-        localStorage.setItem(key);
+      if (expectedValues[key] && !(expectedValues[key] as string).includes(localStorage[key])) {
```
raoulmillais, line 131: 1. `expectedValues[key]` can never be falsy: you're inside an `Object.keys` loop over a literal which is known not to have properties with nullish values.
2. `expectedValues[key]` can either be a string or an array of strings - it's weird casting it to a string here but I guess you did that for the compiler
3. (below) say one of the items expecting "true" or "false" is set to something else then you're now going to set it to the whole array of ["true", "false"] (cast to a string)

This whole function took me quite a few attempts to parse and understand.  I think you've maybe overthought things trying to make the function extensible for future defaults and having it both overwrite unknown values and populate with defaults.  

You provide the defaults elsewhere because burnr has to work when the local storage is empty, so you could drastically simplify this by simply removing any values that were unexpected and doing nothing for missing values. Additionally if you standardise all the expected values to arrays of strings it makes the checks you have to do much easier.  Something like:

```javascript
 // Add type info to avoid having to cast
  const expectedValues: Record<string, string[]> = {
    "theme": ["true", "false"],
    "balanceVisibility": ["true", "false"],
    "endpoint": [ ALL_PROVIDERS.network ] // now an array although we don't even really need this in storage any more
  };

  Object.keys(expectedValues).forEach(key => {
    if (!Object.keys(localStorage).includes(key)) { 
      return;
    }
    
    if (!expectedValues[key].includes(localStorage[key])) {
      localStorage.removeItem(key);
    }
  });
```

raoulmillais (changes requested):

raoulmillais (approved):

PR #404, opened 2021-07-08
Split detector to Connect and Provider functions

wirednkod:
This PR splits the .connect() function to 2. One provider() function that detects and returns the detected Provider (SmoldotProvider or ExtensionProvider) and a simplified connect() that is using the latter for the functionality that was already implementing.

Reason of this split is having the provider() function for advanced usage (e.g. PolkadotJS) and for internal use.

### DIFF
--- a/packages/connect/src/Detector.ts
+++ b/packages/connect/src/Detector.ts
@@ -22,8 +22,8 @@ import polkadot from './specs/polkadot.json';
  *
  * // Create a new UApp with a unique name
  * const app = new Detector('burnr-wallet');
- * const westend = await app.detect('westend');
- * const kusama = await app.detect('kusama');
+ * const westend = await app.connect('westend');
+ * const kusama = await app.connect('kusama');
  *
  * await westend.rpc.chain.subscribeNewHeads((lastHeader) => {
  *   console.log(lastHeader.hash);
@@ -107,15 +107,17 @@ export class Detector {
     return await ApiPromise.create(Object.assign(options ?? {}, {provider}));
   }
 
-  /**
-   * creates and returns a provider (either Smoldot or Extension one) to be used
-   * in PolkadotJS API instance
+  /** 
+   * Detects and returns an appropriate PolkadotJS provider depending on whether the user has the substrate connect extension installed
    * 
    * @param chainName - the name of the blockchain network to connect to
    * @param chainSpec - an optional chainSpec to connect to a different network
    * @returns a provider will be used in a ApiPromise create for PolkadotJS API
    *
    * @internal
+   * 
+   * @remarks 
+   * This is used internally for advanced PolkadotJS use cases and is not supported.  Use {@link connect} instead.
    */
   public provider = (chainName: string, chainSpec?: string): ExtensionProvider | SmoldotProvider => {
     let provider: ExtensionProvider | SmoldotProvider = {} as ExtensionProvider | SmoldotProvider;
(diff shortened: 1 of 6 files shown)


packages/connect/src/Detector.ts
```diff
@@ -100,6 +100,24 @@ export class Detector {
...
   public connect = async (chainName: string, chainSpec?: string, options?: ApiOptions): Promise<ApiPromise> => {
+    const provider: ExtensionProvider | SmoldotProvider = this.provider(chainName, chainSpec);
+    provider.connect().catch(console.error);
+
+    this.#providers[chainName] = provider as ProviderInterface;
+    return await ApiPromise.create(Object.assign(options ?? {}, {provider}));
+  }
+
+  /**
+   * creates and returns a provider (either Smoldot or Extension one) to be used
```
raoulmillais, line 111: "Detects and returns an appropriate PolkadotJS provider depending on whether the user has the substrate connect extension installed"

packages/connect/src/Detector.ts
```diff
@@ -100,6 +100,24 @@ export class Detector {
...
+   * in PolkadotJS API instance
+   * 
+   * @param chainName - the name of the blockchain network to connect to
+   * @param chainSpec - an optional chainSpec to connect to a different network
+   * @returns a provider will be used in a ApiPromise create for PolkadotJS API
+   *
```
raoulmillais, line 117: Add a remark about not being supported too. Something like

```
@remarks

This is used internally for advanced PolkadotJS use cases and is not supported.  Use {@link connect} instead.
```

raoulmillais (changes requested):
Some minor nits on the docs in inline comments and you also need to update the landing page example.

This should PR should include updates to the demos to be 100% it all works. The @substrate/connect API and it's usage in the demos should always change together in the same PR.

You could also bump @substrate/connect version ready to be published when all the above is done and tested :smile:

packages/connect/src/Detector.ts
wirednkod: > Some minor nits on the docs in inline comments and you also need to update the landing page example.
> 
> This should PR should include updates to the demos to be 100% it all works. The @substrate/connect API and it's usage in the demos should always change together in the same PR.

Can you please elaborate what you meam by "updates to the demos"? 
Demos are using afaik `.connect()` and this change should not affect or alter their code

raoulmillais: ignore me I'm going mad! for some reason I'd got it my head we'd changed this

raoulmillais (changes requested):
What about the documentation fixes I proposed for `provider()`?  Also I've apparently been confusing myself about `Detector.connect()` vs `Detector.detect()` for some time!  I got that wrong in the docs for [`Detector`](https://github.com/paritytech/substrate-connect/blob/242ffcaad4d6854c74cd556d59363f725641529b/packages/connect/src/Detector.ts#L25-L26) and [the landing page](https://github.com/paritytech/substrate-connect/blob/242ffcaad4d6854c74cd556d59363f725641529b/projects/landing-page/src/App.tsx#L104-L105).  Can we fix that at the same time since its tiny?

wirednkod:
> What about the documentation fixes I proposed for `provider()`? Also I've apparently been confusing myself about `Detector.connect()` vs `Detector.detect()` for some time! I got that wrong in the docs for [`Detector`](https://github.com/paritytech/substrate-connect/blob/242ffcaad4d6854c74cd556d59363f725641529b/packages/connect/src/Detector.ts#L25-L26) and [the landing page](https://github.com/paritytech/substrate-connect/blob/242ffcaad4d6854c74cd556d59363f725641529b/projects/landing-page/src/App.tsx#L104-L105). Can we fix that at the same time since its tiny?

I added it, but apparently if one does not save the file - then change will not be pushed 😖 

Will push it now and add the change for landing page as well!! 👍🏽

raoulmillais (approved):

PR #406, opened 2021-07-09
Fix issue with extensions .js that was breaking ts-jest

wirednkod:
Added an extension-resolver for jest that does exactly what we are searching for: finding .js extensions and replacing them temporary with .ts for ts-jest to run.

### DIFF
--- a/packages/connect/extension-resolver.cjs
+++ b/packages/connect/extension-resolver.cjs
@@ -1,0 +1,28 @@
+let defaultResolver;
+
+function requireDefaultResolver() {
+  if (!defaultResolver) {
+    try {
+      defaultResolver = import (`jest-resolve/build/defaultResolver`);
+    } catch (error) {
+      defaultResolver = import (`jest-resolve/build/default_resolver`);
+    }
+  }
+
+  return defaultResolver;
+}
+
+
+module.exports = (request, options) => {
+  const {basedir, defaultResolver, extensions} = options;
+
+  if (!defaultResolver) {
+    defaultResolver = requireDefaultResolver();
+  }
+
+  try {
+    return defaultResolver(request, options);
+  } catch (e) {
+    return defaultResolver(request.replace(/\.js$/, '.ts'), options);
+  }
+}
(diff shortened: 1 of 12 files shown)


packages/connect/extension-resolver.cjs
```diff
@@ -0,0 +1,29 @@
+let defaultResolver;
+
+function requireDefaultResolver() {
+  if (!defaultResolver) {
+    try {
+      defaultResolver = import (`jest-resolve/build/defaultResolver`);
+    } catch (error) {
+      defaultResolver = import (`jest-resolve/build/default_resolver`);
```
raoulmillais, line 8: This is here to support older versions of jest .. we only need the first import and no try / catch

packages/connect/extension-resolver.cjs
```diff
@@ -0,0 +1,29 @@
...
+  const {basedir, defaultResolver, extensions} = options;
+
+  if (!defaultResolver) {
+    defaultResolver = requireDefaultResolver();
+  }
+
+  try {
+    return defaultResolver(request, options);
+  } catch (e) {
+    console.log(`${e} - Temporary replace extension '.js' to '.ts' for tests to run.\n`)
```
raoulmillais, line 26: Check `e.code` and throw the error if it is not `MODULE_NOT_FOUND` also check that request ends with `'js'` ... we should only be trying to fix the resolve in those cases otherwise we should throw early

packages/connect/extension-resolver.cjs
```diff
@@ -0,0 +1,29 @@
...
+    return defaultResolver(request.replace(/\.js$/, '.ts'), options);
```
raoulmillais, line 27: Wrap this second call to `defaultResolver` in its own `try` / `catch` and log in the catch block saying something like "errored resolving a module after replacing js extension with ts" and then log the outer `e` (not the error in this catch block) before rethrowing the inner error (the error in this catch block).  That way both messages and stack traces will show up giving us as much context as possible to track down the problem

raoulmillais (changes requested):

raoulmillais (approved):

raoulmillais (changes requested):
We need to add an attribution for this code since we've adapated it from someone else's work (and they used the MIT license, plus it's the nice thing to do.)  Please add a comment at the top including a link to the source repo, the author's name and the fact it was MIT licensed

raoulmillais (approved):

PR #419, opened 2021-07-19
Upgrade to parcel@2.0.0-beta.3.1

raoulmillais:
New tagged beta.  Tested all the projects and they build fine, this also makes parcel understand typescript project references so you can build the demos before you build @substrate/connect and it knows to go and build that first.


wirednkod (changes requested):
SmoldotProvider does not seem to initiate for some reason (With Extension it works fine)

PR #420, opened 2021-07-21
Update Kusama chain spec with a recent checkpoint

tomaka:
Adds a more recent checkpoint around block 8.4 million.

cc https://github.com/paritytech/smoldot/pull/1029


wirednkod (approved):

wirednkod:
Once this is merged - we need to release new version for the chainspecs to pass to substrate-connect and extension

PR #423, opened 2021-07-22
Extension - merge chains with networks

wirednkod:
Merge the entities of networks and chains into one

### DIFF
--- a/projects/extension/src/background/ConnectionManager.test.ts
+++ b/projects/extension/src/background/ConnectionManager.test.ts
@@ -176,23 +176,28 @@ describe('Unit tests', () => {
     expect(manager.apps).toHaveLength(4);
   });
 
-  test('Get networks', () => {
-    expect(manager.networks).toEqual([
-      { name: 'westend', status: "connected", chainspecPath: "westend.json", isKnown: true },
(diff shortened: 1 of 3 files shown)


projects/extension/src/background/ConnectionManager.test.ts
```diff
@@ -176,23 +176,26 @@ describe('Unit tests', () => {
...
-  test('Get networks', () => {
-    expect(manager.networks).toEqual([
-      { name: 'westend', status: "connected", chainspecPath: "westend.json", isKnown: true },
-      { name: 'kusama', status: "connected", chainspecPath: "kusama.json", isKnown: true }
-    ]);
-  });
-
-  test('Get chains', () => {
+  test('Get networks/chains', () => {
     const tmpChains: unknown[] = [];
```
raoulmillais, line 180: Use `map` instead of `forEach` with separate variable and `push` .. then you don't have to declare the var awkwardly as `unknown[]` and 3 lines of code becomes 1 :smile:

raoulmillais (changes requested):

raoulmillais (approved):

PR #428, opened 2021-07-27
Extension - Add/Remove Chains (Smoldot v.0.3.x)

### LINKED ISSUE #427: Extension for smoldot v.0.3.X - Add/Remove Chains
When an app is connected in Smoldot through the extension, it should call Smoldot with the chain specs, no matter if the chain is a relay chain or not. Smoldot will understand and keep only 1 chain connected.
During connection the Extension should keep an Id on the app for each Chain that is connected to.
Upon disconnection the Extension should call again Smoldot with `remove()` function, for the specific chain that app was connected to.

wirednkod:
Fixes #427

### DIFF
--- a/packages/connect-extension-protocol/src/index.ts
+++ b/packages/connect-extension-protocol/src/index.ts
@@ -68,7 +68,7 @@ export interface ProviderMessageData {
   /** The name of the blockchain network the app is talking to **/
   chainName: string;
   /** What action the `ExtensionMessageRouter` should take **/
-  action: 'forward' | 'connect' | 'disconnect';
+  action: 'forward' | 'connect' | 'disconnect' | 'spec';
   /** The message the `ExtensionMessageRouter` should forward to the background **/
   message?: MessageToManager;
 }
@@ -79,7 +79,7 @@ export interface ProviderMessageData {
  */
 export interface MessageToManager {
   /** Type of the message. Defines how to interpret the {@link payload} */
-  type: 'rpc';
+  type: 'rpc' | 'spec';
   /** Payload of the message -  a JSON encoded RPC request **/
   payload: string;
   /** whether an RPC message is a subscription or not **/
--- a/packages/connect-extension-protocol/src/index.ts
+++ b/packages/connect-extension-protocol/src/index.ts
@@ -80,9 +80,9 @@ export interface ProviderMessageData {
  */
 export interface MessageToManager {
   /** Type of the message. Defines how to interpret the {@link payload} */
-  type: 'associate' | 'rpc';
-  /** Payload of the message. Either a JSON encoded RPC response or the name of a network **/
-  payload: string; // smoldot name or JSON encoded RPC message
+  type: 'rpc';
+  /** Payload of the message -  a JSON encoded RPC request **/
+  payload: string;
   /** whether an RPC message is a subscription or not **/
   subscription?: boolean;
 }
--- a/packages/connect-extension-protocol/src/index.ts
+++ b/packages/connect-extension-protocol/src/index.ts
@@ -24,7 +24,7 @@ export interface ExtensionMessageData {
   /** message is telling the `ExtensionProvider` the port has been closed **/
   disconnect?: boolean;
   /** message is the message from the manager to be forwarded to the app **/
-  message?: MessageFromManager
+  message?: MessageFromManager;
 }
 
 /**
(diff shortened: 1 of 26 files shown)


raoulmillais (changes requested):
In general I think this is a really good start :+1: The biggest issue I can see so far is that I really don't think we should be sending the `spec` message using the `method` field of the RPC message. We don't own the JSON RPC protocol that is the domain of smoldot.  We have our own `type` field in the `MessageToManager` which is designed to ask the background - you should extend that instead and use `provider.send()` to send it from the `ExtensionProvider`.

Aside: Our logging strategy is going a bit all over the place in the extension, but I don't think you need to fix that in this PR: we need aticket for that though.

wirednkod:
> Aside: Our logging strategy is going a bit all over the place in the extension, but I don't think you need to fix that in this PR: but we need a ticket for it though.

Add new issue #429

packages/connect-extension-protocol/src/index.ts
```diff
@@ -68,7 +68,7 @@ export interface ProviderMessageData {
   /** The name of the blockchain network the app is talking to **/
   chainName: string;
   /** What action the `ExtensionMessageRouter` should take **/
-  action: 'forward' | 'connect' | 'disconnect';
+  action: 'forward' | 'connect' | 'disconnect' | 'spec';
```
raoulmillais, line 71: I don't think we want a new `action` at this layer.  We want a message from the `ExtensionProvider` to ask the `ExtensionMessage` to **`forward`** a `'spec'` message.

packages/connect/src/ExtensionProvider/ExtensionProvider.ts
```diff
@@ -216,8 +216,19 @@ export class ExtensionProvider implements ProviderInterface {
 
     // Once connect is sent - send rpc to extension that will contain the chainSpecs
     // for the extension to call addChain on smoldot
-    this.send('spec', [this.#chainSpecs]).catch(console.error);
-    
+    const someMsg: ProviderMessageData = {
+      appName: this.#appName,
+      chainName: this.#chainName,
+      action: 'spec',
```
raoulmillais, line 222: See above

projects/extension/src/content/ExtensionMessageRouter.ts
```diff
@@ -112,6 +124,12 @@ export class ExtensionMessageRouter {
       return this.#establishNewConnection(message);
     }
 
+    if (data.action === 'spec') {
+      console.log('data action: ', data.action);
+      console.log('message', message.data)
```
raoulmillais, line 129: Change these to use the `debug` helper (or remove them)

raoulmillais (changes requested):

projects/extension/src/content/ExtensionMessageRouter.ts
wirednkod: Sorry about these stray logs - will be removed

packages/connect-extension-protocol/src/index.ts
wirednkod: The idea behind this was to have complete different flows for spec and other json-rpc forwards for future usage - but if it makes sense to have flow until split this on the message level - I will proceed with that

projects/extension/src/background/AppMediator.ts
```diff
@@ -261,14 +261,14 @@ export class AppMediator extends (EventEmitter as { new(): StateEmitter }) {
     }
     const chainName = this.#chainName as string;
 
-    if (parsed.method === 'spec' && chainName) {
+    if (msg.type === 'spec' && chainName) {
       // When params[0] (chainSpecs in the current case is empty) then this is a
```
raoulmillais, line 265: This comment is no longer valid

raoulmillais (approved):
Just one little stray comment to remove, and this is good to merge :+1:

projects/extension/src/background/AppMediator.ts
```diff
@@ -14,6 +13,18 @@ import {
   StateEmitter,
   SubscriptionMapping
 } from './types';
+import { SmoldotChain } from 'smoldot';
+import westend from '../../public/assets/westend.json';
+import kusama from '../../public/assets/kusama.json';
+import polkadot from '../../public/assets/polkadot.json';
+
+type RelayType = Record<string, string>
```
raoulmillais, line 21: Not a blocker .. but this should be a `Map<string, string>`

wirednkod: This does not work:
```
Type '{ polkadot: string; kusama: string; westend: string; }' is not assignable to type 'RelayType'.
  Object literal may only specify known properties, and '"polkadot"' does not exist in type 'Map<string, string>'.ts(2322)
```

Issue #313, opened 2021-06-10
ExtensionProvider 'connected' should fire under same conditions as SmoldotProvider


raoulmillais:
If you connect to the extension with @substrate/connect and the smoldot in the extension is still syncing then the ExtensionProvider should not emit 'connected' until such time as the smoldot in the extension is finished syncing.  Furthermore it should be firing `disconnected` events when `isSyncing` goes false -> true or `peerCount` drops to zero

PR #434, opened 2021-08-03
Extension Refactor - remove request and subscription tracking

### LINKED ISSUE #438: Remove unneeded request and subscription tracking
The extension no longer needs to do this with smoldot v0.3.x as:
* the `SmoldotChain` will track all the requests and clean up when `remove()` is called
* there is a `SmoldotChain` per `AppMediator` so no need to remap the ids

raoulmillais:
This fixes #438.  It also fixes 2 bugs in the implementation.

* The `AppMediator` was asking the `ConnectionManager` to send messages on its behalf but the `ConnectionManager` would just pick the first chain with the matching name instead of using the chain that was created for that `AppMediator`.  I removed the message sending from the `ConnectionManager` and the `AppMediator` just asks the chain it has to send the message.
* If an RPC message was received before a spec message or before `addChain` had completed then it would crash because the chain was not yet defined.  I now buffer any early RPC messages that come in and when the `addChain` completes check for those messages and send them all.

I spotted a couple more small issues with the implementation but they aren't causing complete crashes / erroneous behaviour so I added TODO comments to clean them up.

### DIFF
--- a/projects/extension/src/background/AppMediator.ts
+++ b/projects/extension/src/background/AppMediator.ts
@@ -38,6 +38,7 @@ export class AppMediator extends (EventEmitter as { new(): StateEmitter }) {
   #chainName: string | undefined  = undefined;
   #chain: SmoldotChain | undefined;
   #state: AppState = 'connected';
+  #pendingRequests: string[] = [];
 
   /**
    * @param port - the open communication port between the app's content page
@@ -78,6 +79,7 @@ export class AppMediator extends (EventEmitter as { new(): StateEmitter }) {
     this.#chainName = this.#port.name.substr(splitIdx + 2, this.#port.name.length);
     return true;
   }
+
   /** 
    * name is the name of the communication port 
    * 
@@ -143,6 +145,12 @@ export class AppMediator extends (EventEmitter as { new(): StateEmitter }) {
       .then(chain => {
         if (chain) {
           this.#chain = chain;
+          // process any RPC requests that came in while waiting for `addChain`
+          // to complete
+          if (this.#pendingRequests.length > 0) {
+            this.#pendingRequests.forEach(req => chain.sendJsonRpc(req));
+            this.#pendingRequests = [];
+          }
         } else {
           // TODO: remove this duplicate error block when `addChain` on the 
           // ConnectionManager has been cleaned up to not catch errors adding
@@ -172,9 +180,10 @@ export class AppMediator extends (EventEmitter as { new(): StateEmitter }) {
     }
 
     if (this.#chain === undefined) {
-      this.#sendError('Cannot send RPC message to chain before spec message');
-      this.#manager.unregisterApp(this);
-      return this.#port.disconnect();
+      // `addChain` hasn't resolved yet after the spec message so buffer the
+      // messages to be sent when it does resolve
+      this.#pendingRequests.push(msg.payload);
+      return;
     }
 
     return this.#chain.sendJsonRpc(msg.payload);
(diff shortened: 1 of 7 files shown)


projects/extension/src/background/AppMediator.ts
```diff
@@ -61,19 +48,38 @@ export class AppMediator extends (EventEmitter as { new(): StateEmitter }) {
...
     this.#manager = manager;
     // Open listeners for the incoming rpc messages
-    this.#port.onMessage.addListener(this.#handleRpcRequest);
+    this.#port.onMessage.addListener(this.#handleMessage);
     this.#port.onDisconnect.addListener(() => { this.#handleDisconnect() });
   }
 
+  /** 
+   * associate parses the name of the network from the port name and associates
+   * the app with the smoldot client or sends an error and disconnects the port
```
wirednkod, line 64: This is slightly misleading. Association should happen with the chain name and not with the smoldot client (there is only 1).

projects/extension/src/background/AppMediator.ts
```diff
@@ -117,187 +123,70 @@ export class AppMediator extends (EventEmitter as { new(): StateEmitter }) {
...
-    // change the message ID to the ID the app is expecting
-    message.id = request.appID
-    this.#port.postMessage({ type: 'rpc', payload: JSON.stringify(message) });
-
-    return true;
-  }
-
-  #addChain = async (chainName: string, chainSpecs: string): Promise<void> => {
-    this.#chain = await this.#manager.addChain(chainName, chainSpecs);
+    this.#manager.addChain(chainName, chainSpec, rpcCallback)
```
wirednkod, line 144: Im not sure why this is communicated first to the manager and then added to the local var.

In addition as discussed (but did not ended up with a decision) - there should not be pending messages arriving until the chain is connected. Am I missing something? If for any reason this does not conclude user will not be aware that the pending messages were not delivered.

The flow below does not guarantee the deliverance of the messages (for eample `if (!chain)` or in case of error the pendingRequests must be answered.

wirednkod: In addition based on this #443 - app should wait for a "connected" state in order to start sending RPC messages. Substrate-connect imo should not be a pending-message-holder

raoulmillais: If we track the app's "connected" state after #443 then suddenly we're tracking and communicating 2 different types of "connected"ness and complicating the protocol to do so.  This pending request tracking is a backstop for a super-edge case so as not to lose requests.  In reality the only supported use case for the extension is by our own `ExtensionProvider` and we know that the `ExtensionProvider` is always going to send the `'spec'` message immediately.  Once #313 (healthchecking) is implemented, we know that polkdotjs API will not make any RPC requests on behalf of the app until we know the chain in the background has stopped syncing.  However its quite possible that the healthchecker will send an 'RPC' message before the `addChain` completed in the background, this `pendingRequest` mechanism will simply hold onto those requests and respond to them (with a slight delay).  

> The flow below does not guarantee the deliverance of the messages (for eample if (!chain) or in case of error the pendingRequests must be answered.

I think that #442 and #439 cover this concern.  When those are implemented we will 1) send proper errors back for problems adding chains and 2) reject all the promises for the cases where the app has made some requests and, before they complete, the app receives a 'disconnect' or 'error' message back.

wirednkod: I agree with the approach above based on your comment @raoulmillais . I think we cover 99.5% of the scenarios and there are extremely edge cases that we may have not thought off - or met yet.

wirednkod (approved):

PR #445, opened 2021-08-06
Improve protocol documentation

raoulmillais:
Added some introductory notes on the protocol as documentation to make understanding the types easier.

### DIFF
--- a/packages/connect-extension-protocol/src/index.ts
+++ b/packages/connect-extension-protocol/src/index.ts
@@ -1,4 +1,29 @@
 /**
+ * @packageDocumentation
+ *
+ * In order to understand the protocol you should realise there are actually
+ * 2 hops that happen in communication because of the architecture of browser
+ * extensions.  The app has to `window.postMessage` messages to the content
+ * script that gets injected by the extension. It is the content script that
+ * has access to the extension APIs to be able to post messages to the 
+ * extension background.
+ *
+ * You can think of the protocol types like layers of an onion. The innermost 
+ * layer is the original JSON RPC request/responses. Then we wrap extra layers 
+ * (types) for the other 2 hops which then get peeled off at each hop. The 
+ * {@link MessageToManager} / {@link MessageFromManager} representing the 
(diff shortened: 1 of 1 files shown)


packages/connect-extension-protocol/src/index.ts
```diff
@@ -1,3 +1,28 @@
...
+ * extension background.
+ *
+ * You can think of the protocol types like layers of an onion. The innermost 
+ * layer is the original JSON RPC request/responses. Then we wrap extra layers 
+ * (types) for the other 2 hops which then get peeled off at each hop. The 
+ * {@link MessageToManager} / {@link MessageFromManager} representing the 
+ * extension communication content script \<\> background. Then the outermost 
+ * {@link ExtensionMessage} / {@link ProviderMessage} representing the 
+ * communication between the PolkadotJS provider in the app and the content 
+ * script.
```
tomaka, line 18: But why do that, instead of just one message? The communication between the PolkadotJS provider and the content script has to be stable, because the extension and the app can update separately. We shouldn't make this communication layer depend on the implementation details of the extension.

raoulmillais: Because the app needs to be able to ask the extension (content script) to connect / disconnect on its behalf

raoulmillais: So the outer layer is either 'connect' / 'forward' / 'disconnect' instructing what the content script should do.  Anything that is 'forward' is in the inner layer which is the bit where we are sending / receiving RPC messages and establishing the chainspec(s)

tomaka: > Because the app needs to be able to ask the extension's content script to connect / disconnect on its behalf to the background of the extension

Why does the app need to be ask the content script to do things? Can the content script not decide by itself how to handle the messages? The app isn't supposed to know how the content script works.

raoulmillais (commented):
This is an interesting discussion but I don't think it should block the docs improvement

wirednkod (approved):

Issue #225, opened 2021-04-13
Unit tests for extension UI


raoulmillais:
**Popup Page:**
- [ ] No tabs active
- [ ] Active tab
- [ ] Non-active tab
- [ ] Settings Button
- [ ] Networks Button
- [ ] About Button

**Option Page**
- Network tab components (Add)
- Settings tab components (Add)

**Notifications**
- [ ] Test notification functionality

PR #447, opened 2021-08-09
Add notifications in extension

### LINKED ISSUE #437: Add Extension notifications
Create Extension Notifications: A pop up that will inform  concerning Apps connected in Substrate-Connect.
A switch should exist in settings in order to have notifications active or nor

wirednkod:
Added a very early version of notification system in the Extension.
For now (if active from Networks screen - will be transfered to separate settings screen) they are popping in the notifications informing that an app got connected to a chain, 
Fixes #437


projects/extension/src/background/ConnectionManager.ts
```diff
@@ -105,20 +105,32 @@ export class ConnectionManager extends (EventEmitter as { new(): StateEmitter })
...
       // if associate fails by returning false it has sent an error down the
       // port and disconnected it, so we should just discard this `AppMediator`
       if (app.associate()) {
         this.registerApp(app);
+        const appInfo = port.name.split('::');
+        notifMsg = `App ${appInfo[0]} connected to ${appInfo[1]}.`
       }
     }
+    chrome.storage.sync.get('notifications', (s) => {
+      s.notifications && chrome.notifications.create('', {
```
raoulmillais, line 127: Can you add a test in the `ConnectionManager` tests that mocks the chrome storage API and spies on `chrome.notifications.create` to verify this behaviour happens please?

raoulmillais (changes requested):

projects/extension/src/background/ConnectionManager.ts
```diff
@@ -105,20 +105,32 @@ export class ConnectionManager extends (EventEmitter as { new(): StateEmitter })
...
    */
   addApp(port: chrome.runtime.Port): void {
+    let notifMsg: string;
     if (!this.#client) {
       throw new Error('Smoldot client does not exist.');
     }
     const app = this.#apps.find(s => s.name === port.name);
     if (app) {
+      notifMsg = `App ${port.name} already exists.`;
       port.postMessage({ type: 'error', payload: `App ${port.name} already exists.` });
```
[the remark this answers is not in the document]
raoulmillais: You know when I said I hadn't added tests for all the error cases? Well this is where the bugs lie! :zipper_mouth_face:

projects/extension/src/background/ConnectionManager.ts
raoulmillais: Reading [this](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/notifications/create)  I think you should probably be using the name of the port as the `notificationId` .. this would clear any previous notifications about that port before displaying this one.

raoulmillais: I think changes to the `ConnectionManager` should be accompanied by tests.  There's already test infrastructure in place to test it.

raoulmillais: > Im not sure why I need the notification Id. (also Im using chrome API) In the example of Multi-network connect there are 3 notifications appearing one after the after denoting that the app connected to each different relay chain (polkadot, kusama, westend).

I don't think you *need* it but it doesn't feel right having an empty string. Using the port name as the id was my first instinct because it identifies the channel uniquely and it might prove useful debugging later.  You'll still get 3 notifications if you use the port name because the ids (for multi-network-demo) will be: `Multiple Network Demo::polkadot`, `Multiple Network Demo::kusama` and `Multiple Network Demo::westend`.

The MDN I linked to gets mapped to the chrome API, it's the same thing as `chrome.notifications.create`.  I just tend to prefer using the MDN documentation because:

1. Firefox is the lowest common denominator, it has the smallest API support out of chrome / firefox
2. It's easy to accidentally end up on the manifest v3 docs in chrome, which we don't support because of Firefox.

wirednkod: > I think changes to the ConnectionManager should be accompanied by tests. There's already test infrastructure in place to test it.

Yup - I deleted that part of my comment when I actually realised your point - but Github issues was polite enough to leave it there :) I will proceed with the test as it is irrelevant to the actual notification (that I want to test on #225 ) and more relevant to the connection manager action

raoulmillais (approved):

PR #454, opened 2021-08-12
fix extension breaking

wirednkod:


projects/extension/src/containers/Options.tsx
```diff
@@ -91,7 +91,7 @@ const Options: React.FunctionComponent = () => {
   }, []);
 
   useEffect(() => {
-    chrome.storage.sync.set({notifications: notifications}).catch(console.error);
+    void chrome.storage.sync.set({notifications: notifications});
```
raoulmillais, line 94: This API is supposed to be callback based not promise-based and firefox only supports callbacks if you use the chrome namespace.  I think the correct thing to do here is:

```
chrome.storage.sync.set({notifications: notifications}, () => {
  console.error(chrome.runtime.lastError);
});
```

https://developer.chrome.com/docs/extensions/reference/storage/
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Chrome_incompatibilities

raoulmillais (changes requested):

raoulmillais (approved):

Issue #442, opened 2021-08-06
Send proper error messages back to app when adding a chainspec fails in the extension


raoulmillais:
We swallow the error at the moment.  [See the TODO in the extension refactor PR](https://github.com/paritytech/substrate-connect/pull/434/files#diff-05bb44f78f8c40ea612398108dccd416ffc760bff6f63cfcd50e83b320948de3R194-R196)

PR #457, opened 2021-08-13
Improve ConnectionManager tests and fix notification behaviour

raoulmillais:
- Test that disconnect is called after sending an error when an app
  already exists
- Improve some of the test names
- Only show a notification to the user if an app successfully connects.  The app will get an error if it fails so it is the app's responsibility to update it's UI to tell the user.

This is the first part of #442. I'm splitting the PRs to make it easier to review instead of having several small unrelated changes mixed together.

### DIFF
--- a/projects/extension/src/background/ConnectionManager.ts
+++ b/projects/extension/src/background/ConnectionManager.ts
@@ -105,33 +105,35 @@ export class ConnectionManager extends (EventEmitter as { new(): StateEmitter })
    * by a content script.
    */
   addApp(port: chrome.runtime.Port): void {
-    let notifMsg: string;
     if (!this.#client) {
       throw new Error('Smoldot client does not exist.');
     }
-    const app = this.#apps.find(s => s.name === port.name);
-    if (app) {
-      notifMsg = `App ${port.name} already exists.`;
+
+    const existingApp = this.#apps.find(s => s.name === port.name);
+    if (existingApp) {
       port.postMessage({ type: 'error', payload: `App ${port.name} already exists.` });
       port.disconnect();
-    } else {
-      const app = new AppMediator(port, this as ConnectionManagerInterface)
-      // if associate fails by returning false it has sent an error down the
-      // port and disconnected it, so we should just discard this `AppMediator`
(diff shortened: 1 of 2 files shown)


projects/extension/src/background/ConnectionManager.ts
```diff
@@ -105,33 +105,35 @@ export class ConnectionManager extends (EventEmitter as { new(): StateEmitter })
...
+      const appInfo = port.name.split('::');
+      notifMsg = `App ${appInfo[0]} connected to ${appInfo[1]}.`
+
+      chrome.storage.sync.get('notifications', (s) => {
+        s.notifications && chrome.notifications.create(port.name, {
+          title: 'Substrate Connect',
+          message: notifMsg,
+          iconUrl: './icons/icon-32.png',
+          type: 'basic'
+        });
```
wirednkod, line 134:
```suggestion
    const app = new AppMediator(port, this as ConnectionManagerInterface)
    // if associate fails by returning false it has sent an error down the
    // port and disconnected it, so we should just discard this `AppMediator`
    if (app.associate()) {
      this.registerApp(app);
      const appInfo = port.name.split('::');
      chrome.storage.sync.get('notifications', (s) => {
        s.notifications && chrome.notifications.create(port.name, {
          title: 'Substrate Connect',
          message: `App ${appInfo[0]} connected to ${appInfo[1]}.`,
          iconUrl: './icons/icon-32.png',
          type: 'basic'
        });
```

wirednkod (changes requested):

projects/extension/src/background/ConnectionManager.ts
raoulmillais: I can't see what's going on in the diff .. is this just whitespace (indentation)?

raoulmillais: Oh no I see: you inlined the message. Why has github shown such a big diff? That's really unhelpful!

raoulmillais:
The build segfaulted .. that's a new one

wirednkod (approved):

PR #460, opened 2021-08-13
Detector And ExtensionProvider API refactor (parachain)

### LINKED ISSUE #441: Change Detector API to support parachains
```javascript
import { Detector } from '@substrate/connect';

const app = new Detector('burnr-wallet');
const westend = await app.connect('westend'); // get an API instance to talk to the relay chain (optional)
const statemine = await app.connect('westend', { chainName: 'statemine', chainSpec: '<chainspec>' });
await westend.rpc.chain.subscribeNewHeads((lastHeader) => {
  console.log(lastHeader.hash);
);

// etc ...
await statemine.disconnect();
await westend.disconnect();
```

---

We should also add a helper to allow app builders to detect when the extension is installed before `connect()`ing.  This is so that they can optionally show some UI to encourage people to install the extension.

```javascript
import { Detector } from '@substrate/connect';

const app = new Detector('burnr-wallet');
if (!app.hasExtension()) {
	// Show UI to encourage installing the extension
    // or disable the app etc
}
```

wirednkod:
Adding implementation on Detector and ExtensionProvider Api in order to support app connectivity to parachains.

- Remove caret from smoldot in package.jsons in order to avoid update to latest version until health check is done
- Add a helper (hasExtension) to allow app builders to detect when the extension is installed before connect()ing.
- Update ts-docs in order to depict API changes
- Update tests for Detector, ExtensionProvider and ConnectionManager to cover new changes.

Fixes #441


packages/connect/src/Detector.ts
```diff
@@ -59,6 +60,14 @@ export class Detector {
     return this.#name;
   }
 
+  /**
+   * Returns a boolean showing if the app that was created is interacting with 
+   * an extension or not
```
raoulmillais, line 65:
```
Returns `true` if the user has the extension installed otherwise false

@remarks

You should check this if you wish to show UI to encourage users to download and install the extension or if you wish your app only to work with the extension installed.
```

packages/connect/src/Detector.ts
```diff
@@ -82,10 +91,14 @@ export class Detector {
    * Alternatively you may supply a chain spec and options to connect to a
    * custom chain with a light client (in-page only).
    *
-   * @param chainName - the name of the blockchain network to connect to
-   * @param chainSpec - an optional chainSpec to connect to a different network
-   * @param options - any extra API options to passed to PolkadotJS when
-   * constructing it.
+   * @param relay - param of ChainInfo or string type. In case of ChainInfo,
```
raoulmillais, line 94: use `{@link ChainInfo}` syntax to reference other items in the docs.

raoulmillais (changes requested):
LGTM :+1: Apart from a couple of nitpicks on the documentation

raoulmillais (approved):

PR #461, opened 2021-08-16
Reject pending requests on disconnect

raoulmillais:
When the extension disconnects an app for an error the `ExtensionProvider` now rejects any pending req	uest promises.  This is a direct copy of the logic from the WsProvider

### DIFF
--- a/packages/connect/src/ExtensionProvider/ExtensionProvider.ts
+++ b/packages/connect/src/ExtensionProvider/ExtensionProvider.ts
@@ -9,7 +9,7 @@ import {
 } from '@polkadot/rpc-provider/types';
 import { logger } from '@polkadot/util';
 import EventEmitter from 'eventemitter3';
-import { isUndefined } from '../utils/index.js';
+import { isUndefined, eraseRecord } from '../utils/index.js';
 import { HealthCheckError } from '../errors.js';
 import {
   MessageFromManager,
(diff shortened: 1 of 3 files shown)


packages/connect/src/ExtensionProvider/ExtensionProvider.ts
```diff
@@ -47,6 +47,17 @@ interface HealthResponse {
   shouldHavePeers: boolean;
 }
 
+function eraseRecord<T> (record: Record<string, T>, cb?: (item: T) => void): void {
```
wirednkod, line 50: This should be in utils directory. 
Also respective tests should be added for the function

wirednkod (changes requested):

packages/connect/src/ExtensionProvider/ExtensionProvider.ts
raoulmillais: I copied it straight from @polkadot/api: https://github.com/polkadot-js/api/blob/c212db26b664a339dc2aa68f6380c40eb72595fb/packages/rpc-provider/src/ws/index.ts#L45-L53

wirednkod: As soon as this is now in substrate-connect's codebase it should be placed along the rest of utils, be documented and tested.

wirednkod (approved):

PR #462, opened 2021-08-17
Replace smoldot with @substrate/smoldot-light

wirednkod:
- replacing deprecated smoldot package with @substrate/smoldot-light
- adds in `packages/smoldot-test-utils/src/index.ts` the mock for healthChecker

### DIFF
--- a/packages/connect/src/SmoldotProvider/SmoldotProvider.ts
+++ b/packages/connect/src/SmoldotProvider/SmoldotProvider.ts
@@ -8,7 +8,7 @@ import {
 } from '@polkadot/rpc-provider/types';
 import { assert, logger } from '@polkadot/util';
 import EventEmitter from 'eventemitter3';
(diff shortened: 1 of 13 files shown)


packages/connect/src/SmoldotProvider/SmoldotProvider.ts
```diff
@@ -8,7 +8,7 @@ import {
 } from '@polkadot/rpc-provider/types';
 import { assert, logger } from '@polkadot/util';
 import EventEmitter from 'eventemitter3';
-import * as smoldot from 'smoldot';
+import * as smoldot from '@substrate/smoldot-light';
```
raoulmillais, line 11: NOt for right now, but it would be good to stop doing this `import * as smoldot...` everywhere so that we don't have all this repetetive `smoldot.SmoldotChain` (etc) usages

raoulmillais (approved):

PR #471, opened 2021-08-23
Remove AppMediator::associate

tomaka:
Instead of doing `const app = new AppMediator(...)` followed with `app.associate()`, this PR merges these two functions into one.
In other words, the constructor implicitly calls what used to be `associate()`, and throws if `associate()` fails.

### DIFF
--- a/projects/extension/src/background/AppMediator.ts
+++ b/projects/extension/src/background/AppMediator.ts
@@ -31,11 +31,11 @@ export class AppMediator extends (EventEmitter as { new(): StateEmitter }) {
   readonly #name: string;
   readonly #appName: string;
   readonly #port: chrome.runtime.Port;
+  readonly #chainName: string;
   // REM: what to do about the fact these might be undefined?
   readonly #tabId: number | undefined;
   readonly #url: string | undefined;
   readonly #manager: ConnectionManagerInterface;
-  #chainName: string | undefined  = undefined;
   #chain: SmoldotChain | undefined;
   #state: AppState = 'connected';
   #pendingRequests: string[] = [];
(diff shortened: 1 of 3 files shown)


projects/extension/src/background/ConnectionManager.ts
```diff
@@ -133,6 +133,8 @@ export class ConnectionManager extends (EventEmitter as { new(): StateEmitter })
           type: 'basic'
         });
       });
+    } finally {
```
wirednkod, line 136: Why do you need a try/finally here?

wirednkod (commented):
lgtm besides one non-blocking comment

projects/extension/src/background/ConnectionManager.ts
tomaka: I need a `try` block in order to silently discard problematic ports, otherwise the whole extension would stop if a port contains a wrong name.
`try { ... }` alone isn't valid JS, you need either `catch` or `finally`, so I put `finally`.

wirednkod (approved):

projects/extension/src/background/AppMediator.ts
```diff
@@ -45,10 +45,23 @@ export class AppMediator extends (EventEmitter as { new(): StateEmitter }) {
...
   constructor(port: chrome.runtime.Port, manager: ConnectionManagerInterface) {
     super();
-    this.#appName = port.name.substr(0, port.name.indexOf('::'));
+
+    const splitIdx = port.name.indexOf('::');
+    if (splitIdx === -1) {
+      const error: MessageFromManager = { type: 'error', payload: `Invalid port name ${port.name} expected <app_name>::<chain_name>` };
+      port.postMessage(error);
+      port.disconnect();
+      throw new Error();
```
raoulmillais, line 60: Can we add a message to the error here and then `console.warn` in a `catch` block with the message below instead of using `finally` please?  Then we won't swallow unexpected error messages and we can debug problems locally more easily.

projects/extension/src/background/ConnectionManager.ts
raoulmillais: See my review comment about using `catch` and adding an error message.

raoulmillais (changes requested):

projects/extension/src/background/AppMediator.ts
tomaka: I went for silencing the error because it's already sent back through the port to the user. The only advantage of also printing this to the console would be to debug situations where the code in the `ExtensionProvider` isn't working properly.

I think the code here should be redesigned to not directly send back the error to the port, in which case we'd indeed catch that error in order to send it back and not just discard it silently, and I'd do that on top of this port.

raoulmillais: The problem I want to avoid is that the `finally` below will swallow *all* errors.  If we warn in a `catch` block then we won't inadvertently swallow something unexpected.  Adding a message here helps disambiguate unexpected from programmer error by the app developer.

raoulmillais (approved):
LGTM :+1: Feel free to merge

PR #472, opened 2021-08-23
Remove subscription field from MessageToManager

tomaka:
This field is useless and shouldn't be there in the first place.


wirednkod (approved):

tomaka:
I'm merging as an "obvious" change. I don't think there's anything controversial.

PR #473, opened 2021-08-24
Integrate HealthChecker in Extension

wirednkod:
- Bump version of smoldot light to 0.4.4
- Add implementation of HealthChecked in Extension: Each Chain should have 1 HealthCheck. As a result the return of the addChain in ConnectionManager is a ChainInstance object that contains a) chain: the addedChain and b) healthChecker for the specific chain. This way the manipulation of the HealthChecker is passed in the AppMediator.
- Alter tests for healthCheck.

### DIFF
--- a/projects/extension/src/background/AppMediator.ts
+++ b/projects/extension/src/background/AppMediator.ts
@@ -169,8 +169,8 @@ export class AppMediator extends (EventEmitter as { new(): StateEmitter }) {
         this.#chain = o.chain
         this.#healthChecker = o.healthChecker;
         // eslint-disable-next-line @typescript-eslint/unbound-method
-        this.#healthChecker.setSendJsonRpc(this.#chain.sendJsonRpc);
-        this.#healthChecker.start(this.#healthCheckCallback);
+        this.#chain && this.#healthChecker?.setSendJsonRpc(this.#chain.sendJsonRpc);
+        this.#healthChecker?.start(this.#healthCheckCallback);
         // process any RPC requests that came in while waiting for `addChain`
         // to complete
         if (this.#pendingRequests.length > 0) {
--- a/projects/extension/src/background/AppMediator.ts
+++ b/projects/extension/src/background/AppMediator.ts
@@ -6,9 +6,10 @@ import {
 import {
   AppState,
   ConnectionManagerInterface,
+  HealthResponse,
   StateEmitter,
 } from './types';
-import { SmoldotChain } from '@substrate/smoldot-light';
+import { SmoldotChain, HealthChecker } from '@substrate/smoldot-light';
 import westend from '../../public/assets/westend.json';
 import kusama from '../../public/assets/kusama.json';
 import polkadot from '../../public/assets/polkadot.json';
@@ -39,6 +40,8 @@ export class AppMediator extends (EventEmitter as { new(): StateEmitter }) {
   #chain: SmoldotChain | undefined;
   #state: AppState = 'connected';
   #pendingRequests: string[] = [];
+  #healthChecker: HealthChecker | undefined = undefined;
+  #healthStatus: HealthResponse | undefined = undefined;
 
   /**
    * @param port - the open communication port between the app's content page
@@ -98,6 +101,20 @@ export class AppMediator extends (EventEmitter as { new(): StateEmitter }) {
     return this.#appName;
   }
 
+  /** healthStatus returns the latest health status
+   * of app as set from the callback
+   */
+  get healthStatus(): HealthResponse {
+    return this.#healthStatus as HealthResponse;
+  }
+
+  /** 
+  * returns the healthChecker that the app is using
+  */
+  get healthChecker(): HealthChecker {
+    return this.#healthChecker as HealthChecker;
+  }
+
   /** 
    * chainName is the name of the chain to talk to; this is the
    * name of the blockchain network.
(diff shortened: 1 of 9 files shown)


projects/extension/src/background/AppMediator.ts
```diff
@@ -133,26 +150,34 @@ export class AppMediator extends (EventEmitter as { new(): StateEmitter }) {
...
+  }
+
   #handleSpecMessage = (msg: MessageToManager, chainName: string): void => {
     const chainSpec: string = relayChains.has(chainName) ?
       (relayChains.get(chainName) || '') : msg.payload;
 
     const rpcCallback = (rpc: string) => {
-      this.#port.postMessage({ type: 'rpc', payload: rpc })
+      const rpcResp = this.#healthChecker?.responsePassThrough(rpc) || rpc;
+      this.#port.postMessage({ type: 'rpc', payload: rpcResp })
```
tomaka, line 163:
```suggestion
      const rpcResp = this.#healthChecker?.responsePassThrough(rpc);
      if (rpcRsp)
            this.#port.postMessage({ type: 'rpc', payload: rpcResp })
```

projects/extension/src/background/AppMediator.ts
```diff
@@ -98,6 +101,20 @@ export class AppMediator extends (EventEmitter as { new(): StateEmitter }) {
...
+  get healthStatus(): HealthResponse {
+    return this.#healthStatus as HealthResponse;
+  }
+
+  /** 
+  * returns the healthChecker that the app is using
+  */
+  get healthChecker(): HealthChecker {
+    return this.#healthChecker as HealthChecker;
+  }
```
tomaka, line 116: This exposes an implementation detail. Is it really necessary?

projects/extension/src/background/types.ts
```diff
@@ -3,6 +3,17 @@ import { AppMediator } from './AppMediator';
...
 
+export interface HealthResponse {
+  isSyncing: boolean;
+  peers: number;
+  shouldHavePeers: boolean;
+}
+
+export interface ChainInstance {
+  chain: smoldot.SmoldotChain
+  healthChecker: smoldot.HealthChecker
```
tomaka, line 14: Similarly, this exposes an implementation detail, I don't see why you would ever need this field.

tomaka (changes requested):

projects/extension/src/background/AppMediator.ts
wirednkod: No it is not. Remained

projects/extension/src/background/types.ts
wirednkod: This is needed from both mocks and ConnectionManager. I can though duplicated it in relevant files.

projects/extension/src/background/AppMediator.ts
tomaka: The tests are whatever you want them to be

wirednkod: I agree that this should not be exposed. This means that If this is removed - relevant tests should be removed as well as they will not work without it.

wirednkod: As agreed this will be removed in order to not expose impl details and along will be removed the relevant tests

projects/extension/src/background/types.ts
tomaka: This should be removed as well, consequently. There's no need for anything outside `AppMediator` to know about the health checker.

wirednkod: This is already removed

projects/extension/src/background/ConnectionManager.ts
```diff
@@ -187,14 +192,21 @@ export class ConnectionManager extends (EventEmitter as { new(): StateEmitter })
...
+    chainSpec: string,
+    jsonRpcCallback: SmoldotJsonRpcCallback,
+    relayChainName?: string): Promise<ChainInstance> {
     if (!this.#client) {
       throw new Error('Smoldot client does not exist.');
     }
     let relay: Network | undefined = undefined;
     let addChainOptions = {} as SmoldotAddChainOptions;
 
+    const healthChecker = await (smoldot as any).healthChecker();
```
tomaka, line 209: This isn't removed

wirednkod: This is the init of the healthChecker that happens for each chain (1 chain = 1 healthChecker). Why would I remove it?
Its initialized here and then appMediator is using it in order to setup the setJsonRpc, start the healthchecker and stop it as needed

tomaka (approved):

projects/extension/src/background/types.ts
```diff
@@ -3,6 +3,13 @@ import { AppMediator } from './AppMediator';
 import EventEmitter from 'eventemitter3';
 import StrictEventEmitter from 'strict-event-emitter-types';
 
+export interface HealthResponse {
```
raoulmillais, line 6: We should just use the `SmoldotHealth` type from @substrate/smoldot-light rather than add a separate new type of our own.  The shape is identical

raoulmillais (changes requested):

projects/extension/src/background/AppMediator.ts
```diff
@@ -98,6 +107,13 @@ export class AppMediator extends (EventEmitter as { new(): StateEmitter }) {
     return this.#appName;
   }
 
+  /** healthStatus returns the latest health status
+   * of app as set from the callback
+   */
+  get healthStatus(): smoldot.SmoldotHealth {
```
raoulmillais, line 113: Why not add `SmoldotHealth` to the other import then you can avoid the repetitious `smoldot.SmoldotHealth`? It's not a blocker but it would be neater

raoulmillais (approved):

Issue #451, opened 2021-08-11
Unique list of connected networks


wirednkod:
With smoldot [v0.3.x](https://gist.github.com/tomaka/764fa8f9f51cddabfb616b37d536340a) only 1 smoldot client is used and multiple chains with smoldot being clever enough to identify and reuse chains.
Unique chains (e.g. kusama, polkadot ,westend) need to be kept in extension for single points of reference.

At the moment due to v.0.3 the networks are multiple based on amount of apps/chains connected through extension.

![multiples](https://user-images.githubusercontent.com/5408605/129004091-7c259593-4fad-4189-ba2b-67fe8da25e29.png)

PR #475, opened 2021-09-02
Remove sending the relay chain name to the extension

tomaka:
This removes passing the `relayChainName` from the foreground to the extension. Which relay chain a parachain should connect to is already found in the chain specification, so there's no need to additionally pass the same value as a separate parameter.

I've noticed that the `connect` method of the detector is broken, but it's already broken before this PR and can be fixed afterwards.

### DIFF
--- a/packages/connect-extension-protocol/src/index.ts
+++ b/packages/connect-extension-protocol/src/index.ts
@@ -107,8 +107,6 @@ export interface MessageToManager {
   type: 'rpc' | 'spec';
   /** Payload of the message -  a JSON encoded RPC request **/
   payload: string;
-  /** The name of the relay chain - if provided then the chain is a parachain  **/
-  relayChainName?: string;
 }
 
 /**
--- a/packages/connect/src/Detector.ts
+++ b/packages/connect/src/Detector.ts
@@ -115,14 +115,10 @@ export class Detector {
    */
   public connect = async (relay: ChainInfo | string, parachain?: ChainInfo, options?: ApiOptions): Promise<ApiPromise> => {
     const chain: ChainInfo | string = parachain || relay;
-    let relayName = "";
-    if (parachain) {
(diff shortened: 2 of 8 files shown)


wirednkod (commented):
@tomaka Extension needs to know the correlation between relayChain and parachain for the Options page - in order to show the structure (which parachain is connected to which chain).
In addition - once healthcheck is merged - based on #451 the networks must be updated as unique and have the correlation with the parachains instead of the current structure

tomaka:
> Extension needs to know the correlation between relayChain and parachain for the Options page - in order to show the structure (which parachain is connected to which chain).

The relay chain ID can be found in the `relay_chain` field of the chain specification of a parachain, there's no need to pass it a second time. Example: https://github.com/paritytech/smoldot/blob/88ff678048378c6b61e9087fffe1278c6e26be34/bin/westend-westmint.json#L13

wirednkod:
[Here](https://github.com/paritytech/substrate-connect/pull/475/files#diff-05bb44f78f8c40ea612398108dccd416ffc760bff6f63cfcd50e83b320948de3R200) all chains that exist in network are mapped as potentialRelayChains even if this is a relayChain. I have understood from the [gist](https://gist.github.com/tomaka/764fa8f9f51cddabfb616b37d536340a) that this was needed only for parachain and not for relay ones. I assume based on the code here that smoldot will understand the relay chain and will not use the `potentialRelayChains` field. If thats the case then this lgtm.

raoulmillais (changes requested):
Can you fix the linting errors please?

tomaka:
Done

raoulmillais (approved):
LGTM :+1: Feel free to merge.  Btw, I noticed your commit message said "Maybe fix lints". You can check locally with `yarn lint`

PR #476, opened 2021-09-06
Change demo licenses to Apache 2

raoulmillais:


wirednkod (changes requested):
We should also change all the LICENCE files to Apache ones

wirednkod (approved):

PR #478, opened 2021-09-06
Fix small stupidities regarding chainName in AppMediator

tomaka:
No longer pass the chain name to `handleSpecMessage` since it's stored in `this` anyway.


raoulmillais (approved):
Thanks... I think this is leftover from an automated refactoring extracting the method.

wirednkod (approved):

PR #481, opened 2021-09-07
Remove initSmoldot function

tomaka:
Removes the `initSmoldot` function from the `ConnectionManager` and puts the smoldot initialization in the constructor.

`initSmoldot` was an asynchronous function. In order to remove it, I changed the type of the `client` field from a `SmoldotClient | undefined` to a `Promise<SmoldotClient>`. Removing the ` | undefined` was the main motivation behind this change.

For what it's worth, `smoldot.start` can never throw any error right now. It is an asynchronous method but it doesn't do anything asynchronously. If I think about it, I could turn it into a non-asynchronous function when publishing a new major version.

### DIFF
--- a/projects/extension/src/background/ConnectionManager.test.ts
+++ b/projects/extension/src/background/ConnectionManager.test.ts
@@ -18,9 +18,7 @@ const doNothing = () => {
 
 test('adding and removing apps changes state', async () => {
   //setup connection manager with 2 chains
-  const manager = new ConnectionManager();
-  manager.smoldotLogLevel = 1;
-  await manager.initSmoldot();
+  const manager = new ConnectionManager(1);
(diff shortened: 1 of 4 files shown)


tomaka:
This PR also fixes a potential minor corner case: if you called `initSmoldot`, not `await`ing on the promise, and calling `shutdown` before this promise finishes, then the client would never be terminated because it was still `undefined` at that time.

tomaka:
There is a small change in behaviour: before this PR smoldot is only initialized as a response to a `chrome.runtime.onInstalled` or `chrome.runtime.onStartup` event. After this PR it is always initialized immediately.

tomaka:
I did this PR in a rush. Let me reopen it when it's properly finished.

PR #485, opened 2021-09-10
Parachain demo

wirednkod:
- Update detector API and providers to receive parachain stringified chainspec - and retrieve the parachain name from chainspec itself instead of user providing it
- Update ExtensionProvider to be aligned with Detector; In case of parachains MessageToManager was enriched with optional parachainPayload in order to pass on same API call the parachain specs
- upgrade smoldot-light to 0.4.11 to support querying parachains
- Fix UI extension details that was breaking Icons in Options page
- Alter extension to support parachains
- Update tests for parachain support
- Create Parachain Demo app that follows the Smoldot-demo simplicity but connects to Parachain instead of Relay chain - Connecting to Westmint and receive and display the first 10 tokens from it


projects/extension/src/background/AppMediator.ts
```diff
@@ -21,9 +21,9 @@ import polkadot from '../../public/assets/polkadot.json';
 type RelayType = Map<string, string>;
 
 export const relayChains: RelayType = new Map<string, string>([
-  ["polkadot", JSON.stringify(polkadot)],
-  ["kusama", JSON.stringify(kusama)],
-  ["westend", JSON.stringify(westend)]
+  ['polkadot', JSON.stringify(polkadot)],
+  [kusama.id, JSON.stringify(kusama)],
+  [westend.id, JSON.stringify(westend)]
```
tomaka, line 26:
```suggestion
  ["polkadot", JSON.stringify(polkadot)],
  ["kusama.id", JSON.stringify(kusama)],
  ["westend.id", JSON.stringify(westend)]
```

These strings are part of the API of substrate-connect. The values found in the chain spec are an implementation detail.

projects/extension/src/background/ConnectionManager.ts
```diff
@@ -188,31 +188,53 @@ export class ConnectionManager extends (EventEmitter as { new(): StateEmitter })
...
-      isKnown: true,
-      chainspecPath: `${name}.json`
-    });
-
-    return addedChain;
+    if (relayChain) {
+      const addedChain = await this.#client.addChain({
+        chainSpec,
+        jsonRpcCallback,
+        potentialRelayChains: [relayChain],
```
tomaka, line 207: Again, this should be all chains that have been previously added by the same uApp. There's no need for a `relayChain` parameter.

projects/extension/src/background/ConnectionManager.ts
```diff
@@ -188,31 +188,53 @@ export class ConnectionManager extends (EventEmitter as { new(): StateEmitter })
...
+   * @param relayChain - optional SmoldotChain for relay chain
    * 
    * @returns addedChain - An the newly added chain info
    */
   async addChain(
-    name: string,
     chainSpec: string,
-    jsonRpcCallback: SmoldotJsonRpcCallback): Promise<SmoldotChain> {
+    jsonRpcCallback?: SmoldotJsonRpcCallback,
+    relayChain?: SmoldotChain): Promise<SmoldotChain> {
```
tomaka, line 198:
```suggestion
    ): Promise<SmoldotChain> {
```

projects/extension/src/background/AppMediator.ts
```diff
@@ -146,34 +146,65 @@ export class AppMediator extends (EventEmitter as { new(): StateEmitter }) {
...
+    } else {
+      // Connect the main Chain only
+      this.#manager.addChain(chainSpec, rpcCallback)
+        .then(chain => {
+          this.#chain = chain;
+          this.#initHealthChecker();
+        }).catch(e => {
+          this.#handleError(e)
+        });
+    }
```
josepot, line 207: A bit of a nit-pick, but IMO this part could be simplified a bit:

```suggestion
    let chainPromise: Promise<SmoldotChain>
    
    // Means this is a parachain trying to connect
    if (msg.parachainPayload) {
      // Connect the main Chain first and on success the parachain with the chain
      // that just got connected as the relayChain
      const relayChainName: string = JSON.parse(msg.parachainPayload).relay_chain
      const parachainSpec: string = msg.parachainPayload
      const relaychainSpec: string | undefined = relayChains.get(relayChainName)

      if (!relaychainSpec)
        throw new Error('Relay chain spec was not found')

      chainPromise =  this.#manager.addChain(chainSpec).then(relayChain =>
        this.#manager.addChain(parachainSpec, rpcCallback, relayChain)
      )
    } else {
      // Connect the main Chain only
      chainPromise = this.#manager.addChain(chainSpec, rpcCallback)
    }
    
    chainPromise.then(chain => {
      this.#chain = chain;
      this.#initHealthChecker();
    }).catch(e => {
      this.#handleError(e)
    });
```

projects/parachain-demo/src/index.ts
```diff
@@ -0,0 +1,75 @@
...
+            }).catch(error => {
+              ui.error(error);
+              if (!resolved) {
+                resolved = true;
+                reject();
+              }
+            });
+          }, 2000);
+        });
+      }
```
josepot, line 63: The `setInterval` is never cleared and the error is not propagated through `reject`.

The following implementation addresses both things:

```suggestion
      const wait = (ms: number) => new Promise<void>(res => { setTimeout(res, ms); })

      const waitForChainToSync = async () => {
        const health = await api.rpc.system.health();

        if (health.isSyncing.eq(false)) {
          ui.showSynced();
        } else {
          ui.showSyncing();
          await wait(2000);
          await waitForChainToSync();
        }
      }
```

Also, in case someone is wondering if this could cause a stackoverflow: it won't because the `wait` function creates a macrotask, which "resets" the event-loop.

projects/extension/src/background/ConnectionManager.test.ts
```diff
@@ -237,46 +237,6 @@ describe('When the manager is shutdown', () => {
...
-    chrome.notifications.create.mockClear();
-    manager.smoldotLogLevel = 1;
-    await manager.initSmoldot();
-  });
-
-  afterEach( () => {
-    manager.shutdown();
-  })
-
-  test('Checks storage for notifications preferences', () => {
```
raoulmillais, line 258: We should reinstate these 2 tests.  You'll need to spy on addChain to trigger the callback

raoulmillais (changes requested):

projects/extension/src/background/ConnectionManager.test.ts
wirednkod: The reason why these were removed is that the notification was added inside the private `#handleSpecMessage` in order to be initiated on correct timing and have the correct information in it.This is triggered through the port of the specific App. In all of our tests we are [mocking the port](https://github.com/paritytech/substrate-connect/blob/69eb7b8c02ff89e8681b481d52fdae10a988ac31/projects/extension/src/background/ConnectionManager.test.ts#L24) - so I am not really sure how the spy on addChain would help. Notifications could be tested with Cypress on a more end to end test suite (instead of a unit one) probably.

raoulmillais: You can add a spy on `ConnectionManager.addChain` on your instance of the `ConnectionManager` just before sending the spec message using your `MockPort` and configure your spy to return a resolved promise then you can check that the mock versions of the chrome API were called appropriately.  This will test the logic in isolation and run much faster than a fullstack test

packages/connect/src/ExtensionProvider/ExtensionProvider.ts
```diff
@@ -321,6 +333,7 @@ export class ExtensionProvider implements ProviderInterface {
     this.#connectionStatePingerId = setInterval(this.#checkClientPeercount, 
       this.healthPingerInterval);
 
+    // this.#isConnected = true;
     return Promise.resolve();
```
raoulmillais, line 337: This shouldn't be commented out

projects/extension/src/background/ConnectionManager.ts
```diff
@@ -229,78 +229,143 @@ export class ConnectionManager extends (EventEmitter as { new(): StateEmitter })
...
+      const relayId: string = JSON.parse(msg.parachainPayload).relay_chain;
+      const parachainSpec: string = msg.parachainPayload;
+
+      const relaychainSpec: string | undefined = relayChains.get(
+        nameIdMapper.get(relayId) || ""
+      );
+
+      if (!relaychainSpec) {
+        const error: Error = new Error("Relay chain spec was not found");
+        this.#handleError(app, error);
```
raoulmillais, line 331: You need to return here to stop further processing of the message

raoulmillais (changes requested):

projects/extension/src/background/ConnectionManager.ts
raoulmillais: This ought to have a test. Send a spec message with a chainspec containing an invalid relay chain and check the port gets sent an error and disconnected

packages/connect/src/ExtensionProvider/ExtensionProvider.ts
wirednkod: This actually should be removed - It is a stray-workaround for checking when there were no Peers - I will remove it

PR #486, opened 2021-09-14
Refactor connection manager and remove AppMediator

wirednkod:
- Update smoldot light to 0.4.6
- Merge AppMediator and ConnectionManager
- Update ConnectionManagerTests and remove AppMediatorTests
- Fix minor popup UI issues (moved to PR #487 )


projects/extension/src/background/index.ts
```diff
@@ -35,7 +35,7 @@ const init = async () => {
       const rpcCallback = (rpc: string) => {
         console.warn(`Got RPC from ${key} dummy chain: ${rpc}`);
       };
-      await manager.addChain(key, value, rpcCallback)
+      await manager.addChain(0, key, value, rpcCallback)
```
tomaka, line 38: Note that this could directly be done in the constructor of the `ConnectionManager` or in `initSmoldot`. This would avoid passing this `0` that makes no sense.
The three relay chains also don't need to be put in `#networks`.

tomaka:
Is it not possible to open one PR per thing? One PR for updating the Westend spec, one PR for the connection manager/appmediator change, one PR for the UI changes, etc.?

wirednkod:
> Is it not possible to open one PR per thing? One PR for updating the Westend spec, one PR for the connection manager/appmediator change, one PR for the UI changes, etc.?

Not really - Besides the Westend spec - all the rest are interconnected. If I open 1 PR for manager/appmediator change, the UI will produce errors,

projects/extension/src/components/ClientItem.tsx
```diff
@@ -117,23 +106,12 @@ const ClientMenu: React.FunctionComponent<Network | Parachain> = ({isKnown, name
...
-          { isKnown
-          ? <InputButton>
-            <Publish fontSize='small'/>
-          </InputButton>
-          : <InputButton>
-            <SystemUpdateAlt fontSize='small' />
-          </InputButton>
-          }
-
         </InputWrap>
```
tomaka, line 115: These changes are clearly unrelated to the ConnectionManager/AppMediator change, aren't they?

wirednkod: These ones I am already putting aside on different PR - But there is a slight alternation of UI (on options.tsx) that needs to be in this PR (related to the tabId introduced in AppProps), which is what I was referring to, at my comment above

projects/extension/src/background/index.ts
wirednkod: IMO the three relay chains should be in the networks - cause the Options page of the extension is "reading" from `#networks`. When no app is on and user goes to that page - he needs to see the "default" chains that exist in smoldot.
I have removed the 0 and default adding it inside `addChain` when `tabId` is not provided.

projects/extension/src/background/ConnectionManager.ts
```diff
@@ -194,25 +236,119 @@ export class ConnectionManager extends (EventEmitter as { new(): StateEmitter })
...
-    });
+    }
+
+    this.#networks.push(network);
+
+    return network;
+  }
+
+  /** Handles the incoming message that contains Spec. */
+  #handleSpecMessage = (msg: MessageToManager, app: AppProps): void => {
```
raoulmillais, line 269: You've missed converting the test for this.  It should be easy to do- you want to check that the `app.chain` has been set after sending the spec message (similar setup to the 'App already disconnected' test but checking what happens after the spec message

projects/extension/src/background/ConnectionManager.ts
```diff
@@ -194,25 +236,119 @@ export class ConnectionManager extends (EventEmitter as { new(): StateEmitter })
...
+      .then(network => {
+        app.chain = network.chain;
+        // eslint-disable-next-line @typescript-eslint/unbound-method
+        app.healthChecker?.setSendJsonRpc(app.chain.sendJsonRpc);
+        app.healthChecker?.start((health: SmoldotHealth) => {
+          app.healthStatus = health;
+        });
+        // process any RPC requests that came in while waiting for `addChain`
+        // to complete
+        if (this.#pendingRequests.length > 0) {
```
raoulmillais, line 289: Also missed the test that messages sent before addChain resolves don't get lost. Again should be quick and easy to port ... fake posting some RPC messages before you `waitForMessageToBePosted` after the spec message

raoulmillais (changes requested):
General implementation changes look good to me.  Id' call`AppProps` just `App` .. its simpler :smile:  I commented inline on a couple of missing tests. At a glance they are really straightforward to port over to the new implemntations.  There are 2 further test cases missing but its harder to test those with this new implementation without adjusting the mocks.

projects/extension/src/background/ConnectionManager.ts
wirednkod: Actually I have not - in tests now I am using the actual ConnectionManager and not the ConnectionManagerMock. I will try to figure out a way to test these, if possible with the ConnectionManager or else I will need to adjust the ConnectionManagerMock. Im not sure though if these should be tested with a mock - these are actually covered from our sample apps (that we use for test)

projects/extension/src/background/ConnectionManager.ts
wirednkod: Actually I have not - in tests now I am using the actual ConnectionManager and not the ConnectionManagerMock. I will try to figure out a way to test these, if possible with the ConnectionManager or else I will need to adjust the ConnectionManagerMock. Im not sure though if these should be tested with a mock - these are actually covered from our sample apps (that we use for test)

raoulmillais (approved):

PR #491, opened 2021-09-16
Extension cleanup and improvements

### LINKED ISSUE #451: Unique list of connected networks

wirednkod:
Fixes #451 

Outcome of extension Options's page can be seen at the screenshot below:

image

### DIFF
--- a/packages/connect/src/SmoldotProvider/SmoldotProvider.ts
+++ b/packages/connect/src/SmoldotProvider/SmoldotProvider.ts
@@ -32,7 +32,9 @@ interface StateSubscription extends SubscriptionHandler {
 }
 
 interface HealthResponse {
+  //indicates whether GrandPa wrap syncing is done 
   isSyncing: boolean;
+  //indicates the amount of connected peers 
   peers: number;
   shouldHavePeers: boolean;
 }
--- a/packages/connect/src/SmoldotProvider/SmoldotProvider.ts
+++ b/packages/connect/src/SmoldotProvider/SmoldotProvider.ts
@@ -260,7 +260,7 @@ export class SmoldotProvider implements ProviderInterface {
     assert(!this.#client && !this.#isConnected, 'Client is already connected');
     try {
       this.#client = await this.#smoldot.start({
-        forbidWs: false, /* suppress console warnings about insecure connections */
+        forbidWs: false,
         maxLogLevel: 3, /* no debug/trace messages */
       });
       this.#chain = await this.#client.addChain({
--- a/packages/connect/src/SmoldotProvider/SmoldotProvider.ts
+++ b/packages/connect/src/SmoldotProvider/SmoldotProvider.ts
@@ -260,7 +260,7 @@ export class SmoldotProvider implements ProviderInterface {
     assert(!this.#client && !this.#isConnected, 'Client is already connected');
     try {
       this.#client = await this.#smoldot.start({
-        forbidWs: true, /* suppress console warnings about insecure connections */
+        forbidWs: false, /* suppress console warnings about insecure connections */
         maxLogLevel: 3, /* no debug/trace messages */
       });
       this.#chain = await this.#client.addChain({
--- a/projects/extension/src/background/ConnectionManager.test.ts
+++ b/projects/extension/src/background/ConnectionManager.test.ts
@@ -9,6 +9,7 @@ import kusama from '../../public/assets/kusama.json';
 import { MockPort } from '../mocks';
 import { chrome } from 'jest-chrome';
 import { App } from './types';
+import { Client } from '../types';
 
 let port: MockPort;
 let manager: ConnectionManager;
@@ -195,7 +196,7 @@ describe('Unit tests', () => {
   test('Get networks/chains', () => {
     // With this look the "chain" is removed intentionally as "chain"
     // object cannot be compared with jest 
-    const tmpChains = manager.registeredClients.map((n: any) => (
+    const tmpChains = manager.registeredClients.map((n: Client) => (
       {
         name: n.name,
         status: n.status
--- a/projects/extension/src/background/ConnectionManager.test.ts
+++ b/projects/extension/src/background/ConnectionManager.test.ts
@@ -183,8 +183,8 @@ describe('Unit tests', () => {
 
   test('Get registered clients', () => {
     expect(manager.registeredClients).toEqual([
-      "westend",
-      "kusama"
+      { name: 'westend', status: "connected" },
+      { name: 'kusama', status: "connected" }
     ]);
   });
 
@@ -195,7 +195,7 @@ describe('Unit tests', () => {
   test('Get networks/chains', () => {
     // With this look the "chain" is removed intentionally as "chain"
     // object cannot be compared with jest 
-    const tmpChains = manager.networks.map(n => (
+    const tmpChains = manager.registeredClients.map((n: any) => (
       {
         name: n.name,
         status: n.status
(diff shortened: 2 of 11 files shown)


projects/extension/src/components/NetworkTab.tsx
```diff
@@ -109,58 +109,47 @@ const useStyles = makeStyles((theme: Theme) =>
...
-    </div>) : (
+  const peers = health && health.peers;
+  const status = health && health.status;
+  const shouldHavePeers = health && health.shouldHavePeers;
+  const isSyncing = health && health.isSyncing;
+  return (
     <div className={classes.info}>
-      <p>{emojis.chain} Syncing <b>{chainName}</b></p>
+      <p>{emojis.chain} Chain is {status} and should {!shouldHavePeers && 'not'} have peers.</p>
+      <p>{emojis.star} Communicating with {peers} peer{peers === 1 ? '' : 's'}.</p>
```
tomaka, line 128: In my opinion there's no point in printing "and should have peers".

```suggestion
      <p>{emojis.chain} Chain is {status}.</p>
```

However if `!shouldHavePeers`, then you should hide the line below ("Communicating with N peers").

projects/extension/src/components/NetworkTab.tsx
```diff
@@ -109,58 +109,47 @@ const useStyles = makeStyles((theme: Theme) =>
...
+      <p>{emojis.info} GrandPa sync is {isSyncing && 'not'} finished.</p>
```
tomaka, line 129: I don't think we should mention Grandpa warp syncing. Users don't care about Grandpa warp syncing, they just want to know whether everything is working correctly or not.

wirednkod: I will remove Grandpa wording then. I think user should be aware if everything is synced or not. What do you think @tomaka ?

tomaka: Yeah they need to be aware whether it's synced, but in my opinion showing either :heavy_check_mark: or a red X works as well. What I'm not a fan of is talking about "Grandpa warp sync" in particular.

projects/extension/src/background/ConnectionManager.ts
```diff
@@ -49,27 +49,26 @@ export class ConnectionManager extends (EventEmitter as { new(): StateEmitter })
    *
    * @returns a list of the networks that are currently connected
    */
-  get registeredClients(): string[] {
-    return this.#networks.map(s => s.name);
+  get registeredClients(): unknown[] {
```
raoulmillais, line 52: Why is this `unknown`?  We should know the type of this

projects/extension/src/background/ConnectionManager.ts
```diff
@@ -49,27 +49,26 @@ export class ConnectionManager extends (EventEmitter as { new(): StateEmitter })
...
   }
 
   /**
    * apps
    *
    * @returns all the connected apps.
    */
-  get apps(): App[] {
-    return this.#apps;
+  get apps(): unknown[] {
```
raoulmillais, line 61: Same as above

projects/extension/src/background/ConnectionManager.test.ts
```diff
@@ -195,7 +195,7 @@ describe('Unit tests', () => {
   test('Get networks/chains', () => {
     // With this look the "chain" is removed intentionally as "chain"
     // object cannot be compared with jest 
-    const tmpChains = manager.networks.map(n => (
+    const tmpChains = manager.registeredClients.map((n: any) => (
```
raoulmillais, line 198: This is because you're using `unknown` below.  You should be able to type this properly to avoid all these warnings

raoulmillais (commented):
There's a proliferation of using `any` across not just the tests but the implementation code.  We shouldn't be shortcutting typing things properly, the compiler's there to help us avoid bugs :smile:

projects/extension/src/background/ConnectionManager.ts
wirednkod: Forgot to check linter - CI grab all those and Im fixing them atm

packages/connect/src/SmoldotProvider/SmoldotProvider.ts
```diff
@@ -260,7 +260,7 @@ export class SmoldotProvider implements ProviderInterface {
     assert(!this.#client && !this.#isConnected, 'Client is already connected');
     try {
       this.#client = await this.#smoldot.start({
-        forbidWs: true, /* suppress console warnings about insecure connections */
```
tomaka, line 261: Woooops

packages/connect/src/SmoldotProvider/SmoldotProvider.ts
```diff
@@ -258,7 +260,7 @@ export class SmoldotProvider implements ProviderInterface {
...
+        forbidWs: false, /* suppress console warnings about insecure connections */
```
raoulmillais, line 263: Remove the comment too

raoulmillais (commented):
Not a blocker more an observation: the new types you are introducing to do with networks and apps and health have somewhat generic names which makes it more difficult to read the code.  It would be nicer to name them after what they are used for so its clearer what they do and why they exist and to use a consistent nameing scheme where we are separating out properties for internals vs UI.

* `App` vs `AppInfo` vs `ReducedApp`
* `HealthResponse` vs using `smoldot.SmoldotHealth` vs `NetworkHealth` 
* `NetworkContentProps`

projects/extension/src/types/index.ts
```diff
@@ -17,18 +17,35 @@ export type uApp = {
...
   name: string;
   icon?: string;
   status: NetworkStatus;
 }
-export interface Network extends ChainSpec {
+export interface Network extends Client {
   chain: smoldot.SmoldotChain;
   tabId: number;
   parachains?: Parachain[];
 }
```
josepot, line 29: At a first glance, I would say that it's probably best to use composition rather than having the `Network` interface extend `Client`. I.e I think this would be better:

```ts
export interface Network {
  client: Client,
  chain: smoldot.SmoldotChain;
  tabId: number;
  paracahins?: Parachain[];
}
```

Also, wouldn't it make more sense for the `parachains` prop to not be optional and use an empty array when there are no `parachains`? IMO that usually makes the API more ergonomic to work with.

projects/extension/src/background/ConnectionManager.ts
```diff
@@ -49,27 +48,26 @@ export class ConnectionManager extends (EventEmitter as { new(): StateEmitter })
    *
    * @returns a list of the networks that are currently connected
    */
-  get registeredClients(): string[] {
-    return this.#networks.map(s => s.name);
+  get registeredClients(): Client[] {
+    return this.#networks.map((s: Network) => ({name: s.name, status: s.status}));
```
josepot, line 52: The comment above says that it's returning a list of networks, should we update the comment?

Another question: Is it possible that the same client is used in more than one network? If that was the case: shouldn't this function return a unique list of clients?

projects/extension/src/types/index.ts
wirednkod: As discussed - I understand the confusion. I already altered the words to make more sense

projects/extension/src/background/ConnectionManager.ts
wirednkod: As discussed - I understand the confusion. I already altered the words to make more sense

raoulmillais (approved):

PR #530, opened 2021-10-20
Add prettier

wirednkod:
Adding prettier to substrate-connect


wirednkod:
> I'm soooooo on board with this! Thank you so much! heart bow
> 
> Just a few suggestions, though:
> 
> * ~IMO it's easier not to use the `.prettierignore` file and do this instead: `"format": "prettier --write README.md \"src/**/*.{js,jsx,ts,tsx,json,md}\""`~ on a second thought... Maybe it makes sense to have the ignore file so that we can run prettier from the root of the monorepo rather than using `yarn workspaces -s run format`? thinking
> * It would be amazing if we could integrate this with husky, I've done that in a few projects using `lint-staged` and it works very well. So, I suggest that we upgrade `husky` to v6 ([the breaking change doesn't affect us](https://github.com/typicode/husky/releases/tag/v6.0.0)) and then we add `lint-staged`
> * Last, but not least, it's important that the CI fails if the `prettier --check` doesn't pass.
> 
> Thanks again for this!

Thank you Josep.
I will upgrade husky and add the `prettier --check` in the CI, in this PR (as it is relevant to prettier).

PR #531, opened 2021-10-20
Improving loading times on substrate-connect

josepot:
I recently tested the performance of the loading times on a project that was using `substrate-connect` and the performance metrics were not looking good.

After some analysis, I was able to confirm that the main reasons were:

- Loading and parsing on the main JS chunk the JSON files for all the known chain-specs.
- Loading on the main JS chunk both providers (SmoldotProvider and ExtensionProvider)
- Loading on the main JS chunk `@polkadot/api`

Luckily for us, we can dynamically load the provider that the user needs when they call `connect`, while in a parallel request we load the `@polkadot/api`. That's what the first commit of this PR does.

However, what was penalizing the performance the most was the issue with the JSON files... I had to get a bit creative to solve that one. This is the solution that I implemented in the second commit:

- At the end of the `build` task we run a script that transforms those JSON files in minified string (on separate js assets that go directly under `dist/specs/generated`), then and only if the SmoldotProvider is needed, we dynamically import the only string that it's needed for that chain, while at the same time in parallel request we are loading the SmoldotProvider. Also, we never parse that string to a JS Object in the main thread, because there is no need for doing that.

These changes should make a significant difference when it comes to the loading times of the App that's using `substrate-connect`.

### DIFF
--- a/packages/connect/src/Detector.ts
+++ b/packages/connect/src/Detector.ts
@@ -1,15 +1,13 @@
 import type { ApiPromise } from "@polkadot/api"
 import { ApiOptions } from "@polkadot/api/types"
 import { ProviderInterface } from "@polkadot/rpc-provider/types"
-import westend from "./specs/westend.json"
-import kusama from "./specs/kusama.json"
-import polkadot from "./specs/polkadot.json"
-import rococo from "./specs/rococo.json"
+import { SupportedChains, getSpec } from "./specs/index.js"
 
 interface ChainInfo {
   name: string
   spec: string
 }
+
 /**
  * Detector is an API for providing an instance of the PolkadotJS API configured
  * to use a WASM-based light client.  It takes care of detecting whether the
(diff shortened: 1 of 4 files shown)


packages/connect/src/Detector.ts
```diff
@@ -159,31 +134,34 @@ export class Detector {
...
+          SupportedChains,
         ).join("', '")}') or provide valid chainSpecs.`,
       )
     }
 
     if (this.#isExtension) {
-      provider = new ExtensionProvider(
+      const { ExtensionProvider } = await import(
+        "./ExtensionProvider/ExtensionProvider.js"
+      )
```
tomaka, line 152: I don't think we need this one, do we? The extension provider is supposed to be very small.

packages/connect/src/Detector.ts
```diff
@@ -159,31 +134,34 @@ export class Detector {
...
-  public provider = (
+  public provider = async (
     chain: ChainInfo,
     parachainSpec?: string,
-  ): ProviderInterface => {
-    let provider: ProviderInterface = {} as ProviderInterface
-
-    if (!chain.name && !Object.keys(this.#chainSpecs).includes(chain.name)) {
+  ): Promise<ProviderInterface> => {
+    if (!chain.name && !SupportedChains[chain.name as SupportedChains]) {
```
tomaka, line 141: The API user **must** be able to use a chain other than the support ones by providing their own JSON.

wirednkod (approved):

PR #536, opened 2021-10-21
refactor: remove `chainName` from the public API

josepot:
This PR addresses a number of issues that we were having with the public API as a result of exposing an internal (the `chainName`).

In reality the `chainName` is only a means for the ExtensionProvider to determine the ownership of a message. Therefore, that's something that the consumer of `SubstrateConnect` doesn't need to know about.

Also, due to exposing this internal we were also exposing a public `disconnect` method on the instance of the `Detector`. This method received that id and then it removed the Provider from the internal cache of the Detector.

However, what happened if the consumer called the `disconnect` method on the instance returned by connect? Well, in that case the `Detector` would never know about that and that could create a memory leak. That's why it's better to only give the user one way to disconnect, which should be through the instance returned when they `connect`.

Summarizing, the changes of this PR are:

- Both `connect` and `provider` will now have 2 different TS overloads, one for when the user wants to provide their own `chainSpec` and the other in case that the user wants to use one of the predefined chainSpecs. The reason for having 2 different overloads, is that the name of this parameter is different for each overload, so that it's even more obvious to the user that this are 2 very different things. Also, the `chainName` has been removed from both `connect` and `provider`. 

- It removes the public `disconnect` method from the `Detector` and it makes sure that when the user calls `disconnect` on the instance returned by the `connect` method the instance gets removed from the cache.

- It ensures that the Smoldot can also receive the chainSpec provided by the user. So, it fixes an existing bug where Smoldot never received user defined chainSpecs.

## EDIT:

In an "offline" conversation with @wirednkod he pointed out that after these changes substrate-connect was no longer working with the Extension. It turns out that the extension was using the `chainName` for determining whether the chain was known or not. In order not to break the Extension what we will do (for now) is to always pass the full spec from the client to either provider, then we will change the Extension so that if it sees that the chainSpec is not a real chainSpec but a key of one of the supported ones, that it uses that the predefined spec for that one. Once that has been addressed in the extension, we will then be able to use the Extension without having to pass the full spec from the client. Thanks again @wirednkod for your help on this!

### DIFF
--- a/packages/connect/src/Detector.ts
+++ b/packages/connect/src/Detector.ts
@@ -79,7 +79,7 @@ export class Detector {
    * "westend" which require no further configuration.
    *
    * Alternatively you may supply a chain spec and options to connect to a
-   * custom chain with a light client (in-page only).
+   * custom chain with a light client .
    *
    * @param chainSpec - string of the chainSpec to connect to a network
    * @param parachainSpec - an optional param. This param is the parachain spec to connect to a different network)
@@ -88,9 +88,6 @@ export class Detector {
    * @returns a promise that resolves to an instance of the PolkadotJS API
    *
    * @remarks
-   *
-   * When providing a custom chain spec, detect will always configure and return
-   * an in-page light client and not use the extension
    *
    * Typically options is used to pass custom types to PolkadotJS API.
    *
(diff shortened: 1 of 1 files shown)


packages/connect/src/Detector.ts
```diff
@@ -103,29 +131,40 @@ export class Detector {
...
    * {@link https://polkadot.js.org/docs/}
    */
-  public connect = async (
-    relay: ChainInfo | string,
+  public async connect(
+    knownChain: SupportedChains,
+    parachainSpec?: string,
+    options?: ApiOptions,
+  ): Promise<ApiPromise>
+  public async connect(
```
wirednkod, line 139: This seems strange a bit (along with lines 101 - 105)

wirednkod (approved):

PR #555, opened 2021-11-23
Upgrade smoldot to 0.5.4

wirednkod:


josepot (changes requested):
I think that we should close this PR. The reason being that I've found a couple significant issues on `@substrate/smoldot-light@0.5.4` that should be addressed in there. Hopefully, we will soon have a new version with better typings and with an `async terminate` method and then we will be able to adapt the code of `substrate-connect` accordingly.

wirednkod:
Replaces with [this PR](https://github.com/paritytech/substrate-connect/pull/559)

PR #565, opened 2021-12-03
Fix routing provider for extension

wirednkod:
Following the findings of @josepot  and using the test on this [PR](https://github.com/paritytech/substrate-connect/pull/550) I tried to solve the issue about the way that the messages are being routed to their own Provider. 
After testing (auto and manual) I can see that the messages that do not belong to specific provider are now ignored and only the intended ones are delivered.

Best way to verify is by running the tests or the Multiple Network demo - on which (as mentioned before) the multiplexing never worked - but does not.

I would prefer to have a better solution of creating a unique Id than using the `Math.random() * 1000000` so please for your feedback 

cc @josepot


packages/connect/src/ScProvider.ts
```diff
@@ -90,8 +90,11 @@ export class ScProvider implements ProviderInterface {
   ) {
     const isExtension = !!document.getElementById("substrateExtension")
 
+    const id = parseInt((Math.random() * 1000000000).toString(), 0)
```
wirednkod, line 93: Im not really fun of this Math.random... If anything better come to mind please share @josepot

josepot:
> I would prefer to have a better solution of creating a unique Id than using the `Math.random() * 1000000` so please for your feedback
> 
> cc @josepot

I would rather to solve this in the way that I suggested on #550:

> If we continue using window.postMessage(message, *), then I think that a sensible approach would be for each Provider to register themselves with the Extension, by sending an initial "handshake" message to the Extension with the metadata of the Provider, and then the Extension would respond with a unique id for that Provider. That unique id would then be used from the providers in the messages that they send to the Extension, and the Extension would also include that id on the message that's sent to the Provider, which would allow for each Provider to ignore the messages that are not meant for them

wirednkod:
This Closes as will be fixed with #572

PR #569, opened 2021-12-10
Use smoldot's database system in the extension

### LINKED ISSUE #568: Use smoldot's database system in the extension
The extension should periodically call `smoldotClient.databaseContent()` for the hardcoded chains (Polkadot, Kusama, Rococo, Westend) and save the result in local storage.
When the extension is loaded, it should load the local storage and pass the content to the configuration passed to `addChain`.

Smoldot will not return any error if a bad database content is passed, so there's no need to worry about this.

wirednkod:
add repetative alarm for locally save and reuse databaseContent
Fixes #568


projects/extension/src/background/ConnectionManager.ts
```diff
@@ -215,6 +217,18 @@ export class ConnectionManager
...
+    for (const [key] of relayChains.entries()) {
+      const network = this.#networks.filter((n) => n.name === key)
+      network[0] &&
+        network[0].chain.databaseContent().then((db) => {
+          chrome.storage.local
+            .set({ [key]: db })
+            .catch((e: Error) => console.error(e))
+        })
+    }
+  }
```
josepot, line 230: There are a couple of things here that I think should be improved:
- It makes for sense to iterate `this.#networks` just once and then only grab those ones that are present in `relayChains`
- the current implementation only catches the error of `chrome.storage.local.set`, the error of the `chain.databaseContent` is being ignored, I think that it makes more sense to take into account both promises when catching.
This suggestion addresses those 2 things:

```suggestion
  runAlarm(): void {
    this.#networks
      .filter(n => relayChains.has(n.name))
      .forEach(async (network) => {
        try {
          const db = await network.chain.databaseContent()
          await chrome.storage.local.set({ [key]: db })
        } catch (e: Error) {
          console.error(e)
        }
      })
    }
```

wirednkod: Thank you for the feedback. 
A) the database content should run only for the 4 relay chains that are hard added to extension for speed purposes
B) PR is still draft - ill add the catch to the rest once I manage to resolve a strange test erroe

projects/extension/src/background/ConnectionManager.ts
```diff
@@ -252,11 +278,24 @@ export class ConnectionManager
...
+          chainSpec,
+          jsonRpcCallback,
+          databaseContent,
+          potentialRelayChains: potentialNetworks.map((r) => r.chain),
+        }
+      : {
+          chainSpec,
+          jsonRpcCallback,
+          potentialRelayChains: potentialNetworks.map((r) => r.chain),
+        }
```
josepot, line 296: This ternary is bogus

```suggestion
    const options: AddChainOptions =  {
        chainSpec,
        jsonRpcCallback,
        databaseContent,
        potentialRelayChains: potentialNetworks.map((r) => r.chain),
    }
```

projects/extension/src/background/ConnectionManager.ts
```diff
@@ -228,6 +242,18 @@ export class ConnectionManager
...
+    return new Promise((resolve, reject) => {
+      try {
+        chrome.storage.local.get(key, (value) => {
+          resolve(value[key] as string)
+        })
+      } catch (ex) {
+        reject(ex)
+      }
+    })
+  }
```
josepot, line 255: A few things here:
- It should return `Promise<string | undefined>`
- The name is a bit confusing and leaky, what about `getDbContent` instead?
- The `async` is bogus.
- According to [these docs](https://developer.chrome.com/docs/extensions/reference/storage/#usage) we should be doing `get([key], ...` rather than ``get(key, ...`

```suggestion
getDbContent = (key: string): Promise<string | undefined> =>
  new Promise((resolve, reject) => {                        
    try {                                                   
      chrome.storage.local.get([key], (value) => {          
        resolve(value[key])                                 
      })                                                    
    } catch (ex) {                                          
      reject(ex)                                            
    }                                                       
  })                                                        

```

projects/extension/src/background/ConnectionManager.ts
josepot: > A) the database content should run only for the 4 relay chains that are hard added to extension for speed purposes

Which is wha the suggsted code is doing in a more efficient way than the previous one.

projects/extension/src/background/ConnectionManager.ts
```diff
@@ -252,11 +278,18 @@ export class ConnectionManager
       ? this.#networks.filter((n) => n.tabId === tabId)
       : this.#networks
 
-    const addedChain = await this.#client.addChain({
+    const databaseContent = await this.getObjectFromLocalStorage(
+      (name as string).toLowerCase(),
+    )
```
tomaka, line 283: This must only be done for the hardcoded chains: Westend, Polkadot, Kusama, Rococo

projects/extension/src/background/ConnectionManager.ts
```diff
@@ -215,6 +218,17 @@ export class ConnectionManager
     this.#client = undefined
   }
 
+  runAlarm(): void {
+    this.#networks
+      .filter((n) => relayChains.has(n.name))
```
tomaka, line 223: Please never use names or anything that comes from the chain spec as a way to identify a chain. Names are completely untrusted input. Please use only the value that smoldot returns when calling `addChain` to know which chains to load the database from.

I think the entire  ̀ConnectionManager` needs a rework.

tomaka (changes requested):

projects/extension/src/background/ConnectionManager.ts
wirednkod: I agree with this comment - but I would avoid reworking the whole `ConnectionManager` in this PR. 
If it is fine for you - we can use the name for identification at the moment - and fix the whole thing in a following PR. What do you think?

tomaka: I think that reworking the `ConnectionManager` is super top notch priority, and is way more important than this database thing

wirednkod: > I think that reworking the `ConnectionManager` is super top notch priority, and is way more important than this database thing

#570

PR #570, opened 2021-12-12
Connection manager rework

wirednkod:
This PR is a rework on the connection manager part of the extension. Major changes have:
- Adjust the Mapping of known chains (polkadot, kusama, rococo, westend) to use Ids (e.g. ksmcc3) instead of names (e.g. kusama);
- Split flow when extension inits for default chains from the external chainspecs (makes flow faster and removes non-needed steps);
- Move all actions from the background `index.js` to `ConnectionManager` class.
- Add fixes concerning linter (TS-y a bit more the code) in order to remove the disable comments from the file
- Replace `substr` deprecated javascript function

### DIFF
--- a/projects/extension/src/background/ConnectionManager.ts
+++ b/projects/extension/src/background/ConnectionManager.ts
@@ -49,6 +49,29 @@ export class ConnectionManager
   #networks: Network[] = []
   smoldotLogLevel = 3
 
+  constructor() {
+    super()
+    this.initSmoldot()
+    const promises = []
+    for (const [key, value] of relayChains.entries()) {
+      const jsonRpcCallback = (rpc: string) => {
+        console.warn(`Got RPC from ${key} dummy chain: ${rpc}`)
+      }
+      const promise = this.#client
+        ?.addChain({
+          chainSpec: value,
+          jsonRpcCallback,
+        })
+        .catch((err) => {
+          console.error("Error", err)
+        })
+      promises.push(promise)
+    }
+    Promise.all(promises).catch((err) =>
+      console.log("Error creating smoldot:", err),
+    )
+  }
+
   /** registeredApps
    *
    * @returns a list of the names of apps that are currently connected
@@ -230,32 +253,6 @@ export class ConnectionManager
   async shutdown(): Promise<void> {
     await this.#client?.terminate()
     this.#client = undefined
-  }
-
-  /**
-   * init Runs on onStartup and onInstalled of extension and
-   * connects the default relayChains to smoldot
-   */
-  init = () => {
-    this.initSmoldot()
-    const promises = []
-    for (const [key, value] of relayChains.entries()) {
-      const jsonRpcCallback = (rpc: string) => {
-        console.warn(`Got RPC from ${key} dummy chain: ${rpc}`)
-      }
-      const promise = this.#client
-        ?.addChain({
-          chainSpec: value,
-          jsonRpcCallback,
-        })
-        .catch((err) => {
-          console.error("Error", err)
-        })
-      promises.push(promise)
-    }
-    Promise.all(promises)
-      .then(() => console.log("all run"))
-      .catch((err) => console.log("Error creating smoldot:", err))
   }
 
   /**
--- a/projects/extension/src/background/ConnectionManager.ts
+++ b/projects/extension/src/background/ConnectionManager.ts
@@ -1,8 +1,4 @@
 /* eslint-disable @typescript-eslint/unbound-method */
-/* eslint-disable @typescript-eslint/no-unsafe-assignment */
-/* eslint-disable @typescript-eslint/no-unsafe-call */
-/* eslint-disable @typescript-eslint/no-unsafe-member-access */
-/* eslint-disable @typescript-eslint/no-explicit-any */
 import {
   Client,
   start,
@@ -34,6 +30,11 @@ const relayChains: RelayType = new Map<string, string>([
 
 const l = logger("Extension Connection Manager")
 
+interface SpecInterface {
+  name: string
+  id: string
+  relay_chain: string
+}
 /**
  * ConnectionManager is the main class involved in managing connections from
  * apps.  It keeps track of apps and it is also responsible for triggering
@@ -130,6 +131,9 @@ export class ConnectionManager
     this.#apps.filter((a) => a).forEach((a) => this.disconnect(a))
   }
 
+  /**
+   * createApp maps the needed info and creates a new app
+   */
   createApp(incPort: chrome.runtime.Port): App {
     const splitIdx = incPort.name.indexOf("::")
     if (splitIdx === -1) {
@@ -140,8 +144,8 @@ export class ConnectionManager
       throw new Error(payload)
     }
     const { name, sender } = incPort
-    const appName: string = name.substr(0, splitIdx)
-    const chainName: string = name.substr(splitIdx + 2, name.length)
+    const appName: string = name.slice(0, splitIdx)
+    const chainName: string = name.slice(splitIdx + 2, name.length)
     const tabId: number = sender?.tab?.id || -1
     const url: string | undefined = sender?.url
     const port: chrome.runtime.Port = incPort
@@ -265,7 +269,9 @@ export class ConnectionManager
   }
 
   /**
-   * addChain adds the Chain in the smoldot client
+   * addChain: adds the Chain in the smoldot client. This is called only on new chains
+   * added and not on default pnes - as the default ones do not need the extra checks
+   * that take place here or be added to the Networks.
    *
    * @param spec - ChainSpec of chain to be added
    * @param jsonRpcCallback - The jsonRpcCallback function that should be triggered
@@ -281,7 +287,9 @@ export class ConnectionManager
     if (!this.#client) {
       throw new Error("Smoldot client does not exist.")
     }
-    const { name, id, relay_chain } = JSON.parse(chainSpec)
+    const { name, id, relay_chain }: SpecInterface = JSON.parse(
+      chainSpec,
+    ) as SpecInterface
 
     // identify all relay_chains init'ed from same app with tabId identifier
     const potentialNetworks: Network[] = relay_chain
@@ -311,11 +319,10 @@ export class ConnectionManager
   }
 
   #initHealthChecker = (app: App, isParachain?: boolean): void => {
-    if (isParachain) {
-      app.parachain &&
-        app.healthChecker?.setSendJsonRpc(app.parachain.sendJsonRpc)
-    } else {
-      app.chain && app.healthChecker?.setSendJsonRpc(app.chain.sendJsonRpc)
+    if (isParachain && app.parachain) {
+      app.healthChecker?.setSendJsonRpc(app.parachain.sendJsonRpc)
+    } else if (app.chain) {
+      app.healthChecker?.setSendJsonRpc(app.chain.sendJsonRpc)
     }
     void app.healthChecker?.start((health: SmoldotHealth) => {
       if (
@@ -336,8 +343,9 @@ export class ConnectionManager
 
   #handleError = (app: App, e: Error): void => {
     const error: MessageFromManager = { type: "error", payload: e.message }
-    app.port.postMessage(error)
-    app.port.disconnect()
+    const { port } = app
+    port.postMessage(error)
+    port.disconnect()
     this.unregisterApp(app)
   }
 
@@ -371,7 +379,7 @@ export class ConnectionManager
         })
         .then((network) => {
           app.parachain = network.chain
-          app.chainName = JSON.parse(parachainSpec).name
+          app.chainName = (JSON.parse(parachainSpec) as SpecInterface).name
           return
         })
     } else {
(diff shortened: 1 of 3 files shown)


josepot:
> This PR is a rework on the connection manager part of the extension. Major changes have:
> 
> * Adjust the Mapping of known chains (polkadot, kusama, rococo, westend) to use Ids (e.g. ksmcc3) instead of names (e.g. kusama);

Why, though? Also, wouldn't it be better to lift the `SupportedChains` enum that's currently used on `@substrate/connect` to the common `@substrate/connect-extension-protocol` package?

> * Split flow when extension inits for default chains from the external chainspecs (makes flow faster and removes non-needed steps);

Could you please elaborate on this? What's the change that is making "the flow faster"? Also, wouldn't it be better to use `Promise.all` instead of sequentially iterating over those `addChain` Promises?

> * Move all actions from the background `index.js` to `ConnectionManager` class.

Interesting... why, though?

> * Add fixes concerning linter (TS-y a bit more the code) in order to remove the disable comments from the file

Cool.

> * Replace `substr` deprecated javascript function

Perhaps a slightly better way to do it:
```ts
const [appName, chainName] = incPort.name.split('::')
if (!appName || !chainName) {
  // throw
}
```
still not perfect, though.

I'm not saying that these changes are wrong. However, I fail to see how these changes address the [concerns raised on #569](https://github.com/paritytech/substrate-connect/pull/569#discussion_r767139514) . Would you mind elaborating on that? :pray:

wirednkod:
> Why, though? Also, wouldn't it be better to lift the `SupportedChains` enum that's currently used on `@substrate/connect` to the common `@substrate/connect-extension-protocol` package?

It is better to use the ids rather than names of the chains (as per Pierre's comment as well). IDs are more "unique" and change per version - e.g. rococo_v_12 -> rococo_v_13). 

> Could you please elaborate on this? What's the change that is making "the flow faster"? 

The `addChain` in ConnectionManager should be used for incoming chainspecs as it parses the specs to retrieve the name, id, relay_chain (needed for extension UI), identifies all relay_chains that are initiated from same app with tabId identifier, and once smoldot's addChain is called, checks if this network exists and if not adds it to the array of `this.#networks`.
During the initialisation of the extension we do not need all these steps - all we need is to initSmoldot and once done start the pre-defined chains. 

> Also, wouldn't it be better to use `Promise.all` instead of sequentially iterating over those `addChain` Promises?

Yes it would! Thank you 🙏🏽 Will do that. Thank you!

> Interesting... why, though?

Mainly due to the fact that the [relaychains](https://github.com/paritytech/substrate-connect/pull/570/files#diff-05bb44f78f8c40ea612398108dccd416ffc760bff6f63cfcd50e83b320948de3R24) are needed (or will be) from `ConnectionManager` (e.g. in the other [PR](https://github.com/paritytech/substrate-connect/pull/569/files#diff-05bb44f78f8c40ea612398108dccd416ffc760bff6f63cfcd50e83b320948de3R23) that is still pending) and I prefer to avoid exporting this kind of info. 

> I'm not saying that these changes are wrong. However, I fail to see how these changes address the [concerns raised on #569](https://github.com/paritytech/substrate-connect/pull/569#discussion_r767139514) . Would you mind elaborating on that? 🙏

I thought is a good chance to face-lift and organise a bit the connection manager based on Pierre's input - though Im not sure what could change [on that part of the code](https://github.com/paritytech/substrate-connect/pull/569/files#diff-05bb44f78f8c40ea612398108dccd416ffc760bff6f63cfcd50e83b320948de3R222). I mean - cannot really understand how I would identify from the value returned which chain is one of the default ones.

josepot:
> > Why, though? Also, wouldn't it be better to lift the `SupportedChains` enum that's currently used on `@substrate/connect` to the common `@substrate/connect-extension-protocol` package?
> 
> It is better to use the ids rather than names of the chains (as per Pierre's comment as well). IDs are more "unique" and change per version - e.g. rococo_v_12 -> rococo_v_13). Just to clarify here - IDs (or names before) are used for identifying in extension which is which - and not for communicating to smoldot.

I don't get it. Also, @tomaka comment's clearly states that :

> Please never use names or anything that comes from the chain spec as a way to identify a chain. Names are completely untrusted input. Please use only the value that smoldot returns when calling addChain to know which chains to load the database from.

I'm guessing that the reason for using those ids, is so that later the `id` (rather than the `name`) from the chainspec can be used for determining if a given network is using one of the "supported chains". If that's the case, then I think that this is not the right approach.

The thing is that from the `ConnectionManger` `addChain` method we could (and should) know if the chain that it's added is a "supported chain" (without having to introspect the chainspec). If we were able to have that knowledge in the moment of adding the chain, then we could use a new property inside the `Network` interface to determine whether the underlying connection is a "supported chain" (or if we don't want to add a new property inside the `Network` interface we could use a `WeakSet`, whatever).

Now the question is: how can know in the moment when `addChain` is called if the passed chain is one of the "supported chains"?

AFAIK `addChain` is only called from 2 different places:

1) from the `init` function (wherever we decide where to locate it, I don't care), in which case all those calls are for known chains, so this one is easy.

2) from the `handleSpecMessage` callback. I've pointed out (many times) in the past that (for "known specs") it makes no sense to pass the whole "known spec" from the tab to the Extension as that has some performance implications, but most importantly because that spec is already loaded on the Extension. So, if we used a message type like `knownSpec`, rather than using the generic message type `spec`, then we could know from the Extension that the chain that's about to get added is a known one. Therefore, there wouldn't be any need to introspect the chainspec.

Also, as I already pointed out in a previous comment: it would make sense to "lift" the `SupportedChains` enum to the common `@substrate/connect-extension-protocol` package so that both the Extension and the connect package can use the same source of truth.

IMO this is the kind of rework that's most urgent on the `ConnectionManager`.

projects/extension/src/background/ConnectionManager.ts
```diff
@@ -215,13 +232,38 @@ export class ConnectionManager
     this.#client = undefined
   }
 
+  /**
+   * init Runs on onStartup and onInstalled of extension and
+   * connects the default relayChains to smoldot
```
tomaka, line 237: You shouldn't document when a function is called, that's the problem of the user of the code. `init` is here to initialize the state machine, and that's it. That's all there is to document.

projects/extension/src/background/ConnectionManager.ts
```diff
@@ -215,13 +232,38 @@ export class ConnectionManager
...
+   */
+  init = () => {
```
tomaka, line 239: Also, IMO we should be calling `init` from the constructor. There's no reason it has to be a different function. If you don't want to initialize stuff, just don't create the `ConnectionManager`.

This would also make it possible to make the types of the various fields more precise. For example `#client` could be just of type `Client` instead of `Client | undefined`.

tomaka:
> Also, as I already pointed out in a previous comment: it would make sense to "lift" the SupportedChains enum to the common @substrate/connect-extension-protocol package so that both the Extension and the connect package can use the same source of truth.

Doing this would prevent us forever from removing support for a specific chain.
I think it's a better idea to just pass the id of the supported chain by string, and the extension returns back an error if this id isn't known to it. In that case, the web page can load the chain spec and pass them to the extension.

projects/extension/src/background/index.ts
```diff
@@ -15,43 +10,12 @@ declare let window: Background
...
 const manager = (window.manager = new ConnectionManager())
 
-type RelayType = Map<string, string>
-
-export const relayChains: RelayType = new Map<string, string>([
-  ["polkadot", JSON.stringify(polkadot)],
-  ["kusama", JSON.stringify(kusama)],
-  ["rococo", JSON.stringify(rococo)],
-  ["westend", JSON.stringify(westend)],
-])
```
tomaka, line 25: IMO this should be left in the `index.ts`, and instead you add a new `trusted: bool` parameter to `addChain` indicating whether the chain is trusted.

projects/extension/src/background/ConnectionManager.ts
```diff
@@ -246,7 +290,9 @@ export class ConnectionManager
     if (!this.#client) {
       throw new Error("Smoldot client does not exist.")
     }
-    const { name, id, relay_chain } = JSON.parse(chainSpec)
+    const { name, id, relay_chain }: SpecInterface = JSON.parse(
+      chainSpec,
+    ) as SpecInterface
```
tomaka, line 295: This `SpecInterface` thing seems wrong. You have no guarantee that a chain spec passed by the user actually has these fields. Which is also a good reason why shouldn't use the `name` anywhere. I don't know how TypeScript behaves if a field is missing, but if it throws an exception then it's definitely wrong.

wirednkod:
> The thing is that from the `ConnectionManger` `addChain` method we could (and should) know if the chain that it's added is a "supported chain" (without having to introspect the chainspec). If we were able to have that knowledge in the moment of adding the chain, then we could use a new property inside the `Network` interface to determine whether the underlying connection is a "supported chain" (or if we don't want to add a new property inside the `Network` interface we could use a `WeakSet`, whatever).

In substrate connect we used to [pass the "name" of a chainspec](https://github.com/paritytech/substrate-connect/blob/4c2135504e721b7621250ef7bb05aac4b392d98c/projects/smoldot-browser-demo/src/index.ts#L17) (e.g. "westend") in the detector in order to identify a supported chain. We removed that as it was preferable to request just the chainspecs (common API for all) from the dev (plus any name could be added at that point)

> 2. from the `handleSpecMessage` callback. I've pointed out (many times) in the past that for "known specs" it makes no sense to pass the whole "known spec" from the tab to the Extension as that has some performance implications, but most importantly because that spec is already loaded on the Extension. So, if we used a message type like `knownSpec` and in the payload we put one of the `SupportedChains` values (rather than using the generic message type `spec` with the whole spec on the payload) then we could know from the Extension that the chain that's about to get added is a known one. Therefore, there wouldn't be any need to introspect the chainspec.

If understand correct - you mean to parse the spec on ExtensionProvider level - understand that this is a known spec (cross check with the `SupportedChains` and if so - instead of passing the spec to the extension, to just inform that this is the "A" known spec - and extension will take off from there? Is that correct? Can you elaborate on that?

The problem that I see here is - if someone uses different chainspec than the one included in the extension (e.g. uses westend but altered the bootnodes - I think this is a valid story)

projects/extension/src/background/ConnectionManager.ts
wirednkod: `init` needs to be called in 2 different parts of the lifecycle of the extension. 
1 (existing) - on installation of the extension
2 (existing) - on startup of the browser
ofc the `init` can be called from the constructor instead of being a different function - though that will limit our control of the function based on the lifecycle (there is also the `onUpdateAvailable`, `onSuspend` and `onSuspendCanceled` that may be useful to have the init as a separate function. 

For now I will move this to the constructor - just mentioning the reasoning behind the init()

projects/extension/src/background/index.ts
wirednkod: Can you please elaborate on the reasoning of having a `trusted: bool`. The idea behind this `Map` is that all of the chains are trusted. Maybe the `relayChains` naming should alter to `trustedChains` to be more obvious

projects/extension/src/background/ConnectionManager.ts
wirednkod: This is wrong - did not see it - thank you for pointing out.
The SpecInterface should be altered to having optional params (e.g. name and relay_chain) but id should always be there

projects/extension/src/background/index.ts
tomaka: > In addition there is no reason for this to be in the index.ts

The reason is to leave the `ConnectionManager` as simple as possible. Adding these chains there just adds complexity to the code.

> Can you please elaborate on the reasoning of having a trusted: bool. The idea behind this Map is that all of the chains are trusted. Maybe the relayChains naming should alter to trustedChains to be more obvious.

This `trusted` parameter is way to move back the `relayChains` back to `index.ts`. If you leave `relayChains` in `ConnectionManager.ts` then indeed you don't need this parameter.

projects/extension/src/background/ConnectionManager.ts
tomaka: If you want to "de-init" a `ConnectionManager` then you should instead use a variable whose type is `ConnectionManager | null` and switch it to `null`. The inside of the `ConnectionManager` shouldn't support a "not initialized" mode.

josepot:
> > The thing is that from the `ConnectionManger` `addChain` method we could (and should) know if the chain that it's added is a "supported chain" (without having to introspect the chainspec). If we were able to have that knowledge in the moment of adding the chain, then we could use a new property inside the `Network` interface to determine whether the underlying connection is a "supported chain" (or if we don't want to add a new property inside the `Network` interface we could use a `WeakSet`, whatever).
> 
> In substrate connect we used to [pass the "name" of a chainspec](https://github.com/paritytech/substrate-connect/blob/4c2135504e721b7621250ef7bb05aac4b392d98c/projects/smoldot-browser-demo/src/index.ts#L17) (e.g. "westend") in the detector in order to identify a supported chain. We removed that as it was preferable to request just the chainspecs (common API for all) from the dev (plus any name could be added at that point)

We have not removed the ability to pass a "known chain" to the Provider, that option still exists... We improved that by having 2 different overloads and an Enum, to help the user not mistake one API with the other. So, we have the ability to let the Extension know that the user is interested in a trusted chain.

> > 2. from the `handleSpecMessage` callback. I've pointed out (many times) in the past that for "known specs" it makes no sense to pass the whole "known spec" from the tab to the Extension as that has some performance implications, but most importantly because that spec is already loaded on the Extension. So, if we used a message type like `knownSpec` and in the payload we put one of the `SupportedChains` values (rather than using the generic message type `spec` with the whole spec on the payload) then we could know from the Extension that the chain that's about to get added is a known one. Therefore, there wouldn't be any need to introspect the chainspec.
> 
> If understand correct - you mean to parse the spec on ExtensionProvider level - understand that this is a known spec (cross check with the `SupportedChains` and if so - instead of passing the spec to the extension, to just inform that this is the "A" known spec - and extension will take off from there? Is that correct? Can you elaborate on that?

No, that is not at all what I meant. What I'm saying is that currently the Provider (which run on the tab) when the user says that they want to use a known chain, what it does is to load that known chain and send a message to the Extension with the spec on the payload. However, what I think that it should do is to send a different kind of message like "trustedChain" with the name of the chain on the payload, because that spec has already been parsed on the extension, doing so would allow us to have a "trusted" flag on the `addChain` method of the Extension.

wirednkod:
This Closes as will be fixed with #572

Issue #572, opened 2021-12-13
Refactor the `connection-extension-protocol` package


tomaka:
This package needs some love. Here's the list of changes that I think we should do (in order):

- [x] Merge `CommonProviderMessageData` and `ProviderMessageData` into one interface. I suspect that the reason why they're two separate thing is to make the code in `ExtensionProvider` easier, but that's really not the problem of `connection-extension-protocol`.

- Right now the protocol mentions the web page, the content script, and the extension as layers of an onion. This is technically correct, but this is _not_ what the `connection-extension-protocol` package is about. This package is only about the communication between the web page and the content script. What happens between the content script and the extension is an implementation detail of the extension and shouldn't be mentioned in `connection-extension-protocol` because it's out of scope.
    - [x] Consequently, we should merge the `ProviderMessageData` and `MessageToManager` interfaces into one single interface, and `ExtensionMessageData` and `MessageFromManager` into one single interface.
    - [x] Afterwards, we can rename them to something more simple such as `ToExtension` and `ToWebpage` or something like that. The words `Provider` and `Manager` shouldn't be used because they're implementation details.
    - [x] Also remove the `ExtensionMessage` and `ProviderMessage` interfaces, again to simplify. The actual type that the callback passed to `addEventListener` receives is [`MessageEvent`](https://developer.mozilla.org/en-US/docs/Web/API/MessageEvent) (which as a field named `data`), and not `ExtensionMessage` or `ProviderMessage`. You can't actually know in advance what the `data` field contains in this context.

- [x] Remove the `extension` and `provider` variables. It doesn't make sense to use the variable `provider` when in the extension and vice versa, so why are they available? Again, sending and listening is out of scope of the `connection-extension-protocol` package which should only provide the protocol.

- [x] Remove the `ExtensionMessageData.disconnect` field, because it's exactly the same thing as `MessageFromManager.type` being equal to `error`.

- Refactor the `MessageToManager` to be similar to what [this gist](https://gist.github.com/tomaka/ec113966d9b322ead50d787b16529796) explains.
    - [x] Remove the `appName` field. Instead the extension should simply use the tab URL. We can always revisit later the decision of removing app names, but for now let's get rid of this field in order to unlock progress.
    - [x] Remove from the `ExtensionProvider` the code that was sending a message with the `connect` action. The extension can do what it was doing when receiving a `connect` event at initialization and/or the first time it receives an actual message.
[...]

PR #575, opened 2021-12-13
chore: merge CommonProviderMessageData and ProviderMessageData

josepot:
It addresses the first bullet point of #572

> Merge CommonProviderMessageData and ProviderMessageData into one interface. I suspect that the reason why they're two separate thing is to make the code in ExtensionProvider easier, but that's really not the problem of connection-extension-protocol

I also took that chance to remove the public properties `name` and `chainSpec` from the ExtensionProvider, since those properties shouldn't be exposed... I could (and perhaps should) do that in a separate PR, though.

### DIFF
--- a/packages/connect-extension-protocol/package.json
+++ b/packages/connect-extension-protocol/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@substrate/connect-extension-protocol",
-  "version": "0.3.0",
+  "version": "0.4.0",
   "description": "Protocol for connect message passing with the extension",
   "main": "dist/index.js",
   "type": "module",
--- a/packages/connect/src/ExtensionProvider/ExtensionProvider.ts
+++ b/packages/connect/src/ExtensionProvider/ExtensionProvider.ts
@@ -16,7 +16,6 @@ import { HealthCheckError } from "../errors.js"
 import {
   MessageFromManager,
   ProviderMessageData,
-  CommonProviderMessageData,
   ExtensionMessage,
   ExtensionMessageData,
   provider,
@@ -79,11 +78,12 @@ export class ExtensionProvider implements ProviderInterface {
   #connectionStatePingerId: ReturnType<typeof setInterval> | null
   #isConnected = false
 
-  #chainId: number
-  #appName: string
-  #chainName: string
   #chainSpecs: string
   #parachainSpecs: string
+  #commonMessageData: Pick<
+    ProviderMessageData,
+    "appName" | "chainId" | "chainName" | "origin"
+  >
 
   /*
    * How frequently to see if we have any peers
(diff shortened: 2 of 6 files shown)


packages/connect-extension-protocol/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@substrate/connect-extension-protocol",
-  "version": "0.3.0",
+  "version": "0.4.0",
```
tomaka, line 3: Should we not bump before releasing instead? It doesn't matter, but I'm personally used to bumping before releasing

tomaka (approved):

packages/connect/src/ExtensionProvider/ExtensionProvider.ts
```diff
@@ -95,44 +95,24 @@ export class ExtensionProvider implements ProviderInterface {
...
-   * @returns the name of the chain this `ExtensionProvider` is talking to.
-   */
-  public get chainSpecs(): string {
-    return this.#chainSpecs
+    this.#commonMessageData = {
+      appName: displayName,
+      chainId: nextChainId++,
+      chainName: JSON.parse(relayChain).name,
+      origin: EXTENSION_PROVIDER_ORIGIN,
+    }
```
wirednkod, line 115: name may not exist on the spec. This could do:
```suggestion
    const { name, id } = JSON.parse(relayChain)
    this.#commonMessageData = {
      appName: displayName,
      chainId: nextChainId++,
      chainName: name || id
      origin: EXTENSION_PROVIDER_ORIGIN,
    }
```

josepot: In a future PR I will get rid of that `chainName`, as it shouldn't be passed at all in these messages... So, for now I don't want to bother improving something that we are about to get rid off.

tomaka: This PR doesn't change anything here, it is just copy-pasting the existing code. Fixing this is out of scope of this PR here, and will be fixed later in #572.

packages/connect-extension-protocol/package.json
josepot: The reason is that `@substrate/connect-extension-protocol` is used from `@substrate/connect` and `@substrate/extension`, and I have to adapt them to the new API... It really feels wrong to declare in the `package.json` of `@substrate/connect` and `@substrate/extension` that their code depends on version `0.3.0` when the code is using an API that it's not compatible with that version... It's also a bit dangerous, because if when releasing the next versions of `@substrate/connect`/`@substrate/extension` we forgot to first release the  `@substrate/connect-extension-protocol`, then our build for those packages would work in this repo, we would be able to publish them, but the packages would be broken for the users as version `0.3.0` is not compatible with the new code...

PR #580, opened 2021-12-14
chore: rework connection-extension-protocol

josepot:
It addresses the second bullet point of #572:

> * Right now the protocol mentions the web page, the content script, and the extension as layers of an onion. This is technically correct, but this is _not_ what the `connection-extension-protocol` package is about. This package is only about the communication between the web page and the content script. What happens between the content script and the extension is an implementation detail of the extension and shouldn't be mentioned in `connection-extension-protocol` because it's out of scope.
>   
>   * [ ]  Consequently, we should merge the `ProviderMessageData` and `MessageToManager` interfaces into one single interface, and `ExtensionMessageData` and `MessageFromManager` into one single interface.
>   * [ ]  Afterwards, we can rename them to something more simple such as `ToExtension` and `ToWebpage` or something like that. The words `Provider` and `Manager` shouldn't be used because they're implementation details.
>   * [ ]  Also remove the `ExtensionMessage` and `ProviderMessage` interfaces, again to simplify. The actual type that the callback passed to `addEventListener` receives is [`MessageEvent`](https://developer.mozilla.org/en-US/docs/Web/API/MessageEvent) (which as a field named `data`), and not `ExtensionMessage` or `ProviderMessage`. You can't actually know in advance what the `data` field contains in this context.

with one extra goodie that also solves #550 and probably the security concerns that we had with 3rd party scripts running on iframes.

### DIFF
--- a/packages/connect-extension-protocol/src/index.ts
+++ b/packages/connect-extension-protocol/src/index.ts
@@ -1,143 +1,98 @@
-/**
- * @packageDocumentation
- *
- * In order to understand the protocol you should realise there are actually
- * 2 hops that happen in communication because of the architecture of browser
- * extensions.  The app has to `window.postMessage` messages to the content
- * script that gets injected by the extension. It is the content script that
- * has access to the extension APIs to be able to post messages to the
- * extension background.
- *
- * You can think of the protocol types like layers of an onion. The innermost
- * layer is the original JSON RPC request/responses. Then we wrap extra layers
- * (types) for the other 2 hops which then get peeled off at each hop. The
- * {@link MessageToManager} / {@link MessageFromManager} representing the
- * extension communication content script \<\> background. Then the outermost
- * {@link ExtensionMessage} / {@link ProviderMessage} representing the
- * communication between the PolkadotJS provider in the app and the content
- * script.
- *
- * The {@link ExtensionProvider} is the class in the app.
- * The {@link ExtensionMessageRouter} is the class in the content script.
- * The {@link ConnectionManager} is the class in the extension background.
- */
+type EventHandler<T> = (message: MessageEvent<T>) => void
 
-/**
- * `MessageFromManager` represents messages from the manager in the extension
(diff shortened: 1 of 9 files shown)


tomaka:
This does way more than these three bullet points. The messages have been completed refactored, many fields have disappeared, and the bottom half of #752 would no longer be relevant.

packages/connect-extension-protocol/src/index.ts
```diff
@@ -1,143 +1,98 @@
...
-  message?: MessageToManager
+export enum ToExtensionMessageType {
+  Connect,
+  Disconnect,
+  Spec,
+  Rpc,
+}
+
+interface ToExtensionConnect {
+  type: ToExtensionMessageType.Connect
```
tomaka, line 57: What is this compiled to? A string containing "Connect"?

josepot: In the way that the enum has been defined it currently [compiles to a number](https://www.typescriptlang.org/docs/handbook/enums.html), we could make it compile to a string... but why does it matter? The type system won't allow using the underlying compiled value, so I don't see why that's important.

tomaka: It matters because the communication is ultimately done in JavaScript. If the webpage <-> extension protocol relies on the implementation details of TypeScript then you couldn't write code that talks to the extension without using TypeScript yourself.
I would strongly prefer strings here.

tomaka: Changing the order of the elements in the enum, or removing some elements of the enum, would be a breaking change to the protocol.

josepot: > It matters because the communication is ultimately done in JavaScript. If the webpage <-> extension protocol relies on the implementation details of TypeScript then you couldn't write code that talks to the extension without using TypeScript yourself.

That's a good point, I will change that.

josepot: done! Thanks for pointing that out!

PR #595, opened 2021-12-25
Prepare extension for release

wirednkod:
- Minor changes for passing error/warnings on browser stores (chrome/firefox)
- Update smoldot-light to 0.5.10 in order to reduce size for approval on Firefox store


tomaka (approved):
I suppose that the thing already published is the version after this PR. If so, this should have been approved and merged before the actual publication.

PR #597, opened 2022-01-03
Update landing page with store's links

wirednkod:
Update landing page to include links for extension for both chrome and firefox stores


tomaka (approved):
The changes to the yarn.lock shouldn't be there. I opened https://github.com/paritytech/substrate-connect/pull/598 to prevent that in the future.

PR #599, opened 2022-01-03
Clean up CI workflow file

tomaka:
Code that we don't want should always be removed, not just commented out.


wirednkod:
This code we do want - just not activated yet

tomaka:
You can always find it here: https://github.com/paritytech/substrate-connect/blob/45fffccb583c9aa4e3f45d57cdb6e71694e91e4b/.github/workflows/yarn.CI.yml#L31-L48
This link points to a specific commit and will thus never 404.

wirednkod (approved):

tomaka:
(force-pushed to sign commit)

wirednkod (approved):

PR #601, opened 2022-01-03
Remove the "engines" requirement from package.json

tomaka:
I don't really see why it's there?

The `>=14` seems completely arbitrary (do we use a feature that was added in Node 14?), and `<=16.13.0` seems completely arbitrary as well considering that `v16.13.1` and `v17` exist.

### DIFF
--- a/packages/connect/package.json
+++ b/packages/connect/package.json
@@ -8,9 +8,6 @@
   "publishConfig": {
     "access": "public"
   },
-  "engines": {
-    "node": ">=14 <=16.13.0"
-  },
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "repository": {


wirednkod (changes requested):
I would not do that. 
There is a webpack error returned with versions > 16.13.0 (and ofc 17) that gives a `RangeError: Maximum call stack size exceeded` (e.g. extension)
Its on my todo list to identify the source of the issue and fix it before removing the <=16.13.

In addition I have not tested this in versions < 14

tomaka:
> There is a webpack error returned with versions > 16.13.0 (and ofc 17) that gives a RangeError: Maximum call stack size exceeded (e.g. extension)

But that's not our problem. Our job is not to handle WebPack bugs, it's just to indicate which version of node the `connect` package works.

wirednkod:
> > There is a webpack error returned with versions > 16.13.0 (and ofc 17) that gives a RangeError: Maximum call stack size exceeded (e.g. extension)
> 
> But that's not our problem. Our job is not to handle WebPack bugs, it's just to indicate which version of node the `connect` package works.

I think it is a good idea to "protect" (quotes are in purpose) devs who uses substrate connect this way. Though if you believe that its no need to do so - please go on and remove that

tomaka:
(force-pushed to sign my commit)

tomaka:
We don't want to be tied to Webpack. People should be able to use substrate-connect with just NodeJS or with another bundler. If the bug indeed has to do with our code, then we could add a note in the README. But denying anyone from using substrate-connect with NodeJS 17 just because of a WebPack is a bit exaggerated.

wirednkod (approved):

PR #602, opened 2022-01-03
Remove node_esm script from bin

tomaka:
Reasons to remove are:

- It's really not worth have a separate script just for passing two flags to `node`.
- There's fundamentally no reason why the `connect` package and the `extension` package would always use the same flags. There's no "DRY" argument for this script, because you _should_ repeat yourself here.
- It adds an indirection when reading the `package.json`. Instead of knowing what `npm test` does, you have to additionally open another file.

### DIFF
--- a/bin/node-esm
+++ b/bin/node-esm
@@ -1,3 +1,0 @@
-#!/usr/bin/env bash
-
-node --no-warnings --experimental-vm-modules "$@"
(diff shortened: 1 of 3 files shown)


wirednkod (approved):

wirednkod (dismissed):

tomaka:
(sorry, force-pushed again to sign my commit)

wirednkod (approved):

PR #603, opened 2022-01-03
Remove specs download and check scripts

tomaka:
There's no reason to download and check specs from the smoldot repo.

Smoldot isn't a better place to hold chain specs than substrate-connect.
We want the chain specs that are part of the extension to be extremely "deterministic".

Furthermore, this script makes tests fail if I ever update the chain specs in the smoldot repo. Anything that can automatically make tests fail is a bad thing.


wirednkod (changes requested):
> There's no reason to download and check specs from the smoldot repo.
> 
> Smoldot isn't a better place to hold chain specs than substrate-connect. We want the chain specs that are part of the extension to be extremely "deterministic".
> 
> Furthermore, this script makes tests fail if I ever update the chain specs in the smoldot repo. Anything that can automatically make tests fail is a bad thing.

The reason that we decided to "fetch" from smoldot instead of local is that we need a single place of specs. As soon as smoldot specs are updated without any notice when versions are updated, it is better to retrieve the specs from smoldot. 

If build breaks due to specs change, its a sign that something needs to be done on substrate-connect side and that is good to know actually.

There is a strong dependency between smoldot/substrate-connect and as soon as smoldot is the one using/handling the specs I would leave this dependency as is and not proceed with this PR.

tomaka:
> The reason that we decided to "fetch" from smoldot instead of local is that we need a single place of specs.

> If build breaks due to specs change, its a sign that something needs to be done on substrate-connect side and that is good to know actually.

You don't strictly need a single place for chain specifications. Chain specs never become invalid, that's the entire point of a blockchain. The only reason why we sometimes update them is when we no longer care about a chain (older versions of Rococo) or to update the checkpoint. In both cases there's no urgency to update.

It would not be a bad idea to automatically generate up-to-date checkpoints, but this should be done in an extremely secure way, by a Parity-owned physical server for example.

Furthermore, it makes more sense for specifications to reside in substrate-connect. In other words, it should be smoldot that downloads chain specs from the substrate-connect repo. Smoldot holds a list of chain specs only because it's convenient for me to test it, while substrate-connect actually ships the list of chain specs in its package and everyone using substrate-connect relies on them.

> As soon as smoldot specs are updated without any notice when versions are updated, it is better to retrieve the specs from smoldot.

No it's not strictly better. Maybe someone stole my password and changed the specs in smoldot. Maybe I made a mistake and named something `kusama.json` while in fact it's Polkadot.

wirednkod:
Latest change that happened to specs was the reduce of size. Taking that as example - it means that you should have commit a PR on substrate-connect to submit the minified specs to substrate-connect since that would be the single repo holding the specs, right?

> Maybe someone stole my password and changed the specs in smoldot

In this case that "someone" could alter lots of nice things around your github account - I dont think that specs in smoldot would be priority no 1.

Another solution I can think is that, we could also revert the endpoint from `smoldot` repo to `pokadot` main repo and retrieve the specs from there.

In case we keep single point of truth the substrate-connect (neither smoldot nor polkadot) probably you should include in this PR the latest specs for extension in respective dir as it is atm ignored from `.gitignore` (that is why the build fails)

tomaka:
My point is that there is no need for a single point of truth, because chain specs never become wrong or obsolete. A chain spec isn't more true than a different chain spec of the same chain.

> Latest change that happened to specs was the reduce of size. Taking that as example - it means that you should have commit a PR on substrate-connect to submit the minified specs to substrate-connect since that would be the single repo holding the specs, right?

I've actually reverted my change in smoldot, because otherwise I can't test the smoldot full node anymore: https://github.com/paritytech/smoldot/pull/1647

I suspect that the script isn't working, because you can see that there two files are clearly not the same:
 - https://github.com/paritytech/substrate-connect/blob/master/.chains/westend.json
 - https://github.com/paritytech/smoldot/blob/main/bin/westend.json

wirednkod:
> My point is that there is no need for a single point of truth, because chain specs never become wrong or obsolete. A chain spec isn't more true than a different chain spec of the same chain. The content of polkadot.json, for example, will always be correct.

Completely agree with that - but we would like to have the "latest" ones (remember version upgrades e.g. Rococo_v12 -> Rococo_v13) - isn't that right? By `latest` I mean latest bootnodes etc etc... 

> I've actually reverted my change in smoldot

Did not know that.

> I suspect that the script isn't working, because you can see that there two files are clearly not the same

Scripts that were removed were only comparing the `id` of each chainspec and not all the content.
Build is failing because during [prebuild](https://github.com/paritytech/substrate-connect/pull/603/files#diff-c503ada501a08565c380a3a344b61b1927e844c73021888e7cdde350309b8cecL26) phase, the specs were copied to the extension's dir for packaging the extension. Thus the errors similar to:
```
ERROR in ./src/background/index.ts 14:0-53
Module not found: Error: Can't resolve '../../public/assets/rococo.json' in '/home/runner/work/substrate-connect/substrate-connect/projects/extension/src/background'
```

Thats why I mentioned:
> In case we keep single point of truth the substrate-connect (neither smoldot repo nor polkadot repo) probably you should include in this PR the latest specs for extension in respective dir as it is atm ignored from .gitignore (that is why the build fails)

tomaka:
https://github.com/paritytech/substrate-connect/pull/870

PR #604, opened 2022-01-03
Remove husky and the precommit hook

tomaka:
I know this is opinionated, but removing this `husky` tool is to me a win, and I don't understand why commit hooks. They're just an annoyance, and CI is here to detect issues anyway.


wirednkod (approved):

.husky/pre-commit
```diff
@@ -1,5 +0,0 @@
-#!/bin/sh
-. "$(dirname "$0")/_/husky.sh"
-
-npx lint-staged
-yarn test
```
josepot, line 5: IMO this is the only line that should be removed, the rest is actually quite helpful and unobtrusive.

tomaka:
If someone wants to format the code in a precommit hook, they can just manually add a `.git/hooks/pre-commit` file that contains something like:

```bash
#!/bin/sh
yarn run format
```

josepot (approved):

PR #606, opened 2022-01-03
Merge and rename interfaces

wirednkod:
Following up tasks on #572 :
> Right now the protocol mentions the web page, the content script, and the extension as layers of an onion. This is technically correct, but this is _not_ what the `connection-extension-protocol` package is about. This package is only about the communication between the web page and the content script. What happens between the content script and the extension is an implementation detail of the extension and shouldn't be mentioned in `connection-extension-protocol` because it's out of scope.
>    - Consequently, we should merge the `ProviderMessageData` and `MessageToManager` interfaces into one single interface, and `ExtensionMessageData` and `MessageFromManager` into one single interface.
>    - Afterwards, we can rename them to something more simple such as `ToExtension` and `ToWebpage` or something like that. The words `Provider` and `Manager` shouldn't be used because they're implementation details.

### DIFF
--- a/packages/connect-extension-protocol/src/index.ts
+++ b/packages/connect-extension-protocol/src/index.ts
@@ -11,7 +11,7 @@
  * You can think of the protocol types like layers of an onion. The innermost
  * layer is the original JSON RPC request/responses. Then we wrap extra layers
  * (types) for the other 2 hops which then get peeled off at each hop.
- * The {@link ProviderMessageToExtension} and {@link ToApplication} represents
+ * The {@link ToExtension} and {@link ToApplication} represents
  * communication between the PolkadotJS provider in the app and the extension
  * (content and background scripts).
  *
@@ -74,10 +74,10 @@ export const extension = {
  * @remarks The browser wraps the message putting it in {@link data}
  */
 export interface ProviderMessage {
-  data: ProviderMessageToExtension
+  data: ToExtension
 }
 
-export interface ProviderMessageToExtension {
+export interface ToExtension {
   /** origin is used to determine which side sent the message **/
   origin: "extension-provider"
   /** The name of the app to be used for display purposes in the extension UI **/
@@ -110,7 +110,7 @@ export type ProviderListenHandler = (message: ExtensionMessage) => void
  */
 export const provider = {
   /** send a message from the app to the extension **/
-  send: (message: ProviderMessageToExtension): void => {
+  send: (message: ToExtension): void => {
     window.postMessage(message, "*")
   },
   /**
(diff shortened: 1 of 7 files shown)


tomaka (approved):

packages/connect-extension-protocol/src/index.ts
```diff
@@ -101,18 +89,10 @@ export interface ProviderMessageData {
...
-}
-
-/**
- * The message from the `ExtensionProvider` in the app that is intended to be
- * sent on to the manager in the extension background.
- */
-export interface MessageToManager {
   /** Type of the message. Defines how to interpret the {@link payload} */
-  type: "rpc" | "spec"
+  type?: "rpc" | "spec"
```
josepot, line 93: Why is the `type` now optional? :thinking:

packages/connect-extension-protocol/src/index.ts
```diff
@@ -101,18 +89,10 @@ export interface ProviderMessageData {
...
   /** Payload of the message -  a JSON encoded RPC request **/
-  payload: string
+  payload?: string
```
josepot, line 95: ditto

projects/extension/src/background/ConnectionManager.ts
```diff
@@ -386,6 +381,10 @@ export class ConnectionManager
         return this.#handleSpecMessage(msg, app)
       }
 
+      if (!msg.payload) {
+        return
```
josepot, line 385: shouldn't this have the same treatment as the other unrecognized messages?

packages/connect-extension-protocol/src/index.ts
wirednkod: `message?: MessageToManager` was optional. Now that the interfaces are merged, as a result these are optional

packages/connect-extension-protocol/src/index.ts
wirednkod: `message?: MessageFromManager` [was optional](https://github.com/paritytech/substrate-connect/blob/master/packages/connect-extension-protocol/src/index.ts#L53). Now that the interfaces are merged, as a result these are optional

projects/extension/src/background/ConnectionManager.ts
wirednkod: msg is already checked on the flow and is valid type-wise. If payload is empty then there is no reason to throw an error,  add it to the `pendingRequests` or propagate it to `healthChecker`.

tomaka (approved):

tomaka (approved):

PR #609, opened 2022-01-04
Remove extension and provider variables from connect-extension-protocol

wirednkod:

- Remove the `extension` and `provider` variables. It doesn't make sense to use the variable `provider` when in the extension and vice versa, so why are they available? Again, sending and listening is out of scope of the `connection-extension-protocol` package which should only provide the protocol.

### DIFF
--- a/packages/connect/src/ExtensionProvider/ExtensionProvider.ts
+++ b/packages/connect/src/ExtensionProvider/ExtensionProvider.ts
@@ -17,7 +17,6 @@ import {
   ToExtension,
   ExtensionMessage,
   ToApplication,
-  provider,
 } from "@substrate/connect-extension-protocol"
 
 const CONTENT_SCRIPT_ORIGIN = "content-script"
@@ -62,6 +61,10 @@ const ANGLICISMS: { [index: string]: string } = {
  */
 const CONNECTION_STATE_PINGER_INTERVAL = 2000
 let nextChainId = 1
+
+const sendMessage = (msg: ToExtension): void => {
+  window.postMessage(msg, "*")
+}
 
 /**
  * The ExtensionProvider allows interacting with a smoldot-based WASM light
@@ -290,7 +293,7 @@ export class ExtensionProvider implements ProviderInterface {
       ...this.#commonMessageData,
       action: "connect",
     }
-    provider.send(connectMsg)
+    sendMessage(connectMsg)
 
     // Once connect is sent - send rpc to extension that will contain the chainSpecs
     // for the extension to call addChain on smoldot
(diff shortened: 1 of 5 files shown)


packages/connect/src/ExtensionProvider/ExtensionProvider.ts
```diff
@@ -89,6 +89,15 @@ export class ExtensionProvider implements ProviderInterface {
...
   healthPingerInterval = CONNECTION_STATE_PINGER_INTERVAL
 
+  #provider = {
+    send: (message: ToExtension): void => {
+      window.postMessage(message, "*")
+    },
+    listen: (handler: ProviderListenHandler): void => {
+      window.addEventListener("message", handler)
+    },
+  }
```
tomaka, line 99: Do we need this? Why not just inline the call to `postMessage` and `addEventListener`?

projects/extension/src/content/ExtensionMessageRouter.test.ts
```diff
@@ -18,6 +18,15 @@ const waitForMessageToBePosted = (): Promise<null> => {
...
 }
 
+const provider = {
+  send: (message: ToExtension): void => {
+    window.postMessage(message, "*")
+  },
+  listen: (handler: ProviderListenHandler): void => {
+    window.addEventListener("message", handler)
+  },
+}
```
tomaka, line 28: Same remark

projects/extension/src/content/ExtensionMessageRouter.ts
```diff
@@ -24,6 +25,18 @@ const EXTENSION_PROVIDER_ORIGIN = "extension-provider"
...
+    send: (message: ToApplication): void => {
+      window.postMessage(message, "*")
+    },
+    listen: (handler: ExtensionListenHandler): void => {
+      window.addEventListener("message", handler)
+    },
+    removeListen: (handler: ExtensionListenHandler): void => {
+      window.removeEventListener("message", handler)
+    },
+  }
```
tomaka, line 38: And same remark

packages/connect/src/ExtensionProvider/ExtensionProvider.ts
wirednkod: I started with that but I decided to reuse functions instead of repeating `window.postMessage` etc etc... Makes lines clearer.

I don't think this should block the PR.

wirednkod: I started with that but I altered it to using a function as it makes reduces repeating `window` etc again and again.

I dont think this should block the PR

josepot: I agree with @tomaka, I would much rather having it inlined, for different reasons:

1) It makes the next change (the removal of ExtensionMessage and ProviderMessage interfaces) a lot easier.
2) IMO it's a bogus abstraction that adds a layer of indirection for no good reason.
3) It's not true that it makes the code more concise, as the current git diff is `+51 -56` while the git diff of its inlined cunter part would be `+34 -63`.

Since @wirednkod would be traveling and he has asked me to take over, I will force push those 2 commits with the inlined code.

tomaka (approved):

PR #611, opened 2022-01-07
remove `ExtensionMessageData.disconnect`

josepot:

- Remove the `ExtensionMessageData.disconnect` field, because it's exactly the same thing as `MessageFromManager.type` being equal to `error`.

### DIFF
--- a/projects/extension/src/content/ExtensionMessageRouter.ts
+++ b/projects/extension/src/content/ExtensionMessageRouter.ts
@@ -71,7 +71,11 @@ export class ExtensionMessageRouter {
 
     // tell the page when the port disconnects
     port.onDisconnect.addListener(() => {
-      sendMessage({ origin: "content-script", disconnect: true })
+      sendMessage({
+        origin: "content-script",
+        type: "error",
+        payload: "Lost communication with substrate-connect extension",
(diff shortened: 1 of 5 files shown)


projects/extension/src/content/ExtensionMessageRouter.ts
```diff
@@ -71,7 +71,11 @@ export class ExtensionMessageRouter {
 
     // tell the page when the port disconnects
     port.onDisconnect.addListener(() => {
-      sendMessage({ origin: "content-script", disconnect: true })
+      sendMessage({
+        origin: "content-script",
+        type: "error",
+        payload: "Disconnected!",
```
josepot, line 77: I don't like this `payload`... Suggestions? @tomaka

tomaka: What happens is the extension has been installed or has decided by ban the application or something like that.
You can put "Lost communication with substrate-connect extension".

josepot: yeah, I like that one better... I will update it, thanks!

tomaka (approved):

PR #614, opened 2022-01-07
remove `action` field from `ExtensionProvider`

josepot:

- Remove the `action` field, because all messages are automatically considered as `forward`. There's no need for the `connect` and `disconnect` messages. The code of `disconnect` can simply be removed for now. Update the content script of the extension to simply connect to the extension and forward messages back and forth. It doesn't have to do anything more.

### DIFF
--- a/packages/connect-extension-protocol/src/index.ts
+++ b/packages/connect-extension-protocol/src/index.ts
@@ -36,12 +36,10 @@ export interface ToExtension {
   chainId: number
   /** The name of the blockchain network the app is talking to **/
   chainName: string
-  /** What action the `ExtensionMessageRouter` should take **/
-  action: "forward" | "disconnect"
   /** The message the `ExtensionMessageRouter` should forward to the background **/
   /** Type of the message. Defines how to interpret the {@link payload} */
-  type?: "rpc" | "spec"
+  type: "rpc" | "spec"
   /** Payload of the message -  a JSON encoded RPC request **/
-  payload?: string
+  payload: string
   parachainPayload?: string
 }
--- a/projects/extension/src/content/ExtensionMessageRouter.ts
+++ b/projects/extension/src/content/ExtensionMessageRouter.ts
@@ -52,6 +52,7 @@ export class ExtensionMessageRouter {
     const port = chrome.runtime.connect({
       name: `${window.location.href}::${chainName}`,
     })
+
     debug(`CONNECTED ${chainName} PORT`, port)
 
     // forward any messages: extension -> page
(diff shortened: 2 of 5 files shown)


packages/connect-extension-protocol/src/index.ts
```diff
@@ -36,12 +36,10 @@ export interface ToExtension {
...
   /** The name of the blockchain network the app is talking to **/
   chainName: string
-  /** What action the `ExtensionMessageRouter` should take **/
-  action: "forward" | "disconnect"
   /** The message the `ExtensionMessageRouter` should forward to the background **/
   /** Type of the message. Defines how to interpret the {@link payload} */
-  type?: "rpc" | "spec"
+  type: "rpc" | "spec"
   /** Payload of the message -  a JSON encoded RPC request **/
-  payload?: string
```
josepot, line 45: Notice that now there is no longer a reason for `type` and `payload` to be optional.

projects/extension/src/content/ExtensionMessageRouter.ts
```diff
@@ -101,56 +102,28 @@ export class ExtensionMessageRouter {
...
-      return console.warn("Malformed message - missing action", msg)
-    }
-
-    if (action === "disconnect") {
-      return this.#disconnectPort(data)
+    if (!type) {
+      // probably someone abusing the extension
+      console.warn("Malformed message - missing message.type", data)
+      return
     }
```
tomaka, line 118: If `type` is no longer optional we can remove this, no?

tomaka (approved):

projects/extension/src/content/ExtensionMessageRouter.ts
josepot: I guess it's still relevant in case that the consumer is not using TS.

tomaka: TS is supposed to check whether this is valid. If we can't rely on TS's checks, then there's a ton of things to verify.

josepot: what I meant is that it's possible that someone writes a client in plain JS that interacts with the Extension... TS doesn't do anything on the runtime.

tomaka: So when the content script does this: https://github.com/paritytech/substrate-connect/blob/73d0e96ae46d7b309eb5c73a8aa2bfdbde90c417/projects/extension/src/content/ExtensionMessageRouter.ts#L43

And someone does `window.postMessage({ "some_garbage": 5 })`, then `handleMessage` will be called with `{ "some_garbage": 5 }` passed as parameter and throw exceptions?

PR #617, opened 2022-01-09
remove `chainName` field from `ToExtension`

josepot:

- Remove the `chainName` field. Instead the extension should use the chain name found in the chain spec.

While I was working on splitting the message type `spec` into `add-chain`/`add-well-known-chain`, I realized that change was going to be a lot smaller and simpler if the `chainName` field had already been removed, so I decided to tackle this one first.

### DIFF
--- a/packages/connect-extension-protocol/src/index.ts
+++ b/packages/connect-extension-protocol/src/index.ts
@@ -34,8 +34,6 @@ export interface ToExtension {
   origin: "extension-provider"
   /** The uniqueId for extension multiplexing **/
   chainId: number
-  /** The name of the blockchain network the app is talking to **/
-  chainName: string
   /** The message the `ExtensionMessageRouter` should forward to the background **/
   /** Type of the message. Defines how to interpret the {@link payload} */
   type: "rpc" | "spec"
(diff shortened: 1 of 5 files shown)


wirednkod (changes requested):
There are some more changes concerning `chainName`. It is used in the [background](https://github.com/paritytech/substrate-connect/blob/master/projects/extension/src/background/ConnectionManager.ts#L127) when app is created, for the notifications of the extension, and as an info for the connected app ([types](https://github.com/paritytech/substrate-connect/blob/master/projects/extension/src/background/types.ts#L9)) and showcase on [Options page](https://github.com/paritytech/substrate-connect/blob/master/projects/extension/src/containers/Options.tsx#L108)).

tomaka:
> There are some more changes concerning chainName. It is used in the background when app is created, for the notifications of the extension, and as an info for the connected app (types) and showcase on Options page).

This PR doesn't remove the concept of name, it only removes the `chainName` field.
Before this PR, the name was passed at the same time in the chain spec and in a separate field. This PR only removes the separate field as it is redundant.

tomaka (approved):

wirednkod (approved):

PR #618, opened 2022-01-09
split `MessageToManager.type` into `add-chain` and `add-well-known-chain`

josepot:

-  `MessageToManager.type` equal to `spec` should instead be two different variants: `add-chain` or `add-well-known-chain`. Depending on which one, you pass either a chain spec or the id of a well known name ("rococo_1_3", "westend", "polkadot", or "ksmcc3").

### DIFF
--- a/packages/connect/src/ExtensionProvider/ExtensionProvider.ts
+++ b/packages/connect/src/ExtensionProvider/ExtensionProvider.ts
@@ -17,6 +17,7 @@ import {
   ToExtension,
   ToApplication,
 } from "@substrate/connect-extension-protocol"
+import { SupportedChains } from "../specs/index.js"
 
 const CONTENT_SCRIPT_ORIGIN = "content-script"
 const EXTENSION_PROVIDER_ORIGIN = "extension-provider"
@@ -273,8 +274,10 @@ export class ExtensionProvider implements ProviderInterface {
     // for the extension to call addChain on smoldot
     const specMsg: ToExtension = {
       ...this.#commonMessageData,
-      type: "spec",
-      payload: this.#chainSpecs || "",
+      type: SupportedChains[this.#chainSpecs as SupportedChains]
+        ? "add-well-known-chain"
+        : "add-chain",
+      payload: this.#chainSpecs,
     }
     if (this.#parachainSpecs) {
       specMsg.parachainPayload = this.#parachainSpecs
--- a/projects/extension/src/background/ConnectionManager.ts
+++ b/projects/extension/src/background/ConnectionManager.ts
@@ -15,6 +15,17 @@ import { StateEmitter, State } from "./types"
 import { NetworkMainInfo, Network } from "../types"
 import { logger } from "@polkadot/util"
 import { ToExtension } from "@substrate/connect-extension-protocol"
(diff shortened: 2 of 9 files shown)


packages/connect/src/ExtensionProvider/ExtensionProvider.ts
```diff
@@ -273,8 +274,10 @@ export class ExtensionProvider implements ProviderInterface {
     // for the extension to call addChain on smoldot
     const specMsg: ToExtension = {
       ...this.#commonMessageData,
-      type: "spec",
-      payload: this.#chainSpecs || "",
+      type: SupportedChains[this.#chainSpecs as SupportedChains]
+        ? "add-well-known-chain"
+        : "add-chain",
+      payload: this.#chainSpecs,
```
tomaka, line 280: That's not great but we can deal with this later

projects/extension/src/background/ConnectionManager.ts
```diff
@@ -15,6 +15,17 @@ import { StateEmitter, State } from "./types"
...
+import kusama from "../../public/assets/kusama.json"
+import polkadot from "../../public/assets/polkadot.json"
+import rococo from "../../public/assets/rococo.json"
+
+export const relayChains: Map<string, string> = new Map<string, string>([
+  ["polkadot", JSON.stringify(polkadot)],
+  ["kusama", JSON.stringify(kusama)],
+  ["rococo", JSON.stringify(rococo)],
+  ["westend", JSON.stringify(westend)],
+])
```
tomaka, line 28: Same remark: that's not great, but let's delay fixing this for after #572

projects/extension/src/background/ConnectionManager.ts
```diff
@@ -15,6 +15,17 @@ import { StateEmitter, State } from "./types"
 import { NetworkMainInfo, Network } from "../types"
 import { logger } from "@polkadot/util"
 import { ToExtension } from "@substrate/connect-extension-protocol"
+import westend from "../../public/assets/westend.json"
+import kusama from "../../public/assets/kusama.json"
+import polkadot from "../../public/assets/polkadot.json"
+import rococo from "../../public/assets/rococo.json"
+
+export const relayChains: Map<string, string> = new Map<string, string>([
```
tomaka, line 23:
```suggestion
export const wellKnownChains: Map<string, string> = new Map<string, string>([
```

I know that in practice well known chains are always relay chains, but they're still separate concepts

tomaka (approved):

projects/extension/src/background/ConnectionManager.ts
josepot: Yes, for sure, this is definitely pretty bad... However, this was [here before, I just moved it to the correct file](https://github.com/paritytech/substrate-connect/pull/618/files#diff-4de63d371461c5b51ea9a173eccc4c3b114374baa9388b564a3033e112f58924L18-L26) in order to avoid circular dependencies.

projects/extension/src/background/ConnectionManager.ts
josepot: totally!

wirednkod (approved):

PR #621, opened 2022-01-10
Add `chainId` into `ToApplication` interface

josepot:

- The `chainId` field should be present in the `ToApplication` interface so that the listener can filter-out those messages that were not targeting its chain.

This also addresses the "routing" issue that's described in #550

### DIFF
--- a/packages/connect-extension-protocol/src/index.ts
+++ b/packages/connect-extension-protocol/src/index.ts
@@ -23,6 +23,8 @@
 export interface ToApplication {
   /** origin is used to determine which side sent the message **/
   origin: "content-script"
+  /** The uniqueId for extension multiplexing **/
+  chainId: number
   /** Type of the message. Defines how to interpret the {@link payload} */
   type: "error" | "rpc"
   /** Payload of the message. Either a JSON encoded RPC response or an error message **/
--- a/packages/connect-extension-protocol/src/index.ts
+++ b/packages/connect-extension-protocol/src/index.ts
@@ -23,7 +23,7 @@
 export interface ToApplication {
   /** origin is used to determine which side sent the message **/
   origin: "content-script"
-  /** The uniqueId for extension multiplexing **/
+  /** Which chain this message applies to **/
   chainId: number
   /** Type of the message. Defines how to interpret the {@link payload} */
   type: "error" | "rpc"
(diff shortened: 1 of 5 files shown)


packages/connect/src/ExtensionProvider/ExtensionProvider.ts
```diff
@@ -96,10 +96,11 @@ export class ExtensionProvider implements ProviderInterface {
...
-    this.#commonMessageData = {
-      chainId: nextChainId++,
-      origin: EXTENSION_PROVIDER_ORIGIN,
-    }
+    this.#chainId = nextChainId++
+  }
+
+  public get chainId(): number {
+    return this.#chainId
   }
```
tomaka, line 104: Why is this public?

packages/connect-extension-protocol/src/index.ts
```diff
@@ -23,6 +23,8 @@
 export interface ToApplication {
   /** origin is used to determine which side sent the message **/
   origin: "content-script"
+  /** The uniqueId for extension multiplexing **/
```
tomaka, line 26:
```suggestion
  /** Which chain this message applies to **/
```

packages/connect/src/ExtensionProvider/ExtensionProvider.ts
josepot: honest answer... So that we can test it more easily... I know, a bit of a shame, but the alternative was quite convoluted to make the tests work.

packages/connect-extension-protocol/src/index.ts
josepot: yeah, sorry, that was a copy-pasta from the `chainId` of the `ToExtension`

packages/connect/src/ExtensionProvider/ExtensionProvider.ts
tomaka: > we leave it private

I mean remove the function

josepot: It's been used in the tests, though. [Here](https://github.com/paritytech/substrate-connect/pull/621/files#diff-162c74379f51c3494d8d9f03d810b9b2a55e688b7b2e72a7e21036516d3c76c3R137) and [here](https://github.com/paritytech/substrate-connect/pull/621/files#diff-162c74379f51c3494d8d9f03d810b9b2a55e688b7b2e72a7e21036516d3c76c3R194).

tomaka (approved):

PR #627, opened 2022-01-11
`chainId` counter -> random string

josepot:

  - [ ] The `chainId` field should become a string, and its value is decided by the web page by [generating a big random number](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues). Its value is here to identify which chain amongst the list of all chains that the web page has opened.

### DIFF
--- a/packages/connect-extension-protocol/src/index.ts
+++ b/packages/connect-extension-protocol/src/index.ts
@@ -24,7 +24,7 @@ export interface ToApplication {
   /** origin is used to determine which side sent the message **/
   origin: "content-script"
   /** Which chain this message applies to **/
-  chainId: number
+  chainId: string
   /** Type of the message. Defines how to interpret the {@link payload} */
   type: "error" | "rpc"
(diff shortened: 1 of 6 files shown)


wirednkod (approved):

tomaka:
I'm really not a fan of this way of testing. It really complicates everything just for the sake of writing tests that do not bring anything.

tomaka (approved):
I guess it unlocks us, so let's go with this

josepot:
> I'm really not a fan of this way of testing. It really complicates everything just for the sake of writing tests that do not bring anything.

As I've mentioned (more than once) in our Element chats: I think that we should improve the tests of `substrate-connect` as IMO the few tests that we have don't make me feel confident that the public APIs work as they should. In the meanwhile, though, rewriting the tests is out of scope for this task, so I'm just trying to make the current tests pass. Although, I'm more than open to suggestions into how to better write these tests without a complete refactor.

PR #629, opened 2022-01-11
Correct Documentation

wirednkod:
Correct documentation of TSDocs and Landing page


tomaka (approved):
This is going to change again, but why not

PR #634, opened 2022-01-12
Burnr - cleanup code

wirednkod:
Update the dependencies of burnr wallet and cleanup the code.


projects/burnr/src/Home.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { useState, useEffect } from "react"
+import * as React from "react"
```
josepot, line 1: For consistency with the rest of the code base, could we please change this to?

```ts
import { useState, useEffect } from "react"
```

projects/burnr/src/index.tsx
```diff
@@ -1,8 +1,5 @@
-// workaround for parcel in order to work with async/await
-import "regenerator-runtime/runtime"
-
-import React from "react"
-import ReactDOM from "react-dom"
+import * as React from "react"
```
josepot, line 1: Since React 17 there is [no need to import React for using JSX](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html), so con we please get rid of this import?

projects/burnr/src/index.tsx
```diff
@@ -1,8 +1,5 @@
...
+import * as ReactDOM from "react-dom"
```
josepot, line 2: No need to import the whole package, this should suffice:
```ts
import { render } from "react-dom"
```

projects/burnr/src/hooks/api/useApiCreate.ts
```diff
@@ -1,17 +1,15 @@
...
 import { ApiPromise } from "@polkadot/api"
 import { logger } from "@polkadot/util"
 import { ScProvider } from "@substrate/connect"
 import { ALL_PROVIDERS, BURNR_WALLET } from "../../utils/constants"
-import { useIsMountedRef } from ".."
+import { useIsMountedRef } from "./useIsMountedRef"
 
 const l = logger(BURNR_WALLET)
 
-export default function useApiCreate(): ApiPromise {
```
tomaka, line 14: Why is the `default` not correct?

projects/burnr/tsconfig.json
```diff
@@ -15,5 +15,6 @@
     "jsx": "react"
   },
   "include": ["src/**/*"],
+  "exclude": ["*node_modules*", "../../node_modules*"],
```
josepot, line 18: why?

josepot (changes requested):

projects/burnr/src/hooks/api/useApiCreate.ts
wirednkod: It is not a matter of `not good`. Its a matter of personal decision and consistency across code-base. 
TBH I prefer naming exports instead of default as it makes less error-prone and more consistent. But this is a personal opinion ofc. Both ways (using default or naming exports) are correct

projects/burnr/src/Home.tsx
wirednkod: This gives an error:
```
'React' refers to a UMD global, but the current file is a module. Consider adding an import instead
```

projects/burnr/src/index.tsx
wirednkod: This gives an error:
```
'React' refers to a UMD global, but the current file is a module. Consider adding an import instead
```

projects/burnr/tsconfig.json
wirednkod: Cause I thought there is no need to include node modules. I will remove it

josepot (approved):

PR #635, opened 2022-01-13
Landing page - Cleanup

wirednkod:
- remove non needed dependencies
- update versions of remaining dependencies
- fix eslint issues


projects/landing-page/src/App.tsx
```diff
@@ -267,7 +266,7 @@ const App: React.FunctionComponent = () => {
             </SectionText>
             <CardProject
               imageProps={{
-                path: Extension,
+                path: Extension as string,
```
tomaka, line 269: Why are these `as string` needed all of a sudden?

wirednkod: These are needed in order to avoid the `/* eslint-disable @typescript-eslint/no-unsafe-assignment */` disabling at the beginning of the file

tomaka (approved):

PR #637, opened 2022-01-13
Connect Extension Protocol - Update dependencies

wirednkod:


packages/connect-extension-protocol/package.json
```diff
@@ -14,9 +14,7 @@
   "author": "Parity Team <admin@parity.io>",
   "license": "GPL-3.0-only",
   "devDependencies": {
-    "@typescript-eslint/parser": "^5.5.0",
-    "eslint": "^8.5.0",
-    "eslint-plugin-tsdoc": "^0.2.14",
-    "typescript": "^4.5.2"
+    "eslint": "^8.6.0",
+    "typescript": "^4.5.4"
```
josepot, line 18: I would much rather for these 2 to be lifted to the root `package.json`, though. Since, these are 2 devDependencies that we use in every single workspace.

wirednkod: I plan to move to root package once all updates are done

josepot (approved):

Issue #571, opened 2021-12-13
Modify the `Background` interface to be more precise


tomaka:
The interface defined [here](https://github.com/paritytech/substrate-connect/blob/bfaf68b2d96f5f02702f97f0d8f6ad17c0004720/projects/extension/src/background/index.ts#L5) is too loose.

It should give access to the strict minimum of information that the extension needs, like the list of chains with their name and health.
In particular, we shouldn't expose the `ConnectionManager`.

tomaka:
To clarify: this `interface` should only expose getters/functions that use primitive types in JavaScript, or interfaces that use only primitives types.

The objective is that this interface should be completely abstract.

josepot:
Addressed on #656

PR #656, opened 2022-01-18
refactor ConnectionManager

josepot:
I feel bad for opening such a large PR. So, I would like to explain how and why it came to this.

When I started working on #571, the very first thing that I did was to define a new public interface that looked like this:

```ts
export interface ExposedChainConnection {
  chainId: string;
  chainName: string;
  tabId: number;
  url: string;
  healthStatus?: SmoldotHealth;
}

export interface Background extends Window {
  manager: {
    onManagerStateChanged: (
      listener: (state: ExposedChainConnection[]) => void
    ) => () => void;
    disconnectTab: (tabId: number) => void;
  };
}
```

Which is all what needs to be exposed so that the UI can display the state of the different connections.

Next, I updated the code in the React components to leverage this new API. So far, so good.

Up until this point, I hadn't even looked at the actual implementation of `ConnectionManager.ts`. However, when I went to make the implementation compliant with the new API, then I noticed that there were lots of types (`PopupAppInfo`, `AppState`, `App`, `NetworkState`, etc) that were no longer needed. Probably the only reason why they existed is b/c `ConnectionManager` was tightly coupled to the UI.

Also, given the way that the `ConnectionManager` was keeping track of all that bogus state, combined with the fact that the public API was exposing too much, that ended up being the perfect recipe for producing some annoying bugs (`disconnectTab` would leave the ports connected, the UI was not always displaying all the active networks in a tab, there were cases when not all `potentialRelayChains` were being passed, etc). Also, the tests mainly consisted in testing a public interface that was too loose and that basically made assertions about the UI state. So, with the new API those tests were no longer relevant (or useful).

So, I started fixing the public API but I ended up falling into a slippery slope that led me to a complete refactor of the `ConnectionManager`. I got rid of all those things that were no longer needed, and I think that the code is now a lot more straight forward. Then I replaced the previous `ConnectionManager` tests with a new set of tests.

About the approach that I took for testing: the `ConnectionManager` is responsible for managing the different connections that come from the content-script ports and ensuring that the communication between smoldot and these ports works correctly, while also being responsible for letting other entities know about the relevant parts of its state (`ExposedChainConnection[]`) whenever it changes.

Therefore, these are the things that the tests should be testing. In order to do that, these new tests mock the smoldot client and the chrome ports in order to make sure that the interactions between them work correctly, and they also make sure that `ConnectionManager` emits an event with the correct `ExposedChainConnection[]` when the chains are being added/removed or when the health of one of them changes.

It's worth pointing out that currently the test-coverage stats on the `master` branch for the `extension` workspace look like this:

```bash
Statements   : 76.52% ( 176/230 )
Branches     : 49.61% ( 64/129 )
Functions    : 81.69% ( 58/71 )
Lines        : 76.36% ( 168/220 )
```

while the same stats on this branch look like this:

```bash
Statements   : 94.62% ( 176/186 )
Branches     : 80.51% ( 62/77 )  
Functions    : 94.82% ( 55/58 )  
Lines        : 94.47% ( 171/181 )
```

So, in the end this PR does a lot more than what #571 was strictly about. However, I think that these tests justify that these changes are solid and worthy of being merged into `master`.

### DIFF
--- a/projects/extension/src/background/ConnectionManager.ts
+++ b/projects/extension/src/background/ConnectionManager.ts
@@ -3,16 +3,16 @@
 /* eslint-disable @typescript-eslint/no-unsafe-call */
 /* eslint-disable @typescript-eslint/no-unsafe-member-access */
 /* eslint-disable @typescript-eslint/no-explicit-any */
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
 import {
   Client,
-  start,
   healthChecker as smHealthChecker,
+  Chain,
 } from "@substrate/smoldot-light"
 import { JsonRpcCallback, SmoldotHealth } from "@substrate/smoldot-light"
-import { ExposedAppInfo, App, ConnectionManagerInterface } from "./types"
+import { ExposedChainConnection, ChainConnection } from "./types"
 import EventEmitter from "eventemitter3"
-import { StateEmitter, State } from "./types"
-import { NetworkMainInfo, Network } from "../types"
+import { StateEmitter } from "./types"
 import { logger } from "@polkadot/util"
 import { ToExtension } from "@substrate/connect-extension-protocol"
 import westend from "../../public/assets/westend.json"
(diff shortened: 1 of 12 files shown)


projects/extension/src/background/ConnectionManager.ts
```diff
@@ -244,195 +163,152 @@ export class ConnectionManager
    *
    * @returns addedChain - An the newly added chain info
    */
-  async addChain(
+  addChain(
     chainSpec: string,
     jsonRpcCallback?: JsonRpcCallback,
     tabId?: number,
-  ): Promise<Network> {
+  ): Promise<Chain> {
```
tomaka, line 170: Not a problem, but I'm surprised that this works, I thought you had to always use both `async` and `Promise`.

tomaka (approved):
The code is obviously not ideal even after this change, but since the goal is just to unlock progress that's ok.

projects/extension/src/background/ConnectionManager.ts
josepot: Nope, you can return a `Promise` without using `async`. There are many instances when that's actually necessary/preferable. On the other hand, creating an `async` function without using `await` is "dangerous" because you may inadvertently creating a `Promise` for code that otherwise could be synchronous.

projects/extension/src/background/index.ts
```diff
@@ -1,15 +1,36 @@
+/* eslint-disable @typescript-eslint/no-floating-promises */
```
wirednkod, line 1: You can avoid this by adding a `.catch((err) => console.log(err))` at the promise

projects/extension/src/components/Tab.tsx
```diff
@@ -70,7 +69,7 @@ const Tab: FunctionComponent<TabProps> = ({
       /* TODO(nik): Fix smoldot definition (see: https://github.com/paritytech/substrate-connect/blob/3350cdff9c4c294393160189816168a93c983f79/projects/extension/src/background/ConnectionManager.ts#L202)
        ** eslint disable below seems to be due to smoldot definition */
       // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
-      manager?.disconnectTab(tab.tabId)
+      disconnectTab(tab.tabId)
```
wirednkod, line 72: If I understood well - and we want actually yo have multiple apps in each tab - then we should not disconnect the Tab but the app inside the tab.

projects/extension/src/background/index.ts
josepot: why? the promise is already handling the errors, the callback function of that catch will never be called.

projects/extension/src/components/Tab.tsx
josepot: whats an "app"?

wirednkod: > whats an "app"?

sorry - I meant networks connections

wirednkod (approved):

PR #660, opened 2022-01-19
replace regenerator-runtime with .babelrc

wirednkod:


tomaka:
You screwed up a bit when merging master, there's files in the diff that shouldn't be there.

wirednkod:
> You screwed up a bit when merging master, there's files in the diff that shouldn't be there.

True.. I'll update this to remove the .babelrc for removed projects. Sorry about that

tomaka (approved):

PR #663, opened 2022-01-20
Remove old Kusama bootnodes

tomaka:
https://github.com/paritytech/polkadot/pull/2319


wirednkod:
This should be merged after #603 ? Or else the added bootnodes may be replaced again

PR #667, opened 2022-01-20
add `chain-ready` message type

josepot:

- When you send a message to add a chain, the extension should return ~~two~~ one message~~s~~ that ~~don't~~ doesn't exist yet: ~~one "add-chain-started" message immediately returned, and that is used to know whether the extension exists, and~~ one "add-chain-ok" or "add-chain-error" message. All ~~three~~ two messages indicate which `chainId` they refer to, and "add-chain-error" also provides an `error` string.

Now that we have the `add-chain-ok` message, there is no reason for the `ConnectionManager` to try to handle `RPC` messages that came too early. Instead, now it should send an error to the content-script if an RPC message comes before the `add-chain-ok` message has been sent. Therefore, this logic and the related tests have been updated accordingly.

Regarding the code of the `ExtensionProvider`: I've tried to make as little changes as possible while still ensuring that the `Promise` returned by `connect` now doesn't resolve until it receives the `add-chain-ok` response from the Extension. I've also tried to make as little changes as possible to the tests of the `ExtensionProvider` b/c well... it's out of the scope of this PR to improve those tests, but IMO they should be improved at some point.

### DIFF
--- a/packages/connect-extension-protocol/src/index.ts
+++ b/packages/connect-extension-protocol/src/index.ts
@@ -26,7 +26,7 @@ export interface ToApplication {
   /** Which chain this message applies to **/
   chainId: string
   /** Type of the message. Defines how to interpret the {@link payload} */
-  type: "error" | "rpc"
+  type: "error" | "rpc" | "add-chain-ok"
   /** Payload of the message. Either a JSON encoded RPC response or an error message **/
   payload: string
 }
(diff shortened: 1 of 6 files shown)


tomaka:
What about naming this `chain-ready` instead of `add-chain-ok`, since there's going to be no `add-chain-error` in the end?

wirednkod (approved):

PR #674, opened 2022-01-21
re-add the web-ext dev dependency

wirednkod:


josepot:
Out of curiosity: where and how is `web-ext` being used?

wirednkod:
For running chrome and mozilla extension
https://github.com/paritytech/substrate-connect/blob/master/projects/extension/package.json#L18

josepot:
> For running extension in chrome and mozilla (without having to add it from dev tools - extremelly helpful especially for mozilla) https://github.com/paritytech/substrate-connect/blob/master/projects/extension/package.json#L18

Got it, thanks! So, since this dependency is very specific to the `extension`, could we please just add it in the `package.json` of its own workspace?

wirednkod:
Sure. Though its a dev dependency. I thought all of them can be on root package

josepot:
> Sure. Though its a dev dependency. I thought all of them can be on root package

All the common ones, yes

josepot (approved):

PR #677, opened 2022-01-24
improve typings of `ToExtension` & `ToApplication`

josepot:

- Split `ToApplication` and `ToExtension` in multiple messages based on their type, so that `add-chain`, `remove-chain`, etc. can be passed different fields. The `chainId` field shouldn't be directly in `ToApplication` and `ToExtension` but in each individual message.

I initially thought that it would make sense to also get rid of the `parachainSpec` property in this PR, but things started to get a bit out of hand. So, I decided to tackle that in the next PR, as it's very much related to the effort of adding the `potentialRelayChainIds` property into the `add-chain` message.

### DIFF
--- a/packages/connect-extension-protocol/src/index.ts
+++ b/packages/connect-extension-protocol/src/index.ts
@@ -22,6 +22,9 @@
 
 interface ToApplicationHeader {
   origin: "content-script"
+}
+
+type ToApplicationChainHeader = ToApplicationHeader & {
   chainId: string
 }
 
@@ -39,11 +42,14 @@ interface ToApplicationRpc {
   payload: string
 }
 
-export type ToApplication = ToApplicationHeader &
+export type ToApplication = ToApplicationChainHeader &
   (ToApplicationError | ToApplicationChainReady | ToApplicationRpc)
 
 interface ToExtensionHeader {
   origin: "extension-provider"
+}
+
+type ToExtensionChainHeader = ToExtensionHeader & {
   chainId: string
 }
 
@@ -66,7 +72,7 @@ interface ToExtensionRemoveChain {
   type: "remove-chain"
 }
 
-export type ToExtension = ToExtensionHeader &
+export type ToExtension = ToExtensionChainHeader &
   (
     | ToExtensionAddChain
     | ToExtensionAddWellKnownChain
--- a/packages/connect-extension-protocol/src/index.ts
+++ b/packages/connect-extension-protocol/src/index.ts
@@ -20,26 +20,56 @@
  * The {@link ConnectionManager} is the class in the extension background.
  */
 
-export interface ToApplication {
-  /** origin is used to determine which side sent the message **/
+interface ToApplicationHeader {
   origin: "content-script"
-  /** Which chain this message applies to **/
   chainId: string
-  /** Type of the message. Defines how to interpret the {@link payload} */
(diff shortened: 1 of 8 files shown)


tomaka:
> The chainId field shouldn't be directly in ToApplication and ToExtension but in each individual message.

So I guess you disagree with me, but what I meant is exactly that `chainId` shouldn't be `ToApplicationHeader` or `ToExtensionHeader`, but in each individual message.

There will inevitably be a day where someone adds a message that doesn't need a `chainId`, and instead of moving the `chainId` field like they should they will instead incorrectly pass a dummy value.

josepot:
> > The chainId field shouldn't be directly in ToApplication and ToExtension but in each individual message.
> 
> So I guess you disagree with me, but what I meant is exactly that `chainId` shouldn't be `ToApplicationHeader` or `ToExtensionHeader`, but in each individual message.
> 
> There will inevitably be a day where someone adds a message that doesn't need a `chainId`, and instead of moving the `chainId` field like they should they will instead incorrectly pass a dummy value.

I don't disagree with you! Please notice that neither `ToApplicationHeader` nor `ToExtensionHeader` are being exported :slightly_smiling_face:  They are just there for convenience. What's exported is:

```ts
export type ToExtension = ToExtensionHeader &
  (
    | ToExtensionAddChain
    | ToExtensionAddWellKnownChain
    | ToExtensionRpc
    | ToExtensionRemoveChain
  )
 ```

and

```ts
export type ToApplication = ToApplicationHeader &
  (ToApplicationError | ToApplicationChainReady | ToApplicationRpc)
```

josepot:
> Yeah I understand that it's the exact same thing anyway from the point of view of the user, but I'm placing myself from the point of view of someone who adds a new message to `substrate-connect-protocol` in the future, and gets "blocked" because they don't know which value of `chainId` to provide for their new message.

I mean, if they know basic TS they wouldn't get blocked, they could simply do:

```ts
type ToExtensionChain = ToExtensionHeader &
  (
    | ToExtensionAddChain
    | ToExtensionAddWellKnownChain
    | ToExtensionRpc
    | ToExtensionRemoveChain
  )

type ToExtensionGenericStop = Pick<ToExntensionHeader, 'origin'> & {
  type: 'stopEverything'
}

export type ToExtension = ToExtensionChain | ToExtensionGenericStop
```

That being said, let me make a small change that perhaps could make things even easier for whenever that happens... Or if you want I can just explicitly add all the properties in all the messages, up to you. I really don't feel so strongly about this.

tomaka:
I mean, if it was me I'd do this because it's the most simple: https://gist.github.com/tomaka/6e1024f72f4590f44aa558a5a3869b89

josepot:
So, with the latest changes if that ever happened the developer would only have to rename the current `ToExtension` type into `ToExtensionChan` while making it private and then add this 2 lines:

```ts
type ToExtensionGenericStop = ToExtensionHeader & {
  type: 'stopEverything'
}

export type ToExtension = ToExtensionChain | ToExtensionGenericStop
```

That's it. But again, if you rather me removing the internal "header" types altogether, I can do that too, for sure!

tomaka:
I don't really care about the `origin` field, you can leave it in a header, but I strongly think it's more correct to have `chainId` in each message individually and not in any sort of header.

tomaka (approved):

PR #680, opened 2022-01-25
Update databaseContent from smoldot to local storage

### LINKED ISSUE #568: Use smoldot's database system in the extension

wirednkod:
- Add functionality to extension in order to periodically call `client.databaseContent()` for the well-known-chains (`Polkadot`, `Kusama`, `Rococo`, `Westend`) and save the result in local storage, using the chrome feature `alarms`.
- The alarm will repeat every 5 (`periodInMinutes`) minutes after the initial event for DatabaseContentAlarm
- `databaseContent()` is called with `QUOTA_BYTES: 5000000` divided by the amount of known chains (as of now 4). The maximum amount (in bytes) of data that can be stored in local storage, as measured by the JSON stringification of every value plus every key's length.
- Upon init of extension for each chain the `databaseContent` will be retrieved from localStorage and sent to `addChain`

Fixes #568

### DIFF
--- a/projects/extension/src/background/ConnectionManager.ts
+++ b/projects/extension/src/background/ConnectionManager.ts
@@ -1,9 +1,12 @@
 /* eslint-disable @typescript-eslint/unbound-method */
 /* eslint-disable @typescript-eslint/no-unsafe-assignment */
 /* eslint-disable @typescript-eslint/no-unsafe-call */
+/* eslint-disable @typescript-eslint/no-misused-promises */
+/* eslint-disable @typescript-eslint/no-floating-promises */
 /* eslint-disable @typescript-eslint/no-unsafe-member-access */
 /* eslint-disable @typescript-eslint/no-explicit-any */
 /* eslint-disable @typescript-eslint/no-non-null-assertion */
+/* eslint-disable @typescript-eslint/no-unsafe-argument */
 import {
   Client,
   healthChecker as smHealthChecker,
@@ -166,6 +169,7 @@ export class ConnectionManager extends (EventEmitter as {
     chainSpec: string,
     jsonRpcCallback?: JsonRpcCallback,
     tabId?: number,
+    databaseContent?: string,
   ): Promise<Chain> {
     if (!this.#client) {
       throw new Error("Smoldot client does not exist.")
@@ -177,6 +181,7 @@ export class ConnectionManager extends (EventEmitter as {
 
     return this.#client.addChain({
       chainSpec,
+      databaseContent,
       jsonRpcCallback,
       potentialRelayChains,
     })
--- a/projects/extension/src/background/ConnectionManager.ts
+++ b/projects/extension/src/background/ConnectionManager.ts
@@ -1,12 +1,9 @@
 /* eslint-disable @typescript-eslint/unbound-method */
 /* eslint-disable @typescript-eslint/no-unsafe-assignment */
 /* eslint-disable @typescript-eslint/no-unsafe-call */
-/* eslint-disable @typescript-eslint/no-misused-promises */
-/* eslint-disable @typescript-eslint/no-floating-promises */
 /* eslint-disable @typescript-eslint/no-unsafe-member-access */
 /* eslint-disable @typescript-eslint/no-explicit-any */
 /* eslint-disable @typescript-eslint/no-non-null-assertion */
-/* eslint-disable @typescript-eslint/no-unsafe-argument */
 import {
   Client,
   healthChecker as smHealthChecker,
--- a/projects/extension/src/background/index.ts
+++ b/projects/extension/src/background/index.ts
@@ -76,7 +76,7 @@ const init = async () => {
         dbContent,
       )
       wellKnownConnections.set(key, chain)
-      !dbContent && saveChainDbContent(key, chain)
+      if (!dbContent) saveChainDbContent(key, chain)
     }
 
     chrome.alarms.create("DatabaseContentAlarm", {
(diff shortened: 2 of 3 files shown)


projects/extension/src/background/index.ts
```diff
@@ -33,22 +35,61 @@ declare let window: Background
 window.manager = publicManager
 
 const l = logger("Extension")
+/**
+ * The amount of minutes that the DatabaseContentAlarm
+ * will rerun
+ */
+const periodInMinutes = 5
```
josepot, line 42: There wouldn't be a need for a comment or for a variable definition if we just inlined the value in the configuration Object (which has the same name and it's quite self-descriptive).

projects/extension/src/background/index.ts
```diff
@@ -33,22 +35,61 @@ declare let window: Background
...
+
 export interface RequestRpcSend {
   method: string
   params: unknown[]
 }
 
+const setLocalStorage = async (key: string, chain: Chain) => {
```
josepot, line 49: let's use a less generic name for this function, something like `saveChainDbContent` or similar.

projects/extension/src/background/index.ts
```diff
@@ -33,22 +35,61 @@ declare let window: Background
...
+  )
+  chrome.storage.local.set({ [key]: db })
+}
+
+const flushDatabases = (): void => {
+  for (const [key, chain] of wellKnownConnections) {
+    setLocalStorage(key, chain)
+  }
+}
+
```
josepot, line 61: Can we please have all the code that's related in the same place, i.e:

```ts
const saveChainDbContent = async (key: string, chain: Chain) => {
  const db = await chain.databaseContent(
    chrome.storage.local.QUOTA_BYTES / wellKnownChains.size,
  )
  chrome.storage.local.set({ [key]: db })
}

const flushDatabases = (): void => {
  for (const [key, chain] of wellKnownConnections) {
    saveChainDbContent(key, chain)
  }
}

chrome.alarms.onAlarm.addListener((alarm) => {
  // make sure that the alarm needed is the one set for databaseContent retrieval
  if (alarm.name === "DatabaseContentAlarm") flushDatabases()
})
```

Also, I would inline the `fluhsDatabases` function, but I don't care that much about that.

projects/extension/src/background/index.ts
```diff
@@ -33,22 +35,61 @@ declare let window: Background
...
     manager = new ConnectionManager(start({ maxLogLevel: 3 }))
     for (const [key, value] of wellKnownChains.entries()) {
       const rpcCallback = (rpc: string) => {
         console.warn(`Got RPC from ${key} dummy chain: ${rpc}`)
       }
-      await manager
-        .addChain(value, rpcCallback)
-        .catch((err) => l.error("Error", err))
+
+      const dbContent = await new Promise<string>((res) =>
```
josepot, line 70:
```suggestion
      const dbContent = await new Promise<string | undefined>((res) =>
```

projects/extension/src/background/index.ts
```diff
@@ -33,22 +35,61 @@ declare let window: Background
...
+        chrome.storage.local.get([key], (val) => {
+          return res(val[key] as string)
```
josepot, line 72:
```suggestion
          res(val[key])
```

projects/extension/src/background/ConnectionManager.ts
```diff
@@ -1,9 +1,12 @@
 /* eslint-disable @typescript-eslint/unbound-method */
 /* eslint-disable @typescript-eslint/no-unsafe-assignment */
 /* eslint-disable @typescript-eslint/no-unsafe-call */
+/* eslint-disable @typescript-eslint/no-misused-promises */
+/* eslint-disable @typescript-eslint/no-floating-promises */
 /* eslint-disable @typescript-eslint/no-unsafe-member-access */
 /* eslint-disable @typescript-eslint/no-explicit-any */
 /* eslint-disable @typescript-eslint/no-non-null-assertion */
+/* eslint-disable @typescript-eslint/no-unsafe-argument */
```
josepot, line 9: I don't think that there is any need to add more of this `eslint-disable` comments for the changes in this file.

josepot (commented):
LGTM overall, but I've left a few comments for some minor things.

projects/extension/src/background/index.ts
wirednkod: this change gives an error: `72:22  error  Unsafe argument of type `any` assigned to a parameter of type `string | PromiseLike<string>`  @typescript-eslint/no-unsafe-argument`

projects/extension/src/background/index.ts
```diff
@@ -33,22 +35,53 @@ declare let window: Background
...
+      )
+
+      const chain = await manager.addChain(
+        value,
+        rpcCallback,
+        undefined,
+        dbContent,
+      )
+      wellKnownConnections.set(key, chain)
+      !dbContent && saveChainDbContent(key, chain)
```
tomaka, line 79:
```suggestion
      if (!dbContent) saveChainDbContent(key, chain)
```

I know `&&` looks cool and all, but the goal is to have easy to read code

tomaka (approved):

josepot (approved):
Awesome! Thanks!

PR #682, opened 2022-01-26
add `potentialRelayChainIds` into `add-chain` message

josepot:
This should be the last task of #572 :tada: 

What it does is:
- It removes the `parachainSpec` field, from both `add-chain` and `add-well-known-chain`
- It adds the `potentialRelayChainIds` field into the payload of the `add-chain` message


packages/connect/src/ExtensionProvider/ExtensionProvider.test.ts
```diff
@@ -186,7 +198,7 @@ test("emits error when it receives an error message", async () => {
   expect(error.message).toEqual(errorMessage.payload)
 })
 
-test("it routes incoming messages to the correct Provider", async () => {
+test.skip("it routes incoming messages to the correct Provider", async () => {
```
josepot, line 201: Ups, I forgot to fix this one! I'm on it!

wirednkod (approved):
LGTM. I think there are few tests missing (especially concerning parachain) - but this maybe can be tackled with the zombienet E2E tests

packages/connect-extension-protocol/src/index.ts
```diff
@@ -51,13 +51,13 @@ interface ToExtensionHeader {
...
   chainId: string
-  payload: { chainSpec: string; parachainSpec?: string }
+  payload: { chainSpec: string; potentialRelayChainIds: Array<string> }
 }
 
 interface ToExtensionAddWellKnownChain {
   type: "add-well-known-chain"
   chainId: string
-  payload: { name: string; parachainSpec?: string }
+  payload: string
```
tomaka, line 60:
```suggestion
  chainSpec: string
```

tomaka (approved):

projects/extension/src/background/ConnectionManager.ts
```diff
@@ -172,9 +172,11 @@ export class ConnectionManager extends (EventEmitter as {
...
     }
 
-    const potentialRelayChains = [...this.#chainConnections.values()]
-      .filter((c) => c.tabId === tabId && !c.parachain && c.chain)
-      .map((c) => c.chain!)
+    const potentialRelayChains = potentialRelayChainIds
+      .map(
+        (chainId) => this.#chainConnections.get(`${chainId}::${tabId}`)?.chain,
+      )
+      .filter((chain): chain is Chain => !!chain)
```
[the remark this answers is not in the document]
josepot: https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

packages/connect-extension-protocol/src/index.ts
josepot: nope, that's not correct... It's the name of the "well-known-chain" what's passed in the payload... If you want I can change the payload to `payload: { name: string }`.

There is a reason for putting everything under `payload`, which is that it makes basic message assertions easier to deal with.

PR #694, opened 2022-02-01
Upgrade "@polkadot/api" and "@polkadot/rpc" to latest versions

wirednkod:


projects/burnr/src/components/SendFundsForm.tsx
```diff
@@ -152,7 +152,7 @@ const SendFundsForm: FunctionComponent = () => {
       const fee = await api.tx.balances
         .transfer(address, new BN(amount))
         .paymentInfo(sender)
-      setFee(fee.partialFee)
+      setFee(fee.partialFee as unknown as Balance)
```
tomaka, line 155: Are you sure there's no other solution? I would find it weird that suddenly PolkadotJS breaks their typing

projects/burnr/src/hooks/api/useApi.ts
```diff
@@ -1,4 +1,5 @@
 import { useContext } from "react"
+import "@polkadot/api-augment"
```
tomaka, line 2: What's that and why is it now necessary?

wirednkod: Based on polkadotJS [FAQ](https://polkadot.js.org/docs/api/FAQ/#since-upgrading-to-the-7x-series-typescript-augmentation-is-missing)  this is advised as the solution on the "breakings" that were appearing:
```
Since the start of the 7.x series, type augmentation is to be applied explicitly. To restore Substrate-only types as found earlier versions, an import '@polkadot/api-augment' is to be added to your codebase entry point before any imports from the API itself. The various shipped aliases available are the folowing -
```

projects/burnr/src/components/SendFundsForm.tsx
wirednkod: I haven't figure out any other way. `Balance` is imported for now from `@polkadot/types/interfaces`, which was working fine. I tried also importing from `@polkadot/types/interfaces/runtime` which is the one that `partialFee` corresponds to [in the API](https://github.com/polkadot-js/api/blob/master/packages/types/src/interfaces/payment/types.ts#L5) but the error keeps appearing.

josepot: > I haven't figure out any other way

Removing the `as unkwown as Balance` will suffice. I.e this:

```ts
setFee(fee.partialFee)
```

is fine, b/c `fee`'s type is `RuntimeDispatchInfo` and the `partialFee` property's type is a `Balance` already.

wirednkod:
```
setFee(fee.partialFee)
```
Code was like that - but build is breaking with that, since I updated the polkadotApi

josepot: :thinking: I checkout out your branch, then I ran `git clean -dfx && yarn` (from the root folder), then I changed that line to `setFee(feet.partialFee)` and I then ran `yarn build`... Everything works just fine :shrug:. Could you please make that change and let's see if the CI is ok with it?

wirednkod: Absolutely. Thank you for the insight...

josepot (approved):

PR #715, opened 2022-02-04
Replace `payload` fields with more accurate names

tomaka:
I don't really understand why we're using the name `payload` everywhere, while these payload represent different things depending on the message.

### DIFF
--- a/projects/extension/src/content/ExtensionMessageRouter.ts
+++ b/projects/extension/src/content/ExtensionMessageRouter.ts
@@ -58,7 +58,7 @@ export class ExtensionMessageRouter {
     port.onMessage.addListener((data): void => {
       debug(`RECEIVED MESSAGE FROM ${chainId} PORT`, data)
       sendMessage({
-        ... data,
+        ...data,
         chainId,
         origin: CONTENT_SCRIPT_ORIGIN,
       })
--- a/projects/extension/src/content/ExtensionMessageRouter.ts
+++ b/projects/extension/src/content/ExtensionMessageRouter.ts
@@ -56,11 +56,9 @@ export class ExtensionMessageRouter {
 
     // forward any messages: extension -> page
     port.onMessage.addListener((data): void => {
-      const { type, payload } = data
       debug(`RECEIVED MESSAGE FROM ${chainId} PORT`, data)
       sendMessage({
-        type,
-        jsonRpcMessage: payload,
+        ... data,
         chainId,
         origin: CONTENT_SCRIPT_ORIGIN,
       })
--- a/projects/extension/src/content/ExtensionMessageRouter.ts
+++ b/projects/extension/src/content/ExtensionMessageRouter.ts
@@ -60,7 +60,7 @@ export class ExtensionMessageRouter {
       debug(`RECEIVED MESSAGE FROM ${chainId} PORT`, data)
       sendMessage({
         type,
-        payload,
+        jsonRpcMessage: payload,
         chainId,
         origin: CONTENT_SCRIPT_ORIGIN,
       })
@@ -72,7 +72,7 @@ export class ExtensionMessageRouter {
         origin: "substrate-connect-extension",
         chainId,
         type: "error",
-        payload: "Lost communication with substrate-connect extension",
+        errorMessage: "Lost communication with substrate-connect extension",
       })
       delete this.#ports[chainId]
     })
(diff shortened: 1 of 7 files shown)


josepot (approved):
> I don't really understand why we're using the name `payload` everywhere, while these payload represent different things depending on the message.

I was trying to have messages that resemble the [flux standard action convention](https://github.com/redux-utilities/flux-standard-action#actions), which is a [widely used convention](https://github.com/redux-utilities/flux-standard-action/network/dependencies) that has been adopted by a [significant number of popular libraries](https://github.com/redux-utilities/flux-standard-action/network/dependents?dependent_type=PACKAGE&package_id=UGFja2FnZS01MDY4MDIwNjU%3D). The main reason for doing that was to make the messages feel more intuitive to those JS developers that have already seen these kind of messages before. That being said, I don't care that much, it's a pretty arbitrary convention TBH. Also, we were not following it to the letter... for instance: according to the convention the `chainId` and `origin` properties should go inside a `meta` property.

I mean, an argument could be made that what these payloads represent is already being expressed by the value of the `type` message itself, so there is no need to come up with another name.

projects/extension/src/content/ExtensionMessageRouter.ts
```diff
@@ -60,7 +60,7 @@ export class ExtensionMessageRouter {
       debug(`RECEIVED MESSAGE FROM ${chainId} PORT`, data)
       sendMessage({
         type,
-        payload,
+        jsonRpcMessage: payload,
```
josepot, line 63: There are 2 issues here:
1) Not all the messages received from the background are going to be `rpc` messages
2) there is no such thing as a `payload` anymore, so that will always be `undefined`

josepot (changes requested):

tomaka:
> I was trying to have messages that resemble the [flux standard action convention](https://github.com/redux-utilities/flux-standard-action#actions), which is a [widely used convention](https://github.com/redux-utilities/flux-standard-action/network/dependencies) that has been adopted by a [significant number of popular libraries](https://github.com/redux-utilities/flux-standard-action/network/dependents?dependent_type=PACKAGE&package_id=UGFja2FnZS01MDY4MDIwNjU%3D).

This feels so much like over-engineering to me.
We just want two things to send messages to each other. We're not using this Redux framework, in fact it's not impossible for substrate-connect to outlive Redux. And having some kind of meta standard for us is completely overkill.

josepot:
> We're not using this Redux framework, in fact it's not impossible for substrate-connect to outlive Redux

Just to be clear though: this convention is used outside of Redux and even outside of flux libraries, also Redux itself does not in any way enforce or expects this convention.

> And having some kind of meta standard for us is completely overkill

I concur.

The only thing that I personally liked about always having a `payload` property is that since the value of `type` already describes what's in the `payload` there is no need to come up with another name, and that it's slightly easier to do message-passing/message-validation... However, as I already said before, I'm cool with these changes. In fact, I initially approved the PR until I noticed that there was a bug with the changes.

josepot (approved):

projects/extension/src/content/ExtensionMessageRouter.ts
```diff
@@ -57,11 +57,15 @@ export class ExtensionMessageRouter {
...
+      const msg = {
         ...data,
         chainId,
         origin: CONTENT_SCRIPT_ORIGIN,
-      })
+      }
+
+      // Because the construction of the message is hacky, we have no choice but to  cast
+      // to `unknown`. This is expected to be fixed in a later refactor.
+      sendMessage(msg as unknown as ToApplication)
```
josepot, line 68: A more accurate way of doing this could have been:

```ts
    type HeaderlessToApplication<T extends ToApplication> = T extends {
      origin: "substrate-connect-extension"                            
    } & infer V                                                        
      ? Omit<V, "chainId">                                             
      : unknown                                                        
    port.onMessage.addListener(                                        
      (data: HeaderlessToApplication<ToApplication>): void => {        
        debug(`RECEIVED MESSAGE FROM ${chainId} PORT`, data)           
        sendMessage({                                                  
          ...data,                                                     
          chainId,                                                     
          origin: CONTENT_SCRIPT_ORIGIN,                               
        })                                                             
      },                                                               
    )                                                                  
```

tomaka: I'm changing that code soon

PR #718, opened 2022-02-06
smoldot-connect: initial implementation

josepot:
This weekend I've bee working on the smoldot provider for my side-project (unstoppablejs). I quickly realized that I needed a package that allowed me to optimally use smoldot without having to worry about whether the unstoppablejs provider was talking to the extension or to the WASM module. So, I started working on a library that handles that.

After I made some progress, I realized that other library authors would also benefit from using a library like this (including `@substrate/connect`, which would become much, much simpler if it leveraged this). That's why I'm thinking that this package belongs into this monorepo. For the lack of a better name, I've named it `smoldot-connect`, but I'm more than open to suggestions.

What this library does is:

- it detects whether the extension is available or not
- If the extension is available, it then leverages the `@substrate/connect-extension-protocol` in order to provide the `SmoldotConnect` API, which looks like this:

```ts
export interface Chain {                                
  sendJsonRpc(rpc: string): void                        
  remove(): void                                        
}                                                       
                                                        
export type JsonRpcCallback = (response: string) => void
                                                        
export type AddChain = (                                
  chainSpec: string,                                    
  jsonRpcCallback?: JsonRpcCallback,                    
  potentialRelayChains?: Array<Chain>,                  
) => Promise<Chain>                                     
                                                        
export type AddWellKnownChain = (                       
  name: string,                                         
  jsonRpcCallback?: JsonRpcCallback,                    
) => Promise<Chain>                                     
                                                        
export interface SmoldotConnect {                       
  addChain: AddChain                                    
  addWellKnownChain: AddWellKnownChain                  
}                                                       
```
- if the extension is not available, it then dynamically imports `@substrate/smoldot-light` in order to provide the exact same `SmoldotConnect` API. That also implies optimally loading the "well-known" specs if that's needed.

It's worth noting that the library doesn't "leak" anything from `@substrate/smoldot-light`.

For instance, when the extension is present, the 2 methods of the returned chains also throw the same kind of errors that the chains from smoldot-light throw (`AlreadyDestroyedError`, `CrashError` and `JsonRpcDisabledError`). However, this library has its own Error classes, and the errors thrown are instances of its own classes. One of the reasons for not re-exporting the Error classes from `@substrate/smoldot-light` is that we want to avoid the huge overhead that represents loading the whole `smoldot-light` library just for the sake of importing those Errors. Also, having parts in the public API that come from smoldot-light, would imply defining `smoldot-light` as a peer-dependency, which unnecessarily complicates things.

Similarly, when the extension is not present, the chains  returned by this library are wrappers of the original chains returned by smoldot-light (which are kept internally), with some added features (like ensuring that we catch the errors from smoldot-light and we convert them the the public Errors exported by the library, or running internal logic on removal, etc), and some removed features that (for now) are not part of the API of this library (i.e: `databaseContent`).

I still have to write a good set of tests for this library. So, I'm marking it as a WIP. However, I wanted to open the PR ASAP in order to know whether we see value on this. Otherwise, I will move this library to unstoppablejs... But I really think that would be a shame, b/c the code of `@substrate/connect` would become a lot easier to maintain and to test if it leveraged this library.


wirednkod:
I really like the idea of merging the 2 providers into 1. I have manually tested the code and I did not find any connectivity errors/issues.

Please do not forget to alter in this branch the documentation of `landing-page`

tomaka:
Please mention when you're no longer going to force push on the branch. I don't want to review if there's going to be any more force push.

josepot:
> Please mention when you're no longer going to force push on the branch. I don't want to review if there's going to be any more force push.

I'm very sorry about this. I actually thought that I converted back into a draft after yesterday's conversation... I will do that now and I will convert it as a "ready-for-review" once I'm done editing it.

packages/connect/src/connector/specs/generated/polkadot.ts
```diff
@@ -0,0 +1 @@
+export {}
```
tomaka, line 1: What is this?

packages/connect/src/connector/specs/index.ts
```diff
@@ -0,0 +1,10 @@
+export async function getSpec(chain: string): Promise<string> {
+  const specRaw = (await import("./generated/" + chain + ".js")) as
+    | string
+    | { default: string }
+
+  return typeof specRaw === "string"
+    ? /* istanbul ignore next line */
```
tomaka, line 7: What's istanbul?

packages/connect/src/connector/smoldot-light.ts
```diff
@@ -0,0 +1,119 @@
...
+let totalActiveChains = 0
+let clientPromise: Promise<Client> | null = null
+const getClient = (): Promise<Client> => {
+  if (clientPromise) return clientPromise
+  return (clientPromise = getStart().then((start) =>
+    start({
+      forbidNonLocalWs: true, // Prevents browsers from emitting warnings if smoldot tried to establish non-secure WebSocket connections
+      maxLogLevel: 3 /* no debug/trace messages */,
+    }),
+  ))
```
tomaka, line 50:
```suggestion
  clientPromise = getStart().then((start) =>
    start({
      forbidNonLocalWs: true, // Prevents browsers from emitting warnings if smoldot tried to establish non-secure WebSocket connections
      maxLogLevel: 3 /* no debug/trace messages */,
    }),
  )
  return clientPromise
```

We want to make code as readable as possible

packages/connect/src/connector/smoldot-light.ts
```diff
@@ -0,0 +1,119 @@
...
+  SubstrateConnector,
+} from "./types.js"
+import { SupportedChains } from "../SupportedChains.js"
+
+let SdAlreadyDestroyedError: typeof IAlreadyDestroyedError
+let SdCrashError: typeof ICrashError
+let SdJsonRpcDisabledError: typeof IJsonRpcDisabledError
+
+let startPromise: Promise<(options: ClientOptions) => Client> | null = null
+const getStart = () => {
```
tomaka, line 31: Why are `start` and `startPromise` separate from `clientPromise` and `getClient`?

packages/connect/src/connector/smoldot-light.ts
```diff
@@ -0,0 +1,119 @@
...
+  if (startPromise) return startPromise
+  return (startPromise = import("@substrate/smoldot-light").then((sm) => {
+    SdJsonRpcDisabledError = sm.JsonRpcDisabledError
+    SdCrashError = sm.CrashError
+    SdAlreadyDestroyedError = sm.AlreadyDestroyedError
```
tomaka, line 36: Why is this meta-typing necessary? If it's just to recognize the errors throw by smoldot, can we not use some `name` property or something instead?

packages/connect/src/connector/index.ts
```diff
@@ -0,0 +1,15 @@
...
+export * from "./types.js"
+
+const isExtension =
+  typeof document === "object" &&
+  typeof document.getElementById === "function" &&
+  !!document.getElementById(DOM_ELEMENT_ID)
+
+export const getConnectorClient = isExtension
+  ? eGetConnectorClient
+  : slGetConnectorClient
```
tomaka, line 15: One thing that this approach doesn't handle is that if the extension returns errors, which could happen for example because it has reached the maximum number of chains, then we should probably fall back to the `SmoldotProvider`.

packages/connect/src/connector/extension.ts
```diff
@@ -0,0 +1,133 @@
...
+          origin: "substrate-connect-client",
+          chainId,
+        },
+        "*",
+      )
+    }
+
+    await new Promise<void>((res, rej) => {
+      const stopListeningToExtension = listenToExtension(chainId, (msg) => {
+        stopListeningToExtension()
```
tomaka, line 69: I agree that this pattern when you register and it returns an unregistration function looks cool and all, but the goal is simplicity:

```suggestion
      listeners.set(chainId, (msg) => {
          listeners.delete(chainId);
```

packages/connect/src/connector/extension.ts
```diff
@@ -0,0 +1,133 @@
...
+        if (msg.type === "chain-ready") return res()
```
tomaka, line 70: This doesn't check the `chainId` field?

packages/connect/src/connector/extension.ts
```diff
@@ -0,0 +1,133 @@
...
+        rej(new Error("There was an error creating the smoldot chain."))
+      })
+
+      postToExtension(
+        isWellKnown
+          ? {
+              type: "add-well-known-chain",
+              chainName: chainSpecOrWellKnownName,
```
tomaka, line 78: One thing that needs to be handled is: if the extension fails to initialize the well-known chain, we should fall back to a regular `add-chain` message.

packages/connect/src/connector/extension.ts
```diff
@@ -0,0 +1,133 @@
...
+        stopListeningToExtension()
+        chains.delete(chain)
+        postToExtension({ type: "remove-chain" })
+      }),
+    }
+    chains.set(chain, chainId)
+
+    let crashError: CrashError | null = null
+    const stopListeningToExtension = listenToExtension(chainId, (msg) => {
+      if (msg.type !== "rpc" || !jsonRpcCallback) {
```
tomaka, line 111: I really don't think this is necessary. The web page assumes that the extension works, it's pointless to try to gracefully handle bugs in the extension.
In my opinion `crashError` can be removed.

packages/connect/src/connector/extension.ts
```diff
@@ -0,0 +1,133 @@
...
+      )
+    })
+
+    const chainFn =
+      <Args extends Array<unknown>>(fn: (...args: Args) => void) =>
+      (...args: Args) => {
+        if (crashError) throw crashError
+        if (!chains.has(chain)) throw new AlreadyDestroyedError()
+        fn(...args)
+      }
```
tomaka, line 94: Same remark: this is cool and all, but you're literally saving 2 lines of code. Please just copy-paste these two lines of code instead of using `chainFn`.

tomaka (changes requested):

packages/connect/src/connector/smoldot-light.ts
josepot: > If it's just to recognize the errors throw by smoldot, can we not use some name property or something instead?

I wish... Unfortunately, though, [the custom errors created from `smoldot-connect` don't have custom `name` property](https://github.com/paritytech/smoldot/blob/main/bin/wasm-node/javascript/src/client.ts#L22-L53), so that's impossible. So, I'm afraid that this is the only way of recognizing the errors thrown by smoldot, yes.

packages/connect/src/connector/extension.ts
josepot: sure...

packages/connect/src/SupportedChains.ts
```diff
@@ -0,0 +1,6 @@
+export enum SupportedChains {
```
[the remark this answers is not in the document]
josepot: Makes sense! However, please notice that this was already exposed through the public API with this name. I'll change it, though. Np!

packages/connect/src/connector/specs/index.ts
josepot: I just removed it. It lets the test framework know that line is not to be taken into account for the code-coverage.

packages/connect/src/connector/smoldot-light.ts
josepot: The `startPromise` either gets resolved once (the Extension is not present), or it never gets resolved at all. The `clientPromise` on the other hand will exist as long as there is one active client, once the client gets terminated, then `clientPromise` gets set to `null`, so that if in the future a new client needs to be created it doesn't try to access the terminated one.

packages/connect/src/connector/smoldot-light.ts
tomaka: Yeah but I'm implying that we can just modify smoldot. It's not a walled garden.

packages/connect/src/connector/index.ts
josepot: Yeah, I agree that would be a nice feature that have. However, we currently don't have this behavior and it's a bit tricky both to implement and to get the right API for it. So, I would rather tackle that in the future, once we know how we want to handle that, and how much the consumer should be aware of what's happening...

packages/connect/src/connector/smoldot-light.ts
tomaka: I was thinking that browsers are probably smart enough to not actually import smoldot twice if you run `getStart` twice, but yeah it doesn't hurt to explicitly only ever do it once.

packages/connect/src/connector/extension.ts
josepot: how would that work, though? we would take the chainSpecs from the known ones and try again as a normal `addChain` message? What does that accomplish, though?

packages/connect/src/connector/extension.ts
josepot: I mean, it doesn't hurt either and OTOH it helps TS and eslint be happy so that I don't leave unhandled branches behind... Also, if one day we had an issue with the Extension, this would provide us with more context of what happened.

packages/connect/src/connector/specs/generated/polkadot.ts
josepot: Yeah... So, we currently have this "hacky" script named [`generate-specs`](https://github.com/paritytech/substrate-connect/blob/master/packages/connect/package.json#L28) which what it does is after the `build` has finished it adds a js file with the stringified version of each well known chain-spec into the `dist` folder... So, basically, it's like if we had those files before the build.

What happens, though, is that I have a test that checks that the `addWellKnownChain` API actually imports the chainspec of its corresponding file, in order to do that I have to mock that import, but when I try to mock it `jest` complains saying that the file doesn't exist (because it doesn't, it will be created after build, but that shouldn't be the case). So, in order to work around that I've put an empty file. In the future I'm planing to improve this and make sure that the [`checkSpecs` script](https://github.com/paritytech/substrate-connect/blob/master/packages/connect/package.json#L22) (not to be confused with the `generate-specs`) will add the ts files before build with the stringified content of the well known specs, that way we won't need to have this silly file so that jest doesn't crash.

Alternatively, though, if you want I can disable that test and remove this file until that work is done.

packages/connect/src/connector/extension.ts
tomaka: > we would take the chainSpecs from the known ones and try again as a normal addChain message?

Yes

> What does that accomplish, though?

The idea is to handle situations where the extension removes support for a well-known chain.
Basically, it's as if you asked the extension "do you support chain N?" and if the extension says no you send the chain spec, except that instead of asking the extension if it supports the chain we simply directly ask the extension to initialize the chain, and check if an error is returned.

josepot: It makes sense... I will do that, thanks!

packages/connect/src/connector/specs/generated/polkadot.ts
tomaka: Yeah I'd actually prefer to remove that test.

packages/connect/src/connector/smoldot-light.ts
josepot: Sure! [here you go](https://github.com/paritytech/smoldot/pull/2019/files). Once, `0.5.21` is published I will update the code accordingly, np!

tomaka (approved):
I could nitpick some things, but it can always be changed afterwards

wirednkod (changes requested):
3 points:
- As far as I understand you renamed the `SupportedChains` to `WellKnownChains` - You should replace all comment docs with the new name;
- The Landing Page should be updated with the new changes;
- The version should be updated to 0.6.0 as `SupportedChains` is used from 3rd apps (e.g. polkadotJS apps)

josepot:
> 3 points:
> 
> * As far as I understand you renamed the `SupportedChains` to `WellKnownChains` - You should replace all comment docs with the new name;

Good point! Done!

> * The Landing Page should be updated with the new changes;
> * The version should be updated to 0.6.0 as `SupportedChains` is used from 3rd apps (e.g. polkadotJS apps)

I think that it makes more sense to do these 2 things once I publish the next `0.6.0` release. We already discussed offline that we were going to release all these changes at once, and we still have more changes to make to the public API, so let's wait until that's ready.

wirednkod (approved):

PR #719, opened 2022-02-07
Rename the injected element, and document

tomaka:


packages/connect/src/ScProvider.ts
```diff
@@ -80,7 +80,9 @@ export class ScProvider implements ProviderInterface {
     parachainSpec?: string,
     autoConnect = true,
   ) {
-    const isExtension = !!document.getElementById("substrateExtension")
+    const isExtension = !!document.getElementById(
+      "substrateConnectionExtensionAvailable",
```
wirednkod, line 84: I would suggest to name this as the extension (`substrateConnectExtensionAvailable`)

wirednkod (approved):

packages/connect/src/ScProvider.ts
tomaka: Ah yeah, that was my intention but I'm not fully awake yet apparently

PR #721, opened 2022-02-07
Inline the ExtensionPageInjector

tomaka:
It doesn't really seem necessary to me to have a `ExtensionPageInjector` class.
All it does is contained within 10 lines of code, so I just moved these 10 lines of code to `index.ts`.


projects/extension/src/content/ExtensionPageInjector.ts
```diff
@@ -1,36 +0,0 @@
-import { debug } from "../utils/debug"
-
-const pkg_name = process.env.PKG_NAME as string
-const version = process.env.PKG_VERSION as string
```
tomaka, line 4: I have no idea how that was supposed to work. I couldn't find any `process` module/variable in the Chrome API.

wirednkod: This `pkg_name` and `version` are [set from webpack](https://github.com/paritytech/substrate-connect/blob/master/projects/extension/webpack.config.js#L94-L97).

Having that, the injected element from the extension contains specific version of the installed extension:
```
<span id="substrateExtension" name="@substrate/extension" version="0.0.4"></span>
```

tomaka: I would suggest to remove this

wirednkod (approved):

josepot (approved):

PR #725, opened 2022-02-07
Properly check type of received messages

tomaka:
Adds a new `checkReceivedMessage` function in the content script of the extension that makes sure that messages actually conform to the `ToExtension` interface.

All functions received on the `window` must pass this check.

The messages received on the port between the content script and extension are assumed to always match the interface. Consequently I've removed the check in the `ConnectionManager`.

This might solve issues where a malicious or buggy web page sends invalid messages and generates exceptions in the extension.


josepot (approved):
Nice!

PR #726, opened 2022-02-07
Remove disconnected entries from chainConnection

tomaka:
We close the port but don't remove the entry from `chainConnections`


tomaka:
I've just realized that it's done in `port.onDisconnect`
Will probably refactor that better in the future, but for now let's close this.

PR #727, opened 2022-02-07
Remove `id` from ChainConnection

tomaka:
Replace the `id` field with a `[tabId, chainId]` tuple.

This removes this weird code where we do `${chainId}::${tabId}`.


tomaka:
Ok, this is apparently not possible. Needs more refactoring.

PR #728, opened 2022-02-07
Use only one port per tab

tomaka:
I haven't finished the PR yet, as I haven't tested whether it works and haven't updated the tests.

What this changes is that instead of opening one port between the content script and extension for each chain, we open one only port for the entire tab.

The `ExtensionMessageRouter` only has one port and is generally considerably more simple.

The `ConnectionManager` no longer has a `Map<string, ChainConnection>`, but instead has a `Map<number, Tab>` where `Tab` contains a `Map<string, ChainConnection>`.

This bring benefits when it comes to typing, and will make it possible to continue simplifying the rest of the `ConnectionManager` afterwards.

### DIFF
--- a/projects/extension/src/background/ConnectionManager.ts
+++ b/projects/extension/src/background/ConnectionManager.ts
@@ -89,11 +89,7 @@ export class ConnectionManager extends (EventEmitter as {
   disconnectTab(tabId: number): void {
     const tab = this.#tabs.get(tabId)
     if (!tab) return
-
-    tab.chainConnections.forEach((a) => {
-      tab.port.disconnect()
-    })
-    this.#tabs.delete(tabId)
+    tab.port.disconnect()
   }
 
   /**
@@ -120,8 +116,6 @@ export class ConnectionManager extends (EventEmitter as {
       port.onMessage.addListener(onMessageHandler)
 
       const onDisconnect = () => {
-        this.disconnectTab(tabId)
-
         port.onMessage.removeListener(onMessageHandler)
         port.onDisconnect.removeListener(onDisconnect)
 
@@ -226,6 +220,8 @@ export class ConnectionManager extends (EventEmitter as {
       const rpcResp = chainConnection.healthChecker.responsePassThrough(rpc)
       if (rpcResp)
         chainConnection.port.postMessage({
+          origin: "substrate-connect-extension",
+          chainId,
           type: "rpc",
           jsonRpcMessage: rpcResp,
         })
@@ -266,7 +262,11 @@ export class ConnectionManager extends (EventEmitter as {
     }
 
     chainConnection.chain = chain
-    chainConnection.port.postMessage({ type: "chain-ready", chainId })
+    chainConnection.port.postMessage({ 
+      origin: "substrate-connect-extension",
+      type: "chain-ready",
+      chainId
+    })
 
     // Initialize healthChecker
     chainConnection.healthChecker.setSendJsonRpc((rpc: string) => {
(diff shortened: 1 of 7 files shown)


projects/extension/src/background/ConnectionManager.ts
```diff
@@ -82,59 +87,54 @@ export class ConnectionManager extends (EventEmitter as {
...
+        this.#emitStateChanged()
+        this.#tabs.delete(tabId)
       }
       port.onDisconnect.addListener(onDisconnect)
 
-      this.#chainConnections.set(id, connection)
+      this.#tabs.set(tabId, {
+        chainConnections: new Map(),
+        port,
+      })
```
josepot, line 137: A way to further simplify this code is to define the `tab` as a clojure, i.e:

```ts
      const tabId: number = port.sender!.tab!.id!
      const url: string = port.sender!.url!
      const tab = { port, chainConnections: new Map() }
      this.#tabs.set(tabId, tab)

      const onMessageHandler = (msg: ToExtension) => {
        this.#handleMessage(msg, tab)
      }
      port.onMessage.addListener(onMessageHandler)

      const onDisconnect = () => {
        port.onMessage.removeListener(onMessageHandler)
        port.onDisconnect.removeListener(onDisconnect)

        tab.chainConnections.forEach((chain) => {
          chain.chain?.remove()
        })
        
        this.#emitStateChanged()
        this.#tabs.delete(tabId)
      }
      port.onDisconnect.addListener(onDisconnect)
```

projects/extension/src/background/ConnectionManager.ts
```diff
@@ -193,31 +193,45 @@ export class ConnectionManager extends (EventEmitter as {
...
+    const chainConnection: ChainConnection = {
+      chainId,
+      tabId: tab.port.sender!.tab!.id!,
+      healthChecker: smHealthChecker(),
+      url: tab.port.sender!.url!,
+      port: tab.port,
+      chainName: JSON.parse(chainSpec).name.toLowerCase() || "unknown",
+    }
+
+    // TODO: should check for duplicates? or should the content script do that?
```
josepot, line 214: The content-script doesn't have the means for checking for duplicates, b/c it doesn't have the full picture (an iframe would have a different content-script). So, it should be the background the one checking for duplicates.

Also, what IMO should happen if there was a duplicate is that the background should send an error message for that chainId, so that the existing chainConnection and the one that was possibly trying to impersonate the existing one get removed from the equation.

projects/extension/src/background/ConnectionManager.ts
```diff
@@ -239,13 +253,20 @@ export class ConnectionManager extends (EventEmitter as {
     // for smoldot to return the chain. The way to know this is by checking
     // whether this `chainConnection` is still present insinde `#chainConnections`.
     // If it isn't, that means that the port got disconnected, so we should stop.
-    if (!this.#chainConnections.has(chainConnection.id)) {
+    if (
+      !this.#tabs.has(chainConnection.tabId) ||
+      !this.#tabs.get(chainConnection.tabId)!.chainConnections.has(chainId)
```
josepot, line 258:
```suggestion
      !this.#tabs.get(chainConnection.tabId)?.chainConnections.has(chainId)
```

projects/extension/src/background/ConnectionManager.ts
tomaka: This is not the same thing. I do `this.#tabs.get(tabId)` within callbacks to check whether the tab is still in the list.

josepot: But the thing is that it will always be in the list, it's impossible that it won't be in the list. That's why I think that the code is a bit misleading.

tomaka:
Closing in favour of https://github.com/paritytech/substrate-connect/pull/730

PR #730, opened 2022-02-08
Rewrite the ConnectionManager

tomaka:
This PR does three things:

- Changes the content-script to open only one port and redirect all messages to this port. This considerably simplifies the `ExtensionMessageRouter`

- Rewrites the `ConnectionManager` from scratch. The new `ConnectionManager` now has the concept of sandboxes, which should be understood as ports. However please note that **the ConnectionManager doesn't need to know what a "port" is**.

- Removes `ConnectionManager.test.ts`, because these tests were on the verge of giving me an anxiety attack.


tomaka:
Some details on the exact changes:

- There is now only one port open per content script, and all messages go unaltered (i.e. including the chainId and the header) through it.

- Similarly, the extension immediately creates the final message, including the header and chainId. This is not for technical reasons but because it makes typing easier. Everything is either a `ToApplication` or a `ToExtension` and there's no ambiguity anywhere.

- The `ConnectionManager` no longer holds the `chrome.runtime.Port`. Instead it is parametrized over a `SandboxId` that is set to be the port in `index.ts`.

- Consequently, the port disconnection and messages exchanges are now done in `index.ts`. This de-spaghettifies a bit the code.

- The `ConnectionManager` no longer derives `StateEmitter`, because really there's no reason to. Instead it is `index.ts` that generates automatically calls the listeners.

projects/extension/src/background/ConnectionManager.ts
```diff
@@ -1,297 +1,425 @@
...
+          } else {
+            sandbox.chains.delete(chainId)
+            sendSandbox(sandbox, {
+              origin: "substrate-connect-extension",
+              type: "error",
+              chainId: message.chainId,
+              errorMessage: result,
+            })
+          }
+        })()
```
josepot, line 338: A bit of a nit-pick, but I would much prefer having the error handling logic inside the `catch` clause, rather than the `if/else`s for a result that can have either the error message or the chain.

```suggestion
        ;(async () => {                                                         
          let sandbox: Sandbox | undefined                                      
          try {                                                                 
            const smoldotChain = await chainInitialization.finally(() => {      
              sandbox = this.#sandboxes.get(sandboxId)                          
            })                                                                  
                                                                                
            if (                                                                
              sandbox?.chains.get(chainId)?.smoldotChain !== chainInitialization
            ) {                                                                 
              smoldotChain.remove()                                             
              return                                                            
            }                                                                   
            healthChecker.setSendJsonRpc((rq) => smoldotChain.sendJsonRpc(rq))  
            const readyChain: ReadyChain = {                                    
              isReady: true,                                                    
              name,                                                             
              smoldotChain,                                                     
              healthChecker,                                                    
            }                                                                   
            healthChecker.start((health) => {                                   
              readyChain.latestHealthStatus = health                            
            })                                                                  
            sandbox.chains.set(chainId, readyChain)                             
            sendSandbox(sandbox, {                                              
              origin: "substrate-connect-extension",                            
              type: "chain-ready",                                              
              chainId: message.chainId,                                         
            })                                                                  
          } catch (err) {                                                       
            if (!sandbox) return                                                
            sandbox.chains.delete(chainId)                                      
            sendSandbox(sandbox, {                                              
              origin: "substrate-connect-extension",                            
              type: "error",                                                    
              chainId: message.chainId,                                         
              errorMessage:                                                     
                err instanceof Error                                            
                  ? err.message                                                 
                  : "Unknown error when adding chain",                          
            })                                                                  
          }                                                                     
        })()                                                                    
```

josepot (commented):
Overall I personally like these changes. The code is IMO much cleaner than before.

However, I'm not sure that these changes actually make the code less complex, as most developers that I've worked with really struggle with async iterators, generators and deferred Promises. Although, I have to say that the comments on the code are really helpful, thanks for that!

However, what I think this PR is missing is a good set of tests that show how the `ConnectionManager` is supposed to interact between smoldot and the content-script in each scenario. I don't mind helping with that, if you want I can open a PR against this branch for adding (and fixing) some tests.

projects/extension/src/background/ConnectionManager.ts
tomaka: Just want to point out that it's not the same code. Re-grabbing the sandbox from `this` is intentionally done after the `await`. The comment explains that.

josepot: > Re-grabbing the sandbox from this is intentionally done after the await

So it is in the code that I'm suggesting.

projects/extension/src/background/ConnectionManager.ts
```diff
@@ -1,297 +1,425 @@
...
+    })
+
+    const wellKnownChain: WellKnownChain = { chain, spec, healthChecker }
 
-  #emitStateChanged() {
-    // let's make sure that we don't unnecessary spam the UI with notifications
-    if (this.#emitScheduled) return
+    healthChecker.setSendJsonRpc((rq) => chain.sendJsonRpc(rq))
+    healthChecker.start(
+      (health) => (wellKnownChain.latestHealthStatus = health),
```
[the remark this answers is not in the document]
tomaka: I've noticed that and it's already the case before this PR.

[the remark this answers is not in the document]
josepot: Is it, though? Because I'm looking at the cod of the `ConnectionManager` I can see that [it does emit on health updates](https://github.com/paritytech/substrate-connect/blob/3f585477eba1594b12ff8a33253dfb7f0ac473d3/projects/extension/src/background/ConnectionManager.ts#L261), there is also [a test](https://github.com/paritytech/substrate-connect/blob/3f585477eba1594b12ff8a33253dfb7f0ac473d3/projects/extension/src/background/ConnectionManager.test.ts#L240-L282) to ensure that it does.

Perhaps we have a bug in the UI, if it isn't updating accordingly... I haven't checked yet, but it's possible, since we don't have any tests and there is room for improvement on the UI code, but at least the `ConnectionManager` is doing its part.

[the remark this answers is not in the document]
tomaka: Missed that line

josepot (approved):
I've been trying to add new and better tests for this `ConnectionManager` but I haven't succeeded (yet) and I don't want to further delay this being merged, so I'm approving these changes while hoping that they have been manually tested and that everything works as expected.

projects/extension/src/background/index.ts
```diff
@@ -14,19 +28,44 @@ export interface Background extends Window {
   }
 }
 
-let manager: ConnectionManager
-
-const wellKnownConnections: Map<string, Chain> = new Map()
+let manager: ConnectionManager<chrome.runtime.Port>
+let listeners: Set<(state: ExposedChainConnection[]) => void> = new Set()
```
wirednkod, line 32: This should be a const

projects/extension/src/background/index.ts
```diff
@@ -39,57 +78,67 @@ export interface RequestRpcSend {
...
 
+    waitAllChainsUpdate()
+
     chrome.alarms.create("DatabaseContentAlarm", {
       periodInMinutes: 5,
     })
   } catch (e) {
     l.error(`Error creating smoldot: ${e}`)
-    manager?.shutdown()
+    //manager?.shutdown()
```
wirednkod, line 123: remove unused comment

wirednkod (changes requested):

tomaka:
Will open another PR that cleans up `index.ts`

PR #737, opened 2022-02-09
Transfer health checker from Smoldot to Substrate Connect

### LINKED ISSUE #735: Move HealthChecker from Smoldot-light to Substrate Connect
`HealthChecker` should be moved  to the substrate-connect source code from [smoldot](https://github.com/paritytech/smoldot/blob/main/bin/wasm-node/javascript/src/health.ts). 
The initial idea was to make the `ScProvider` use smoldot's health checker - But if this was implemented that then smoldot would have to be imported unconditionally - so it seems like a better idea to copy/paste the healthChecker to substrate connect.

wirednkod:
Fixes #735


packages/connect/src/ScProvider/Health.ts
```diff
@@ -0,0 +1,319 @@
...
+
+  destroy = (): void => {
+    if (this.#currentHealthTimeout) {
+      clearTimeout(this.#currentHealthTimeout)
+      this.#currentHealthTimeout = null
+    }
+  }
+}
+
+export class HealthCheckError extends Error {
```
tomaka, line 305: I would suggest to completely remove this error (either in this PR or another), because there's no point in gracefully handling bugs in smoldot.

wirednkod: Will address in different PR - if thats ok with you

tomaka (dismissed):

packages/connect/src/ScProvider/Health.ts
```diff
@@ -0,0 +1,319 @@
...
+export interface HealthChecker {
+  setSendJsonRpc(sendRequest: (request: string) => void): void
+  start(healthCallback: (health: SmoldotHealth) => void): void
+  stop(): void
+  sendJsonRpc(request: string): void
+  responsePassThrough(response: string): string | null
+}
+
+interface RequestType {
+  id: string
```
tomaka, line 16: `id` is not necessarily a string

packages/connect/src/ScProvider/Health.ts
```diff
@@ -0,0 +1,319 @@
...
+  jsonrpc: string
+  method: string
+  params: Array<string>
```
tomaka, line 19: `params` doesn't necessarily contain strings

tomaka (changes requested):
What's wrong with leaving types as unknown like that code you copied does?
We don't know what the user sends, it's untrusted input.

packages/connect/src/ScProvider/Health.ts
```diff
@@ -0,0 +1,319 @@
...
+}
+
+interface ResponseType {
+  jsonrpcs: string
```
tomaka, line 23:
```suggestion
  jsonrpc: string
```

wirednkod: 🤦🏽‍♂️

packages/connect/src/ScProvider/Health.ts
```diff
@@ -0,0 +1,303 @@
+/* eslint-disable @typescript-eslint/no-unsafe-assignment */
+/* eslint-disable @typescript-eslint/no-unsafe-member-access */
+/* eslint-disable @typescript-eslint/no-unsafe-argument */
+/* eslint-disable @typescript-eslint/no-unsafe-call */
+export interface SmoldotHealth {
```
tomaka, line 5:
```suggestion

export interface SmoldotHealth {
```

tomaka (approved):

Issue #736, opened 2022-02-09
Clarify/Alter the list of well-known chains


wirednkod:
Clarify the list of well-known chains. Right now we have for example `westend` and `rococo`, but they should actually be called `westend2` and `rococo_v1_14`

tomaka:
Basically their name should be equal to the `id` found in their chain spec.

PR #740, opened 2022-02-10
Fix well-known name in doc

tomaka:
cc https://github.com/paritytech/substrate-connect/issues/736


wirednkod:
This is addressed in #743

wirednkod (approved):

PR #743, opened 2022-02-11
Alter/Clarify names of well known chains

### LINKED ISSUE #736: Clarify/Alter the list of well-known chains

wirednkod:
Fixes #736


wirednkod:
> Note that we're at rococo_v1_14 though

Actually I saw that -> [rococo_v2](https://github.com/paritytech/polkadot/blob/master/node/service/res/rococo.json) :
```
{
  "name": "Rococo 2",
  "id": "rococo_v2",
  "chainType": "Live",
```

Whatever version we are - I will update it in a different PR (dont want to update everything in one)

packages/connect/src/connector/specs/index.ts
```diff
@@ -1,5 +1,13 @@
+const mapNameToId: Map<string, string> = new Map<string, string>([
+  ["polkadot", "polkadot"],
+  ["ksmcc3", "kusama"],
+  ["rococo_v1_13", "rococo"],
+  ["westend2", "westend"],
+])
```
josepot, line 6: Safer:

```suggestion
const mapNameToId: Record<WellKnownChains, string> = {
  [WellKnownChains.polkadot], "polkadot",
  [WellKnownChains.ksmcc3], "kusama",
  [[WellKnownChains.rococo_v1_13] "rococo",
  [WellKnownChains.westend2], "westend",
}
```

This is better and safer for 2 reasons: one the one hand it ensures that all the entries of `WellKnownChains` have an entry inside the Record, on the other when accessing `mapNameToId` we don't have to worry about possible `undefined` values.

packages/connect/src/connector/specs/index.ts
```diff
@@ -1,5 +1,13 @@
...
+
 export async function getSpec(chain: string): Promise<string> {
-  const specRaw = (await import("./generated/" + chain + ".js")) as
+  const fileName = mapNameToId.get(chain) || ""
+  const specRaw = (await import("./generated/" + fileName + ".js")) as
```
josepot, line 10: And then here we can do:

```ts
export async function getSpec(chain: WellKnownChains): Promise<string> {
  const fileName = mapNameToId[chain]
```

which is simpler and more accurate.

josepot (changes requested):
Sorry, 2 more small change requests! :pray:

packages/connect/src/connector/specs/index.ts
wirednkod: I wanted to avoid this as in polkadot repo the same names are used. If you think this is better, then i will proceed (and thus the mapping above will not be needed)

tomaka: We don't have to follow what the Polkadot repo does

projects/landing-page/src/App.tsx
```diff
@@ -212,7 +212,9 @@ const App: React.FunctionComponent = () => {
                 <Box>{`await scClient.addWellKnownChain(WellKnownChains.westend);`}</Box>
                 <Box
                   mt={2}
-                >{`const provider = await scClient.addChain(JSON.stringify(parachainSpecs));`}</Box>
+                >{`const provider = await scClient.addWellKnownChain(WellKnownChains.westend2);`}</Box>
```
josepot, line 215: The change should be on the previous line which is still using the old `WellknownChains.wested`, also with this change the `const provider` gets defined twice... So, this line should be removed and the previous line should be edited.

projects/landing-page/src/App.tsx
```diff
@@ -232,7 +234,8 @@ const App: React.FunctionComponent = () => {
...
```
josepot, line 237: likewise....

josepot (changes requested):

packages/connect/src/connector/specs/index.ts
wirednkod: This is not needed anymore - since the naming of the files have altered

packages/connect/src/connector/specs/index.ts
wirednkod: This is not needed anymore - since the naming of the files have altered

josepot (approved):

PR #745, opened 2022-02-12
@substrate/connect new `createScClient` API

### LINKED ISSUE #742: New "client" based API for accessing the `ProviderInterface`
We've realized that exposing the `ScProvider` class is not the best way to return a polkadotJs `ProviderInterface`, it would be much better to have an API that looks like the API of the `getConnectorClient`.  The interface of the new API would look like this:

```ts
import { WellKnownChains } from "@substrate/connect"
import type { ProviderInterface } from "@polkadot/rpc-provider/types"

declare function createScClient(displayName?: string): {
  addWellKnownChain: (name: WellKnownChains) => Promise<ProviderInterface>,
  addChain: (chainSpec) => Promise<ProviderInterface>
} 
```

A contrived example of its usage:
```ts
async function createProviders(
  myParachainSpec: string,
  myOtherParachainSpec: string,
): Promise<[ProviderInterface, ProviderInterface]> {
  const scClient = createScClient("myContrivedApp")
  await scClient.addWellKnownChain("polkadot")

  return await Promise.all([
    scClient.addChain(myParachainSpec),
    scClient.addChain(myOtherParachainSpec)
  ])
}
```

A few remarks and observations about this new API::
1) The exported `ProviderInterface` will eagerly connect, this is already the default behavior when creating a polkadotJs Provider like the `WsProvider`. However, since constructors can't be `async`, there is no way to detect whether the connection was successful. With the API that we are proposing OTOH it's possible to `catch` the possible errors that can happen while initializing the connection.
2) With this API it will be possible to group the related connections under the same client. Which means that now the user is responsible for creating this connections in the right order. For instance, if they want to connect to a parachain, they first must create the connection to its relayChain.
3) The optional `displayName` may be used for debugging purposes, but it doesn't have any other value apart from that. It's possible that in the future we will improve the `connect-extension-protocol` to accommodate for this field when creating a new chain, so that the connections can be grouped on the UI of the Extension. However, that effort is not part of the scope of this issue.

josepot:
What this PR does is: it closes #742 

What this PR does **not** do is: address some of the issues that we have on the ScProvider (dealing with possible calls to `connect` while smoldot is syncing, using the proper healthChecker, etc). I was tempted to address some of those issues in this PR, and then I realized that was a bad idea. So, I will create separate issues for those things and address those issues in separate PRs.


tomaka (approved):

packages/connect/src/createScClient/ScProvider/ScProvider.ts
```diff
@@ -87,57 +85,23 @@ const CONNECTION_STATE_PINGER_INTERVAL = 2000
...
-  ) {
-    this.#chainSpec = chainSpec
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-    this.#connectionStatePingerId = null
-    if (parachain) {
-      this.#parachainSpecs = parachain
-    }
-    // eslint-disable-next-line @typescript-eslint/no-floating-promises
-    if (autoConnect) this.connect()
+  public constructor(getChain: (handler: JsonRpcCallback) => Promise<Chain>) {
```
tomaka, line 103: Would be great if this was not public

josepot: yeah... I see what you are saying: despite the fact that we are not explicitly making this part of the public API, someone could still import it and use it via

```ts
import { ScProvider } from "@substrate/connect/dist/createScClient/ScProvider"
```

I really want to avoid that, so I will address that before mergin. Thanks for raising this!

josepot: Actually, on a second thought, I will address this later as it's easier to do this as part of the refactor that I'm planning to do on the `ScProvider`.

PR #751, opened 2022-02-14
Fix erroneous imports 

wirednkod:
Prepare for smoldot 0.6.1 update

### DIFF
--- a/packages/connect/src/createScClient/index.ts
+++ b/packages/connect/src/createScClient/index.ts
@@ -1,2 +1,2 @@
 export * from "./createScClient.js"
-export * from "./ScProvider/index.js"
+export * from "./ScProvider/Health.js"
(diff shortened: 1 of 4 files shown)


packages/connect/src/createScClient/index.ts
```diff
@@ -1,2 +1,2 @@
 export * from "./createScClient.js"
-export { healthChecker } from "./ScProvider/index.js"
```
josepot, line 2: We don't want to export `ScProvider`. So, probably the easiest is to leave the `./ScProvider/index.js` untouched, and then from here do: `export * from "./ScProvider/Health.js"

josepot (changes requested):

packages/connect/src/createScClient/index.ts
wirednkod: Thank you 🙏🏽

josepot: No, thank **you**!  I was the one who forgot to export the types of the `healthChecker`.

josepot (approved):

PR #752, opened 2022-02-14
bump smoldot to 0.6.1

wirednkod:


wirednkod:
There is a dependabot that do this

PR #771, opened 2022-02-17
Add mergify

wirednkod:
Added mergify following the example of Smoldot.

Im not sure if we should have autoapproval for dependabot:
```
  - name: automatic approval for Dependabot pull requests
    conditions:
      - author~=^dependabot(|-preview)\[bot\]$
    actions:
      review:
        type: APPROVE
        message: Automatically approving dependabot
```

Opinions @tomaka ?


.github/mergify.yml
```diff
@@ -0,0 +1,22 @@
+queue_rules:
+  - name: default
+    conditions:
+      - check-success=all-ci
```
tomaka, line 4: This "all-ci" is the name of the CI pipeline which has to succeed for the PR to be mergeable

.github/mergify.yml
```diff
@@ -0,0 +1,22 @@
+queue_rules:
+  - name: default
+    conditions:
+      - check-success=build
```
tomaka, line 4: Let's depend on `all`
https://github.com/paritytech/substrate-connect/pull/774

wirednkod: That was my first approach but we have only 1 for now, and when a second is added, then this can be added... But sure

tomaka: We already have three: `build (14.x)`, `build (16.x)` and `build (17.x)`

tomaka (approved):

PR #784, opened 2022-02-20
Revert change to health checker made during transfer

tomaka:
This was modified in https://github.com/paritytech/substrate-connect/pull/737 for no good reason.

The only thing we want to catch is errors in `JSON.parse`. Nothing else. If `requestToSmoldot` throws, we must **not** catch the exception.

### DIFF
--- a/packages/connect/src/createScClient/ScProvider/Health.ts
+++ b/packages/connect/src/createScClient/ScProvider/Health.ts
@@ -110,9 +110,9 @@ class InnerChecker {
 
   sendJsonRpc = (request: string): void => {
     // Replace the `id` in the request to prefix the request ID with `extern:`.
-    let parsedRequest;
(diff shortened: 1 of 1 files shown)


wirednkod:
> This was modified in #737 for no good reason.
> 
> The only thing we want to catch is errors in `JSON.parse`. Nothing else. If `requestToSmoldot` throws, we must **not** catch the exception.

Changes were made to fix lint issues. Tests will not pass unless you disable linter for the `let parsedRequest` or add a `typeof`

tomaka:
> Changes were made to fix lint issues.

Yeah I get it, but please don't change the logic of the code to fix lint issues without a deep understanding of the code. The behavior before and after is not the same.

wirednkod (approved):

PR #785, opened 2022-02-20
Clean up health checker

tomaka:
Better fix than https://github.com/paritytech/substrate-connect/pull/783

### DIFF
--- a/packages/connect/src/createScClient/ScProvider/Health.ts
+++ b/packages/connect/src/createScClient/ScProvider/Health.ts
@@ -230,7 +230,7 @@ class InnerChecker {
           }),
         )
       }
-      
+
       if (startNow) startHealthRequest()
       else this.#currentHealthTimeout = setTimeout(startHealthRequest, 1000)
     }
--- a/packages/connect/src/createScClient/ScProvider/Health.ts
+++ b/packages/connect/src/createScClient/ScProvider/Health.ts
@@ -209,28 +209,30 @@ class InnerChecker {
     }
 
     if (!this.#currentHealthTimeout) {
-      this.#currentHealthTimeout = setTimeout(() => {
-        this.#currentHealthTimeout = null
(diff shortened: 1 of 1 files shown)


josepot (approved):

packages/connect/src/createScClient/ScProvider/Health.ts
```diff
@@ -202,12 +201,38 @@ class InnerChecker {
...
+            JSON.stringify({
+              jsonrpc: "2.0",
+              id: this.#currentHealthCheckId,
+              method: "system_health",
+              params: [],
+            }),
+          )
+        },
+        startNow ? 0 : 10000,
+      )
```
josepot, line 235: There is a small change in behavior with these changes that's making some tests to fail. What's changed is that the `startNow` request used to start synchronously. However, with these changes the request is scheduled in a new macrotask.

Either we update the tests or we make the `startNow` request to start synchronously. I suggest keeping the original behavior:

```suggestion
      const startHealthRequest = () => {
        this.#currentHealthTimeout = null

        // No matter what, don't start a health request if there is already one in progress.
        // This is sane to do because receiving a response to a health request calls `update()`.
        if (this.#currentHealthCheckId) return

        // Actual request starting.
        this.#currentHealthCheckId = "health-checker:".concat(
          this.#nextRequestId.toString(),
        )
        this.#nextRequestId += 1
        this.#requestToSmoldot(
          JSON.stringify({
            jsonrpc: "2.0",
            id: this.#currentHealthCheckId,
            method: "system_health",
            params: [],
          }),
        )
      }
      
      if (startNow) startHealthRequest()
      else this.#currentHealthTimeout = setTimeout(startHealthRequest, 1000)
```

josepot: I kinda need these changes... So, I will commit the suggestion. I hope that's ok.

PR #787, opened 2022-02-21
improvement and fixes on the polkadot RPC Provider

### LINKED ISSUE #747: Improvements and fixes on the @polkadot RPC Provider
The current Provider that's being exported from the `screateScClient` API needs a bit of love:

- Despite the fact that the `ScProvider` is not explicitly exposed from the public API, it's still possible to import it via `import { ScProvider } from "@substrate/connect/dist/createScClient/ScProvider"`. We should prevent that from happening.
- The Provider should use the  `healthChecker` that we recently borrowed from `smoldot-light`
- The Provider should distinguish between real/hard disconnections and the disconnections that happen when smoldot is syncing or it has no peers
- Also, the Provider should gracefully handle the fact that the consumer can call the public `connect` method after  having received a `disconnect` event.

josepot:
It closes #747


packages/connect/src/createScClient/createScClient.ts
```diff
@@ -11,19 +20,293 @@ export interface ScClient {
...
+
+      let decodedResponse: string | Error
+      try {
+        decodedResponse = this.#coder.decodeResponse(response) as string
+      } catch (e) {
+        decodedResponse = e as Error
+      }
+
+      // It's not a subscription message, but rather a standar RPC response
+      if (response.method === undefined) {
```
tomaka, line 94: I think you should check this with `response.params.subscription`, not with  ̀response.method`

packages/connect/src/createScClient/createScClient.ts
```diff
@@ -11,19 +20,293 @@ export interface ScClient {
...
+
+      const callback = this.#subscriptions.get(subscriptionId)?.[0]
+      if (callback) return callback(decodedResponse)
+
+      // It's possible to receive subscriptions messages before having received
+      // the id of the subscription. In that case We should keep these
+      // messages around until we receive the subscription-id message.
+      if (!this.#orphanMessages.has(subscriptionId))
+        this.#orphanMessages.set(subscriptionId, [])
+      this.#orphanMessages.get(subscriptionId)!.push(decodedResponse)
```
tomaka, line 109: Smoldot will always send back the id before the first messages.
You're reaching this because you're using callbacks in the `subscribe` methods, meaning that the response containing the "id" is processed later despite having arrived before.

packages/connect/src/createScClient/createScClient.ts
```diff
@@ -11,19 +20,293 @@ export interface ScClient {
...
+        } else killStaleSubscriptions()
+
+        this.#isChainReady = isReady
+        this.#eventemitter.emit(isReady ? "connected" : "disconnected")
+      })
+
+      return {
+        ...chain,
+        sendJsonRpc: hc.sendJsonRpc.bind(hc),
+        remove: () => {
```
tomaka, line 172: It's really confusing to also name this `remove`. I have no idea which `remove` you're calling where.

packages/connect/src/createScClient/createScClient.ts
```diff
@@ -11,19 +20,293 @@ export interface ScClient {
...
+
+    this.#chain = this.#getChain(onResponse).then((chain) => {
+      hc.setSendJsonRpc(chain.sendJsonRpc)
+
+      this.#isChainReady = false
+      const cleanup = () => {
+        // If there are any callbacks left, we have to reject/error them.
+        // Otherwise, that would cause a memory leak.
+        const disconnectionError = new Error("Disconnected")
+        this.#requests.forEach((cb) => cb(disconnectionError))
```
tomaka, line 120: You must do `this.#requests.clear()`
The reason why we track requests is so that if PolkadotJS disconnects then reconnects we don't want it to receive responses to requests that it has sent before the disconnection.

tomaka (changes requested):

packages/connect/src/createScClient/createScClient.ts
josepot: > You're reaching this because you're using callbacks in the subscribe methods, meaning that the response containing the "id" is processed later despite having arrived before.

Nah... The reason why I'm doing this is because this is what happens over the wire when using the WsProvider... I thought that smoldot was going to do the same thing.

josepot: > You're reaching this because you're using callbacks in the subscribe methods, meaning that the response containing the "id" is processed later despite having arrived before.

Nah... The reason why I'm doing this is because this is what happens over the wire when using the WsProvider... I thought that smoldot was going to do the same thing.

josepot: I guess that we should keep this code for now, at least until [this bug](https://github.com/paritytech/substrate/issues/10910) is fixed.

packages/connect/src/createScClient/createScClient.ts
josepot: I think it's the same thing, but I don't mind changing it...

packages/connect/src/createScClient/createScClient.ts
tomaka: Why? The code in this PR will never interact with Substrate

packages/connect/src/createScClient/createScClient.ts
josepot: It's a way to ensure that whenever the consumer of the exposed `chain` calls `remove` and/or `sendJsonRpc` that the right logic takes place... So, we ware making the consumer unaware of the "magic" that goes behind enhancing the `Chain`.

In other words: this code is mostly making up for the convoluted API of the `healthChecker`. You see, the `healthChecker` is in fact a leaky "Chain enhancer". If the `healthChecker` had had a cleaner API, like this  one:

```ts
type HealthHandler = (healthUpdate: SmoldotHealth) => void              
type HealthChain = Chain & {                                            
  onHealthUpdate: HealthHandler                                         
  getCurrentHealth: () => SmoldotHealth                                 
}                                                                       
type GetHealthChain = (handler: JsonRpcCallback) => Promise<HealthChain>
                                                                        
type GetChain = (handler: JsonRpcCallback) => Promise<Chain>            
                                                                        
type HealthChecker = (getChain: GetChain) => GetHealthChain             
```

or like this one:

```ts
type HealthHandler = (healthUpdate: SmoldotHealth) => void  
type GetHealthChain = (                                     
  handler: JsonRpcCallback,                                 
  healthHandler: HealthHandler,                             
) => Promise<Chain>                                         
                                                            
type GetChain = (handler: JsonRpcCallback) => Promise<Chain>
                                                            
type HealthChecker = (getChain: GetChain) => GetHealthChain 
```

Then I wouldn't have to "hack" the exposed chain that much.

That being said: I'm open to suggestions. Is there a concrete action that you want me to take here?

packages/connect/src/createScClient/createScClient.ts
josepot: Oh! My bad... So, what you are saying is that the smoldot  chain doesn't have this issue? That's awesome! Well, then in that case I will update the PR...

josepot: [Done](https://github.com/paritytech/substrate-connect/pull/787/commits/f6bd57eaddcf8cb95834abd0b83229acc5e74145)!

FTR those changes and their tests should be a pretty convincing prove that it's never been the case that the code of this PR was:

> reaching this because you're using callbacks in the subscribe methods, meaning that the response containing the "id" is processed later despite having arrived before

tomaka (approved):

wirednkod (approved):

PR #788, opened 2022-02-21
improve eslint configuration

josepot:
Our current eslint configuration is quite painful and not very useful,  mostly because it has tons of stylistic rules that don't really correct mistakes. A great eslint default is IMO the `eslint-config-react-app` which has a curated list of eslint rules that are very useful (not only for react projects!).

So, I've removed all the other rules/plugins in favor of this one, then I have removed all the `eslint-disable` comments and I've also addressed a few things that the new eslint config has found. Although, after doing all those things, the new eslint config only displayed a few (very useful IMO) warnings, which I addressed in this PR.


tomaka (approved):
I don't know anything about eslint

wirednkod (changes requested):
Why are `tests` removed in this PR? (projects/extension/src/utils/debug.test.ts, projects/extension/src/utils/utils.test.ts). I know they are nothing "special" but I don't see the reason to be removed.

projects/burnr/src/components/PopoverInfo.tsx
```diff
@@ -43,13 +43,13 @@ const PopoverInfo: FunctionComponent<Props> = ({ children }: Props) => {
 
   return (
     <>
-      <a
+      <button
```
wirednkod, line 46: is there a reason why <a> was converted to <button>?

projects/burnr/src/themes/substrate/colors.ts
```diff
@@ -1,4 +1,4 @@
-export default {
+const colors = {
```
wirednkod, line 1: Is this necessary (compared to the way it was before)?

wirednkod (changes requested):

projects/burnr/src/components/PopoverInfo.tsx
josepot: yes, accessibility. An `a` tag must have a valid `href` attribute, otherwise we should use a button. There was an eslint-rule that caught this. Thankfully as a warning, not an error, but since I noticed that we have custom styles I thought that it was better to make the change.

projects/burnr/src/themes/substrate/colors.ts
josepot: it's not necessary, bu it's a lot better for debugging reasons. There is now an eslint rule that raises a warning when creating anonymous default exports... And I really like that rule, b/c it really helps with debugging.

josepot:
> Why are `tests` removed in this PR? (projects/extension/src/utils/debug.test.ts, projects/extension/src/utils/utils.test.ts). I know they are nothing "special" but I don't see the reason to be removed.

Well, the new eslint rules were showing a bunch of warnings on those tests and when I looked at them I thought that those tests were pretty much useless because they are only testing a bunch of internals, so I thought it was better to remove them than to maintain them... I can bring them back, but really, they are quite useless.

PR #794, opened 2022-02-22
add 'automerge' label for dependabot PRs

wirednkod:


tomaka (approved):
So the idea is that you just need to approve a PR from dependabot for it to be merged

PR #802, opened 2022-02-24
fix rococo icon

wirednkod:


tomaka:
Can you explain what's happening? We download icons from somewhere?

wirednkod:
> Can you explain what's happening? We download icons from somewhere?

In extension we are using instead of icons (pngs or SVGs etc) a font created by @goldsteinsveta ([demo](https://goldsteinsveta.github.io/web3-regular/)). With this font, instead of importing an icon, you only need to set the font on the element, and then by using the "name" (e.g. polkadot, kusama etc) the respective logo will appear.

Font is contained in the repo (no npm dependency or downloading anything from anywhere fwiw)

tomaka (approved):
I still don't understand, but as long as it works

PR #804, opened 2022-02-25
replace onInstalled and onStartup with plain init()

wirednkod:


projects/extension/src/background/index.ts
```diff
@@ -111,6 +103,8 @@ chrome.storage.local.get(["notifications"], (result) => {
   }
 })
 
+init()
+
 // TODO (nik): once extension is on chrome/ff stores we need to take advantage
 // of the onBrowserUpdateAvailable and onUpdateAvailable lifecycle event
 // NOTE: onSuspend could be used to cleanup things but async actions are not guaranteed to complete :(
```
tomaka, line 110:
```suggestion
```

tomaka (approved):

PR #807, opened 2022-02-28
Add package-level documentation

tomaka:


packages/connect/src/index.ts
```diff
@@ -1,3 +1,101 @@
+/**
+ * The substrate-connect package makes it possible to connect to Substrate-compatible blockchains.
+ *
+ * Connecting to a chain is done in two steps:
+ *
+ * - Call {getConnectorClient} or {createPolkadotJsScClient}, which gives you a so-called *client*.
+ * - Call `addChain` or `addWellKnownChain` on this client.
```
wirednkod, line 7:
```suggestion
 * 1 - Call {getConnectorClient} or {createPolkadotJsScClient}, which gives you a so-called *client*.
 * 2 - Call `addChain` or `addWellKnownChain` on returned client.
```

packages/connect/src/index.ts
```diff
@@ -1,3 +1,101 @@
...
+ * Failing to do so will lead to an error at the initialization of the parachain.
+ *
+ * Furthermore, the parachain must be added to the same client object as the one the relay chain
+ * was added to.
+ *
+ * In other words, this will work:
+ *
+ * ```js
+ * const client = getConnectorClient();
+ * await client.addChain(relayChain);
```
wirednkod, line 43: Should there be mentioned in all these examples the `JSON.stringify(....)` or its irrelevant for current example-show-case?

packages/connect/src/index.ts
```diff
@@ -1,3 +1,101 @@
...
+ *
+ * This means that there is no problem in calling {getConnectorClient} or
+ * {createPolkadotJsScClient} from within a library for example.
+ *
+ * # Well-known chains
+ *
+ * This package contains a list of so-called {WellKnownChains}. This is a list of popular chains
+ * that users are likely to connect to. Instead of calling `addChain` with a chain specification,
+ * one can call `addWellKnownChain`, passing only the name of a well-known chain as parameter.
+ *
```
wirednkod, line 80:
```suggestion
 * At the moment these are polkadot, kusama, rococo and westend.
 * See <https://github.com/paritytech/substrate-connect/blob/main/packages/connect/src/WellKnownChains.ts>.
```

packages/connect/src/index.ts
tomaka: The point of this example is to illustrate what the text says, not to demonstrate how to use the library

packages/connect/src/index.ts
tomaka: Normally the JS docs generator should link to `WellKnownChains`, there's no reason to put a link.

wirednkod (approved):

PR #809, opened 2022-02-28
Fix race condition in smoldot-light

tomaka:
Before this PR, you can do this:

```js
const client = getConnectorClient();
const chain1 = await client.addChain(...);
const chain2Promise = client.addChain(...);
chain1.remove();
await chain2Promise;
```

And calling `chain1.remove()` will call `terminate()` on the underlying smoldot client, which will lead to `await chain2Promise` throwing an `AlreadyDestroyedError` exception.


josepot (approved):
Thanks for this! I've improved the tests a bit in the light of this bug, please feel free to further improve them or to alter them.

PR #814, opened 2022-02-28
Refactor the extension's index.ts

tomaka:
This can be reviewed commit by commit.

This PR properly deals with the fact that the initialization of the manager is asynchronous.

This PR notably fixes a race condition: it was theoretically possible for a content script to open a port and send a `addWellKnownChain` message before all the well-known chains have been added to the manager (thus leading to an error).
Note that it was also theoretically possible (but unlikely) for a content script to open a port even while the `manager` is still `null`.


josepot (approved):

wirednkod (approved):
Other than annoying eslint warnings this lgtm

PR #821, opened 2022-03-01
Bump versions

wirednkod:


projects/extension/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@substrate/extension",
-  "version": "0.1.0",
+  "version": "0.2.0",
```
tomaka, line 3: This should be `0.1.1`
We didn't do any breaking change to the API of the extension

tomaka (approved):

PR #822, opened 2022-03-01
Fix missing dependency in package.json

tomaka:


wirednkod (approved):

josepot (commented):
This is something that's also being addressed on #823. I think that the easiest is to close this PR, but no big deal if it gets merged.

PR #825, opened 2022-03-01
Don't always open port between content script and extension

tomaka:
https://github.com/paritytech/substrate-connect/pull/730 changed something: the port between the content script and the extension opens for every single web page.
This isn't really a problem, I think, but I think it's better to only open it on web pages that actually try to use the extension, instead of all the time.


josepot (approved):
Nice!

wirednkod (approved):

tomaka:
Didn't realize this wasn't merged

PR #836, opened 2022-03-03
@substrate/connect@0.6.1 fixes usage on node

josepot:
I made a mistake when I wrote the `createScClient`: its index file is importing both `createScClient` functions (the one from the extension and the one from its smoldot-light counterpart)  and then it exposes one or the other depending on whether the extension is installed.

However, those functions are not idempotent and they rely on some caches and listeners that are instantiated just before the function definitions. One of those being the global `window.addEventListener('message, ...` used from the `createScClient` function of the extension... Which means that despite the efforts that were made on the index file not to access `document.getElementId` on node, `@substrate/connect` doesn't work on node because when importing the `createScClient` function from the extension we are eagerly attempting to create a listener on `window` :facepalm:.

This PR addresses this issue by exporting a `getCreateScClient` function from both sides, so that we only instantiate the relevant one.


packages/connect/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@substrate/connect",
-  "version": "0.6.0",
+  "version": "0.6.1",
```
wirednkod, line 3: You should not update the version. we will do that once ready to release

wirednkod (approved):

tomaka:
Why not just add `if (window)` line 24 in `extension.ts`, and throw an error if `createScClient` is called and `window` is undefined?

josepot:
> Why not just add `if (window)` line 24 in `extension.ts`, and throw an error if `createScClient` is called and `window` is undefined?

`if (window)` would throw on node. Although, I suppose that you meant `if (typeof window === "object")`. It's true that would work, and the diff would also be easier to review :sweat_smile:. Yeah, ok, I will do it like that... I don't know, in a way it felt wrong to instantiate a bunch of stuff that wasn't going to be used (caches, variables for Promises, etc)... However, well, it's not being used so it doesn't really matter :smile:

tomaka (approved):

PR #842, opened 2022-03-06
fix: stale unsubscriptions

josepot:
I made a mistake when I tried to tackle the issue with stale subscriptions that remain on the smoldot chain after the chain has changed its status to "syncing". This PR addresses the issue. The list of supported subscription methods was taken from [here](https://github.com/paritytech/smoldot/blob/17425040ddda47d539556eeaf62b88c4240d1d42/src/json_rpc/methods.rs#L338-L462). While reviewing, please ensure that I didn't miss any methods :pray:.

### DIFF
--- a/packages/connect/src/createPolkadotJsScClient/createPolkadotJsScClient.ts
+++ b/packages/connect/src/createPolkadotJsScClient/createPolkadotJsScClient.ts
@@ -22,12 +22,22 @@ export interface PolkadotJsScClient {
 
 type ResponseCallback = (response: string | Error) => void
 
+const subscriptionUnsubscritionMethods = new Map<string, string>([
+  ["author_submitAndWatchExtrinsic", "author_unwatchExtrinsic"],
+  ["chain_subscribeAllHeads", "chain_unsubscribeAllHeads"],
+  ["chain_subscribeFinalizedHeads", "chain_unsubscribeFinalizedHeads"],
+  ["chain_subscribeNewHeads", "chain_unsubscribeNewHeads"],
+  ["state_subscribeRuntimeVersion", "state_unsubscribeRuntimeVersion"],
+  ["state_subscribeStorage", "state_unsubscribeStorage"],
+  ["transaction_unstable_submitAndWatch", "transaction_unstable_unwatch"],
+])
+
 class Provider implements ProviderInterface {
   readonly #coder: RpcCoder = new RpcCoder()
   readonly #getChain: (handler: JsonRpcCallback) => Promise<Chain>
   readonly #subscriptions: Map<
     string,
-    [ResponseCallback, { method: string; id: string | number }]
+    [ResponseCallback, { unsubscribeMethod: string; id: string | number }]
   > = new Map()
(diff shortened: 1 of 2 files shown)


packages/connect/src/createPolkadotJsScClient/createPolkadotJsScClient.ts
```diff
@@ -22,12 +22,22 @@ export interface PolkadotJsScClient {
...
 type ResponseCallback = (response: string | Error) => void
 
+const subscriptionUnsubscritionMethods = new Map<string, string>([
+  ["author_submitAndWatchExtrinsic", "author_unwatchExtrinsic"],
+  ["chain_subscribeAllHeads", "chain_unsubscribeAllHeads"],
+  ["chain_subscribeFinalizedHeads", "chain_unsubscribeFinalizedHeads"],
+  ["chain_subscribeNewHeads", "chain_unsubscribeNewHeads"],
+  ["state_subscribeRuntimeVersion", "state_unsubscribeRuntimeVersion"],
+  ["state_subscribeStorage", "state_unsubscribeStorage"],
+  ["transaction_unstable_submitAndWatch", "transaction_unstable_unwatch"],
```
tomaka, line 32: All the methods of the old API are there, however `transaction_unstable_submitAndWatch` belongs to the new API.
We should either add all methods of both the old and new API (keeping in mind that the new API might change), or just the old API. I'd go for just the old API at the moment, leaving an explication as comment.

packages/connect/src/createPolkadotJsScClient/createPolkadotJsScClient.ts
```diff
@@ -22,12 +22,22 @@ export interface PolkadotJsScClient {
 
 type ResponseCallback = (response: string | Error) => void
 
+const subscriptionUnsubscritionMethods = new Map<string, string>([
```
tomaka, line 25:
```suggestion
const subscriptionUnsubscriptionMethods = new Map<string, string>([
```

josepot: wow, good eye! thanks!

packages/connect/src/createPolkadotJsScClient/createPolkadotJsScClient.ts
```diff
@@ -22,12 +22,26 @@ export interface PolkadotJsScClient {
 
 type ResponseCallback = (response: string | Error) => void
 
+// These methods have been taken from:
+// https://github.com/paritytech/smoldot/blob/17425040ddda47d539556eeaf62b88c4240d1d42/src/json_rpc/methods.rs#L338-L462
+// It's important to take into account that smoldot is adding support to the new
+// json-rpc-interface https://paritytech.github.io/json-rpc-interface-spec/
+// However, these list should only include methods that belong to the "old" API
```
tomaka, line 29:
```suggestion
// However, at the moment this list only includes methods that belong to the "old" API
```

It's not a final decision

tomaka (approved):

PR #846, opened 2022-03-07
Introduce Changelog files for connect, connect-extension-protocol and…

wirednkod:
… extension

### DIFF
--- a/packages/connect/CHANGELOG.md
+++ b/packages/connect/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
-### Added
+# Fixed
 
-- Changelog
+- Add missing subscription/unsubscription methods ([#850](https://github.com/paritytech/substrate-connect/pull/850/))
+- Wrong unsubscription JSON-RPC methods being used when calling `ProviderInterface.disconnect()` ([#842](https://github.com/paritytech/substrate-connect/pull/842)).
--- a/packages/connect/CHANGELOG.md
+++ b/packages/connect/CHANGELOG.md
@@ -4,5 +4,4 @@
 
 # Fixed
 
-- Add missing subscription/unsubscription methods ([#850](https://github.com/paritytech/substrate-connect/pull/850/))
-- Wrong unsubscription JSON-RPC methods being used when calling `ProviderInterface.disconnect()` ([#842](https://github.com/paritytech/substrate-connect/pull/842)).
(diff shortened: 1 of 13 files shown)


tomaka:
This needs to be merged right after a new version is published, or contain the list of unreleased changes

wirednkod:
> This needs to be merged right after a new version is published, or contain the list of unreleased changes

When next PR is merged (e.g. logs) then we need to add this to the changelog (under unreleased). Thats why I think its a good idea to merge it prior to any other merge - so that files are there for updating

tomaka:
I also don't think it's worth saying that we've added a changelog. It's pretty clear when you see that the changelog starts that there was no changelog before

tomaka:
> When next PR is merged (e.g. logs) then we need to add this to the changelog (under unreleased). Thats why I think its a good idea to merge it prior to any other merge - so that files are there for updating

What I mean is that you need to add a line about https://github.com/paritytech/substrate-connect/pull/842/

packages/connect/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
-### Added
+# Fixed
 
-- Changelog
+- Add missing subscription/unsubscription methods ([#850](https://github.com/paritytech/substrate-connect/pull/850/))
+- Wrong unsubscription JSON-RPC methods being used when calling `ProviderInterface.disconnect()` ([#842](https://github.com/paritytech/substrate-connect/pull/842)).
```
tomaka, line 8: There's no need for it to be two lines
Users don't care that we fixed a bug introduced in between two versions

```suggestion
- Fixed wrong unsubscription JSON-RPC methods being used when calling `ProviderInterface.disconnect()` ([#842](https://github.com/paritytech/substrate-connect/pull/842)).
```

projects/extension/CHANGELOG.md
```diff
@@ -2,6 +2,6 @@
 
 ## [Unreleased]
 
-### Added
+# Removed
 
-- Changelog
+- Usage of the `SmoldotHealth` interface ([#845](https://github.com/paritytech/substrate-connect/pull/845))
```
tomaka, line 7: This is an internal change, users don't care about that

```suggestion
```

projects/extension/CHANGELOG.md
```diff
@@ -0,0 +1,6 @@
+# Changelog
+
+## [Unreleased]
+
+# Removed
+
```
tomaka, line 6:
```suggestion
```

tomaka (approved):

PR #847, opened 2022-03-07
Fix typo error of chain_subscribeNewHeads

wirednkod:
[This](https://github.com/paritytech/substrate-connect/runs/5453459596?check_suite_focus=true#step:7:150) revealed that the `chain_subscribeNewHeads` was typo and needed to be renamed to `chain_subscribeNewHead`


tomaka:
Both functions are actually valid.

[In the list of methods](https://github.com/paritytech/smoldot/blob/17425040ddda47d539556eeaf62b88c4240d1d42/src/json_rpc/methods.rs#L338-L405), you can find between [] the alternative names.

Issue #429, opened 2021-08-02
Fix  extension logging strategy


wirednkod:
Our logging strategy is going a bit all over the place in the extension.
This should be organised bit better.

raoulmillais:
Also all our debug logs are ALL CAPS. I don't know why I started doing that. It's a bit over the top! 😲

raoulmillais:
From @tomaka in chat:

Quick remark/brainstorm: when it comes to logging on the extension side, I think it'd be a good idea to keep in memory the last 500 or so lines of logs level 4 (the debug logs) and add some sort of semi-hidden button in the UI to show them
If we assume that things are working just fine, there's no point in showing logs to the end user (i.e. the person having the extension installed in their browser) anyway. So the only utility for logs is debugging in case something bad happens.

tomaka:
What in my opinion we should do in terms of logging in the extension:

- Use debug level 4 in smoldot.
- Don't output the logs in the console, but instead keep the logs in some buffer. Only the last ~500 or so lines should to be kept, otherwise our memory usage would continue increasing forever.
- Add a semi-hidden way somewhere in the extension that can display these logs. The average Joe doesn't need to know that these logs exist, it's more for developers to be able to understand problems.
- (maybe) If the ~500 log lines that we keep contain a warning or error, there should be some indication. Again, the average Joe doesn't need to see this, but for developers it's useful to be able to easily see if a warning or error has happened. Warnings or errors are generally bugs in smoldot that should be fixed, but don't necessarily prevent applications from working normally.
- If smoldot crashes: extract the log lines, and internally restart smoldot and reset the logs. Show a big indication in the UI (this time, visible for the average Joe) indicating that something terribly wrong has happened and that they should open an issue in smoldot. This should contain the logs that have been extracted.

cc @Goranch3

wirednkod:
- [ x ] Use log level 4 in smoldot.

- [ x ] Don't output the logs in the console, but instead keep the logs in some buffer. Only the last ~500 or so lines should to be kept, otherwise our memory usage would continue increasing forever.

[...]

PR #853, opened 2022-03-08
Extension logging

wirednkod:
This PR covers all-but-one (the last) bullets of Pierre's [comment](https://github.com/paritytech/substrate-connect/issues/429#issuecomment-1059754226) in issue #429 .

Some details

> Don't output the logs in the console, but instead keep the logs in some buffer. Only the last ~500 or so lines should to be kept, otherwise our memory usage would continue increasing forever.

I left the `info` logs to still appear in the console. 1000 lines are kept in memory and shown in the extension's Option page.

> Add a semi-hidden way somewhere in the extension that can display these logs. The average Joe doesn't need to know that these logs exist, it's more for developers to be able to understand problems.

The `semi-hidden` way for now is another tab in the Options page. This will be adjusted in the next UI upgrade from @Goranch3 

> (maybe) If the ~500 log lines that we keep contain a warning or error, there should be some indication. Again, the average Joe doesn't need to see this, but for developers it's useful to be able to easily see if a warning or error has happened. Warnings or errors are generally bugs in smoldot that should be fixed, but don't necessarily prevent applications from working normally.

2 placeholders are keeping separate from rotating debug logs, the errors and warnings. Even though only 1000 latest lines are  are kept from debug logs and then older ones are dissappearing, warning and errors are kept.

Last bullet of Pierre's comment will be addressed in different PR.

A screenshot on the UI perspective can be seen below:
image

### DIFF
--- a/projects/extension/CHANGELOG.md
+++ b/projects/extension/CHANGELOG.md
@@ -1,0 +1,3 @@
+# Changelog
+
+## [Unreleased]
--- a/projects/extension/CHANGELOG.md
+++ b/projects/extension/CHANGELOG.md
@@ -4,4 +4,4 @@
 
 ## Added
 
-- Add extension logging strategy (#429](https://github.com/paritytech/substrate-connect/issues/429))
+- Add a new tab in the options showing the logs of the underlying client (#429](https://github.com/paritytech/substrate-connect/issues/429))
--- a/projects/extension/CHANGELOG.md
+++ b/projects/extension/CHANGELOG.md
@@ -1,3 +1,7 @@
 # Changelog
 
 ## [Unreleased]
+
+## Added
+
+- Add extension logging strategy (#429](https://github.com/paritytech/substrate-connect/issues/429))
--- a/projects/extension/CHANGELOG.md
+++ b/projects/extension/CHANGELOG.md
@@ -2,6 +2,6 @@
 
 ## [Unreleased]
 
-## Added
+### Added
 
 - Add a new tab in the options showing the logs of the underlying client (#429](https://github.com/paritytech/substrate-connect/issues/429))
--- a/projects/extension/src/background/index.ts
+++ b/projects/extension/src/background/index.ts
@@ -24,7 +24,44 @@ export interface Background extends Window {
       listener: (state: ExposedChainConnection[]) => void,
     ) => () => void
     disconnectTab: (tabId: number) => void
+    getLogger: () => LogKeeper
   }
+}
+
+interface LogKeeper {
+  debug: string[]
+  warn: string[]
+  error: string[]
+}
+
+const logKeeper: LogKeeper = {
+  debug: [],
+  warn: [],
+  error: [],
+}
+
+const getTime = () => {
+  const date = new Date()
+  return `${date.getHours() < 10 ? "0" : ""}${date.getHours()}:${
+    date.getMinutes() < 10 ? "0" : ""
+  }${date.getMinutes()}:${
+    date.getSeconds() < 10 ? "0" : ""
+  }${date.getSeconds()} ${date.getMilliseconds()}`
+}
+
+const logger = (level: number, target: string, message: string) => {
+  const logLevels = ["Error", "Warn", "Info", "Debug"]
+  if (level !== 4) {
+    // log all non-debug logs to background console
+    console.log(`${logLevels[level - 1]}: ${message}`)
+    if (level == 1) {
+      logKeeper.error.push(`${getTime()} | ${target} | ${message}`)
+    } else if (level === 2) {
+      logKeeper.warn.push(`${getTime()} | ${target} | ${message}`)
+    }
+  }
+  logKeeper.debug.length >= 1000 && logKeeper.debug.shift()
+  logKeeper.debug.push(`${getTime()} | ${target} | ${message}`)
 }
 
 const listeners: Set<(state: ExposedChainConnection[]) => void> = new Set()
@@ -72,6 +109,7 @@ window.manager = {
       }
     })
   },
+  getLogger: () => logKeeper,
 }
 
 const saveChainDbContent = async (
@@ -91,7 +129,10 @@ const saveChainDbContent = async (
 const managerPromise: Promise<ConnectionManager<chrome.runtime.Port>> =
   (async () => {
     const managerInit = new ConnectionManager<chrome.runtime.Port>(
-      smoldotStart(),
+      smoldotStart({
+        maxLogLevel: 4,
+        logCallback: logger,
+      }),
     )
     for (const [key, value] of wellKnownChains.entries()) {
       const dbContent = await new Promise<string | undefined>((res) =>
--- a/projects/extension/src/background/index.ts
+++ b/projects/extension/src/background/index.ts
@@ -28,7 +28,7 @@ export interface Background extends Window {
   }
 }
 
-type logStructure = {
+interface logStructure {
   time: string
   level: string
   target: string
@@ -56,11 +56,6 @@ const getTime = () => {
   }${date.getSeconds()} ${date.getMilliseconds()}`
 }
 
-const maintainArr = (arr: logStructure[], incLog: logStructure): void => {
-  if (arr.length >= 1000) arr.shift()
-  arr.push(incLog)
-}
-
 // duplicate "Error" is intentional for the extreme case of level === 0
 const logLevels = ["Error", "Error", "Warn", "Info", "Debug"]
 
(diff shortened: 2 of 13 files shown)


tomaka:
Logs only make sense if they're in the correct context.
You shouldn't split debug, info, warn and error logs in different windows, they should all keep the order they came in

wirednkod:
> Logs only make sense if they're in the correct context.
> Please don't split debug, info, warn and error logs in different windows, they should all keep the order they came in

In debug lis all errors are kept there (i should fix probably the naming. The logs and warns are kept separate on top in order to be able to "filter" them faster. In addition errors and warns do not disappear once the 1000 lines pass. This way even if not identified when it happened, dev can now that something happened, see the warn/error log even out of context at a later time

projects/extension/src/background/index.ts
```diff
@@ -24,9 +24,46 @@ export interface Background extends Window {
       listener: (state: ExposedChainConnection[]) => void,
     ) => () => void
     disconnectTab: (tabId: number) => void
+    getLogger: () => LogKeeper
   }
 }
 
+interface LogKeeper {
+  debug: string[]
```
tomaka, line 32:
```suggestion
  all: string[]
```

Since this also contains non-debug

projects/extension/src/background/index.ts
```diff
@@ -24,9 +24,46 @@ export interface Background extends Window {
...
+  }${date.getMinutes()}:${
+    date.getSeconds() < 10 ? "0" : ""
+  }${date.getSeconds()} ${date.getMilliseconds()}`
+}
+
+const logger = (level: number, target: string, message: string) => {
+  const logLevels = ["Error", "Warn", "Info", "Debug"]
+  if (level !== 4) {
+    // log all non-debug logs to background console
+    console.log(`${logLevels[level - 1]}: ${message}`)
```
tomaka, line 56: I know it's a bit annoying, but I'd prefer something like

```suggestion
    switch (level) {
        case 0: case 1: console.error(...); break;
        case 2: console.warn(...); break;
        case 3: console.info(...); break;
        case 4: console.log(...); break;
    }
```

So that we keep the proper level

projects/extension/src/background/index.ts
```diff
@@ -24,9 +24,46 @@ export interface Background extends Window {
...
+    if (level === 1) {
+      logKeeper.error.push(`${getTime()} | ${target} | ${message}`)
+    } else if (level === 2) {
+      logKeeper.warn.push(`${getTime()} | ${target} | ${message}`)
+    }
+  }
+  logKeeper.debug.length >= 1000 && logKeeper.debug.shift()
```
tomaka, line 63:
```suggestion
  if (logKeeper.debug.length >= 1000) logKeeper.debug.shift()
```

tomaka (changes requested):
Mostly good!

projects/extension/src/background/index.ts
wirednkod: Is there a `case 0:`

projects/extension/src/background/index.ts
tomaka: Also, I think it's a good idea to do the same for warnings and errors, just in case

projects/extension/src/background/index.ts
tomaka: No normally there can't be `0`, but in case there is it's better to not just discard it

projects/extension/src/containers/Options.tsx
```diff
@@ -181,6 +227,146 @@ const Options: React.FunctionComponent = () => {
...
+                badgeContent={warnLogs.length}
+                showZero
+                color="primary"
+              ></Badge>
+            </Typography>
+          </AccordionSummary>
+          <AccordionDetails>
+            <div style={{ maxHeight: "500px", overflow: "auto" }}>
+              {warnLogs.map((d: string) => {
+                const res = d.split("|")
```
tomaka, line 300: Also, I don't think this is a good idea. What if the `target` or the message contains a `|`?
I'd prefer if we just kept the date, log level, target, and message as individual fields in the `LogKeeper`, instead of concatenating them and splitting them later

projects/extension/src/background/index.ts
wirednkod: fwiw the suggestion is not entirely correct. I mean that case 4 will not be logged (as we do not want `debug` to appear in the browser console. I will alter accordingly

projects/extension/src/containers/Options.tsx
```diff
@@ -37,6 +37,13 @@ interface StyledTabProps {
   label: string
 }
 
+type logStructure = {
+  time: string
+  level: string
```
tomaka, line 42: If you don't use this field, then don't include it here

projects/extension/src/background/index.ts
```diff
@@ -49,19 +56,40 @@ const getTime = () => {
...
     // log all non-debug logs to background console
-    console.log(`${logLevels[level - 1]}: ${message}`)
-    if (level === 1) {
-      logKeeper.error.push(`${getTime()} | ${target} | ${message}`)
-    } else if (level === 2) {
-      logKeeper.warn.push(`${getTime()} | ${target} | ${message}`)
+    switch (level) {
+      case 0:
+      case 1:
+        maintainArr(error, incLog)
```
tomaka, line 80:
```suggestion
  if (error.length >= 1000) error.shift()
  error.push(incLog)
```

Please just copy-paste this, there's no need for a separate function

projects/extension/src/background/index.ts
```diff
@@ -49,19 +56,40 @@ const getTime = () => {
...
+}
+
+// duplicate "Error" is intentional for the extreme case of level === 0
+const logLevels = ["Error", "Error", "Warn", "Info", "Debug"]
+
 const logger = (level: number, target: string, message: string) => {
-  const logLevels = ["Error", "Warn", "Info", "Debug"]
+  const incLog = {
+    time: getTime(),
+    level: logLevels[level],
```
tomaka, line 70:
```suggestion
    level,
```

projects/extension/src/background/index.ts
```diff
@@ -28,14 +28,21 @@ export interface Background extends Window {
   }
 }
 
+type logStructure = {
+  time: string
+  level: string
```
tomaka, line 33:
```suggestion
  level: number
```

projects/extension/src/background/index.ts
```diff
@@ -28,14 +28,21 @@ export interface Background extends Window {
   }
 }
 
+type logStructure = {
```
tomaka, line 31: Why not an interface?

```suggestion
interface LogStructure = {
```

projects/extension/src/containers/Options.tsx
wirednkod: I am supposed to use it (all logs) - erroneously (copy/paste) left out. Thank you

projects/extension/src/background/index.ts
wirednkod: level is a number. I prefer to keep the description of the error on the logs

projects/extension/src/background/index.ts
wirednkod: I use type on a structure when its used inside an interface - but in latest versions of TS type and interface tend to be the same so it does not make much of difference

projects/extension/src/background/index.ts
tomaka: But it's not the job of `index.ts` to decide which words the user is going to see in the UI

projects/extension/src/background/index.ts
wirednkod: level is a number. I prefer to keep the description of the error on the logs

projects/extension/src/background/index.ts
wirednkod: Yeah - took me few mins to "get" it - thus I removed my comment. Thank you 🙏🏽

projects/extension/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
 # Changelog
 
 ## [Unreleased]
+
+## Added
+
+- Add extension logging strategy (#429](https://github.com/paritytech/substrate-connect/issues/429))
```
tomaka, line 7:
```suggestion
- Add a new tab showing the logs of the underlying client (#429](https://github.com/paritytech/substrate-connect/issues/429))
```

"Add extension logging strategy" doesn't mean anything from the point of view of a user

projects/extension/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
 # Changelog
 
 ## [Unreleased]
+
+## Added
```
tomaka, line 5:
```suggestion
### Added
```

tomaka (approved):

PR #860, opened 2022-03-10
Add extra features in logs (copy to clipboard) and remove accordions

wirednkod:


projects/extension/src/background/index.ts
```diff
@@ -29,7 +29,7 @@ export interface Background extends Window {
 }
 
 interface logStructure {
-  time: string
+  time: number
```
tomaka, line 32:
```suggestion
  unix_timestamp: number
```

tomaka (approved):

PR #861, opened 2022-03-11
Remove waitAllChainChanged and return an event instead

### LINKED ISSUE #857: Remove ConnectionManager.waitAllChainChanged and return an event from sandboxOutput instead

tomaka:
Fix #857 

This PR removes the `waitAllChainChanged()` functions from `ConnectionManager` and `ConnectionManagerWithHealth`.
This function was used to notify that the list of chains or their status has changed in some way.
Users of the `ConnectionManager` or `ConnectionManagerWithHealth` are now supposed to do that by themselves by looking at the messages that are being exchanged.

As for the changes in health, `ConnectionManagerWithHealth.nextSandboxMessage` can now return a `ChainsStatusChanged` that notifies that one chain has had its status modified. This extra message could be expanded upon in the future.

This simplifies the behavior of `ConnectionManager` and `ConnectionManagerWithHealth` by removing asynchronicity and ambiguities.


projects/extension/src/background/ConnectionManagerWithHealth.ts
```diff
@@ -328,7 +333,11 @@ export class ConnectionManagerWithHealth<SandboxId> {
                   })
                   this.#nextHealthCheckRqId += 1
 
-                  break
+                  // Notify of the change in status.
+                  return {
+                    origin: "connection-manager",
+                    type: "chains-status-changed",
+                  }
```
wirednkod, line 340: Re-use object here:
```
return statusChanged

File: projects/extension/src/background/ConnectionManagerWithHealth.ts [hunk 2]
```diff
@@ -289,9 +296,11 @@ export class ConnectionManagerWithHealth<SandboxId> {
...
               chain.peers = result.peers
 
-              // Notify the `allChainsChangedCallbacks`.
-              this.#allChainsChangedCallbacks.forEach((cb) => cb())
-              this.#allChainsChangedCallbacks = []
+              // Notify of the change in status.
+              return {
+                origin: "connection-manager",
+                type: "chains-status-changed",
+              }
```
wirednkod, line 303: Add a const object with the interface and reuse it instead of copy/pasting the object:

 ```tsx
              const statusChanged: ChainsStatusChanged = {
                origin: "connection-manager",
                type: "chains-status-changed",
              }
              
              ...............
              
              return statusChanged
```

wirednkod (approved):

projects/extension/src/background/ConnectionManagerWithHealth.ts
tomaka: I don't see the advantage of doing that?

wirednkod: Reduce copy/pasting

tomaka: But copy-pasting code is not a problem by itself.
It can cause problems in some situations, because you can modify the copy-pasted code at one place and not at another, but this doesn't apply here because TypeScript would detect it.

PR #862, opened 2022-03-11
Remove projects/extension/utils

tomaka:
I initially wanted to just remove `debug.ts` because it's weird and unnecessary, but then I noticed that `contexts.ts` was unused, and the two functions in `utils.ts` are both only used once.


wirednkod (changes requested):
Please do not remove these functions. They may be needed in the new UI implementation. Once UI is finished we can cleanup the code and remove non-repeatable or not-needed functions. New UI is the reason why I am postponing the cleanup of extension so far.

tomaka:
But you can always add them back. They're literally one line of code, and we have a git history. It's for the same reason as you don't leave commented out code in the code by remove it.

wirednkod (approved):
> But you can always add them back. They're literally one line of code, and we have a git history. It's for the same reason as you don't leave commented out code in the code by remove it.

We can - or just leave it there 🤷🏽 . Doesn't really matter

PR #863, opened 2022-03-11
Always disable TCP connections in smoldot-light

tomaka:
This is an opinionated decision.

Right now it is possible for users to run substrate-connect on NodeJS, think that everything works, then run the same script in the browser and realize that it can't connect to their chain.

I think that in the future we should add a configuration option specific to `smoldot-light` that allows TCP connections, but that by default they should always be disabled.


tomaka:
Similarly, there should be an option to "emulate" the extension by having `forbidTcp: true` and `forbidWs: true`.

wirednkod (approved):

tomaka:
Just realized this need a changelog entry.

PR #864, opened 2022-03-14
Document in landing page that best block cannot be trusted

### LINKED ISSUE #535: Document in the landing page that the best block can't be trusted
One important property of substrate-connect, as opposed to the `WsProvider`, is that the best block can't be trusted. It might be invalid. Only the finalized block is guaranteed to be valid.

This is really important if you're creating an app on top of substrate-connect, to not give the impression to the user of the app that something reflects the reality while in fact it might be completely wrong.

Since it's really important, in my opinion it makes sense to put this info in a section of the landing page.

wirednkod:
Fixes #535


packages/connect/src/index.ts
```diff
@@ -23,6 +23,16 @@
...
  * PolkadotJS. Proper light-client-oriented high-level libraries built on top of
  * {createScClient} are currently in development.
+ * 
+ * # Security concern
+ *
+ * As an important note - a light client only verifies the authenticity of blocks, but not their
+ * correctness. For this reason, the block that is reported as being the best block might be
+ * incorrect. In the context of a light client, only the finalised block can be assumed
+ * to be correct. Consequently, accessing the storage of any non-finalised block is also
+ * not guaranteed to report correct values. This security concerni is mostly relevant
```
raoulmillais, line 33:
```suggestion
 * not guaranteed to report correct values. This security concern is mostly relevant
```

packages/connect/src/index.ts
```diff
@@ -23,6 +23,16 @@
  * for demos and prototypes, it is not possible to achieve the best user experience by using
  * PolkadotJS. Proper light-client-oriented high-level libraries built on top of
  * {createScClient} are currently in development.
+ * 
+ * # Security concern
+ *
+ * As an important note - a light client only verifies the authenticity of blocks, but not their
```
raoulmillais, line 29:
```suggestion
 * It is important to note that a light client only verifies the authenticity of blocks and not their
```

packages/connect/src/index.ts
```diff
@@ -23,6 +23,16 @@
...
+ * correctness. For this reason, the block that is reported as being the best block might be
+ * incorrect. In the context of a light client, only the finalised block can be assumed
```
raoulmillais, line 31:
```suggestion
 * incorrect. For application built with a light client, only finalised blocks should be assumed
```

tomaka:
In https://github.com/paritytech/substrate-connect/issues/535 I explicitly request this change, but when seeing it I realize that it's very out of place. We give just a brief overview, and then in the middle there's suddenly a warning about a very precise detail.

I think that this warning would belong where we explain that you can use substrate-connect though PolkadotJS, because this untrusted best block problem comes from the old JSON-RPC API and thus relates to PolkadotJS. But we don't really explain that you can use PolkadotJS anywhere on the landing page.

wirednkod:
Will replace with a new PR with polkadotJS reference

PR #866, opened 2022-03-14
CI - extension versions check

### LINKED ISSUE #827: Add a CI script that verifies that the extension version in package.json is always the same as in manifest.json

wirednkod:
Fixes #827


tomaka:
Could you add this to `yarn.CI.yml` instead, rename `yarn.CI.yml` to something else, and add the job to the `needs` thing of the `all` job at the end?
There's IMO no good reason why this would be a different workflow than the rest of the CI checks.

.github/workflows/ci.yml
```diff
@@ -22,3 +22,28 @@ jobs:
       if: ${{ steps.package-version.outputs.version != steps.webpack-version.outputs.version }}
     - run: exit 1
       if: ${{ steps.package-version.outputs.version != steps.manifest-version.outputs.version }}
+  build:
```
tomaka, line 25:
```suggestion

  build:
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,49 @@
...
+    branches: [ main ]
+
+jobs:
+  check-extension-versions:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+    - id: package-version
+      run: echo "::set-output name=version::`grep -rw './projects/extension/package.json' -e 'version' | sed 's/[^0-9.]//g'`"
```
tomaka, line 16: Is it possible to use `jq` instead maybe? It feels a bit cleaner to me.
Something like `grep package.json | jq .version` should work, I believe?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,49 @@
...
+    - id: webpack-version
+      run: echo "::set-output name=version::`grep -rnw './projects/extension/webpack.config.js' -e 'PKG_VERSION' | grep -o '".*"' | tr -d '"'`"
+    - id: manifest-version
+      run: echo "::set-output name=version::`grep -rw './projects/extension/public/manifest.json' -e 'version' | sed 's/[^0-9.]//g'`"
```
tomaka, line 20: Same remark about `jq`

.github/workflows/ci.yml
```diff
@@ -0,0 +1,49 @@
...
+    - run: exit 1
+      if: ${{ steps.package-version.outputs.version != steps.webpack-version.outputs.version }}
```
tomaka, line 22: Maybe add a third check where we make sure that `version` is not `""`, so that we detect if the CI breaks.

wirednkod: I do not think this is needed - as for this to happen one needs to remove versions in all 3 files and I assume that will be noticed in the PR

tomaka (approved):

PR #868, opened 2022-03-16
Make the ConnectionManager robust to smoldot exceptions

tomaka:
This modifies the connection manager to properly absorb exceptions coming from the smoldot library.

If smoldot crashes, the connection manager kills all active chains and sets an internal `hasCrashed` boolean to true.
When the `index.ts` file sees that errors are being sent to the frontend, it checks this internal boolean. The rest isn't implemented as it is out of scope of this PR.


tomaka:
Note that the way it is implemented means that crashes are not necessarily detected instantly. However smoldot also doesn't expose any function in its API that makes it possible to detect crashes instantly.
I'm not sure if it's worth the efforts for things that aren't supposed to happen.

wirednkod (approved):
lgtm but these `catch (error) {}` really grind my gears

PR #872, opened 2022-03-16
ignore json chain files when publishing to npm

### LINKED ISSUE #871: Chain specs are found twice in the connect package
If you go in `package/connect` and do `npm publish --dry-run` (yarn doesn't have any equivalent :face_exhaling: ) you can see that we upload the four chain specs twice.

wirednkod:
Fixes #871

### DIFF
--- a/packages/connect/.npmignore
+++ b/packages/connect/.npmignore
@@ -7,5 +7,5 @@ yarn-error.log*
 tsconfig.*
 extension-resolver.cjs
 jest.config*
-
 src/**/*
+dist/connector/specs/*.json


tomaka:
Is there no way to, say, not generate the chain specs at all instead?
Not publishing is a receipt for potential disaster because when testing everything will work well but once published there might be files missing.

wirednkod:
> Is there no way to, say, not generate the chain specs at all instead? Not publishing is a receipt for potential disaster because when testing everything will work well but once published there might be files missing.

The generated specs are the [only one used in the code](https://github.com/paritytech/substrate-connect/blob/main/packages/connect/src/connector/extension.ts#L103) and they are [generated](https://github.com/paritytech/substrate-connect/blob/main/packages/connect/package.json#L28) during build. the `json` that are not published are not actually used from code afaik. Im not sure why `json` files that are not used can lead the `npmjs` package to disaster.

tomaka:
> Im not sure why json files that are not used

Because if you start using them, accidentally or not, the CI will not detect it. You will only notice the problem after it's been published.

wirednkod:
> if you accidentally start using them the CI or local testing will not detect it

Understood. As said before the generation was a performance improvement. We can either roll back to using the json files instead of the generated ones (which will bring back the performance case), maybe we could remove these JSON files in order to avoid accidental usage and download them, use and remove them during build, or maybe document not to use these files.

PR #873, opened 2022-03-16
Improve well-known chain spec import

tomaka:
Changes two things:

- We forbid accessing any file outside of the `generated` directory.
- The error message is better if the `import` throws.


packages/connect/src/connector/specs/index.ts
```diff
@@ -1,9 +1,20 @@
 export async function getSpec(chain: string): Promise<string> {
-  const specRaw = (await import("./generated/" + chain + ".js")) as
-    | string
-    | { default: string }
+  // We don't want API users to be able to `import` a file outside of the `generated` directory.
+  // While it is probably harmless, better be safe than sorry.
+  // This is done by make sure that the name doesn't contain `..`. This also means that we can't
+  // support well-known chain whose name contains `..`, but that seems unlikely to ever be
+  // problematic.
+  if (chain.search("..") !== 1) throw new Error("Invalid chain name")
```
wirednkod, line 7:
```suggestion
  if (chain.indexOf("..") !== -1) throw new Error("Invalid chain name")
```

wirednkod (approved):

PR #875, opened 2022-03-16
Fix breaking 'peers of undefined' error

wirednkod:


tomaka:
Wait, no, `chain` is never supposed to be null/undefined, adding `?` just silences a bug

wirednkod:
> Wait, no, `chain` is never supposed to be null/undefined, adding `?` just silences a bug

The "silencing" is actually receiving `undefined` as peers and `isSynching = false`, which I think it is fine (and its not crashing the extension). Is it preferable for the extension to break?

PR #877, opened 2022-03-16
Fix burnr for next updates of @polkadot dependencies

wirednkod:
This PR will fix #848 , #865 and #874


wirednkod:
Replaced by #880

PR #881, opened 2022-03-18
Fix race condition between lists of chains in managers

tomaka:
It turns out that we need to store in `ConnectionManagerWithHealth` the list of chains that are still initializing, otherwise there might be inconsistencies with the list in `ConnectionManager`.


tomaka:
This is https://github.com/paritytech/substrate-connect/pull/875 but fixed properly

wirednkod (approved):

PR #900, opened 2022-03-25
Limit smoldot's CPU usage to 50%

tomaka:
This change is there really to avoid a worst case scenario where smoldot has an issue that causes it to never go to sleep.

This will very slightly slow down syncing (verifying the warp sync proof is a pure CPU operation), but avoiding big CPU spikes is more desirable than saving a few milliseconds.


projects/extension/CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 ### Changed
 
 - Update @substrate/smoldot-light to [version 0.6.8](https://github.com/paritytech/smoldot/blob/main/bin/wasm-node/CHANGELOG.md#068---2022-03-23) ([#890](https://github.com/paritytech/substrate-connect/pull/890))
+- The smoldot background worker will now bound its CPU usage to 50% of one CPU on average. ([#900](https://github.com/paritytech/substrate-connect/pull/900))
```
wirednkod, line 13: Im not sure if this is something that end user should know. Its extension-internal - right?

wirednkod (approved):

projects/extension/CHANGELOG.md
tomaka: It's visible to the user that the usage of their CPU is bounded to 50%

wirednkod:
There is something print when zombienet tests are running other than `uncaught exception` - but it is happening after the 140 seconds of the timeout (timeout comes first - prior to log so script exits).
When spawn locally zombienet the actual error print can be found below:
```
Timeout duration was set to 1.
(node:27030) TimeoutOverflowWarning: 19327215118 does not fit into a 32-bit signed integer.
Timeout duration was set to 1.
(node:27030) TimeoutOverflowWarning: 19327215118 does not fit into a 32-bit signed integer.
Timeout duration was set to 1.
(node:27030) TimeoutOverflowWarning: 19327215118 does not fit into a 32-bit signed integer.
Timeout duration was set to 1.
(node:27030) TimeoutOverflowWarning: 19327215117 does not fit into a 32-bit signed integer.
Timeout duration was set to 1.
(node:27030) TimeoutOverflowWarning: 19327215117 does not fit into a 32-bit signed integer.
Timeout duration was set to 1.
(node:27030) TimeoutOverflowWarning: 19327215117 does not fit into a 32-bit signed integer.
Timeout duration was set to 1.
(node:27030) TimeoutOverflowWarning: 19327215117 does not fit into a 32-bit signed integer.
Timeout duration was set to 1.
(node:27030) TimeoutOverflowWarning: 19327215117 does not fit into a 32-bit signed integer.
Timeout duration was set to 1.
(node:27030) TimeoutOverflowWarning: 19327215117 does not fit into a 32-bit signed integer.
Timeout duration was set to 1.
uncaughtException
Error: Timeout running custom js-script (140)
```
This sample is from a param of 0.75 (I think) passed to the `cpuRateLimit` param. Similar results were identified with `0.5`, `0.75`, `0.85`, `0.9` - though with param set to `1` everything works as expected.

With my none-existent rust skills I can only assume that the issue comes either in [here](https://github.com/paritytech/smoldot/blob/main/bin/wasm-node/rust/src/cpu_rate_limiter.rs#L46-L53) or [here](https://github.com/paritytech/smoldot/blob/main/bin/wasm-node/rust/src/cpu_rate_limiter.rs#L85-L92) but again my rust is very bad to say for sure.

tomaka:
https://github.com/paritytech/smoldot/pull/2188

tomaka:
https://github.com/paritytech/smoldot/pull/2189

tomaka:
Despite the bugs in smoldot, it would still be great to figure out why this `uncaughtException` isn't catched, and properly catch it

wirednkod:
> Despite the bugs in smoldot, it would still be great to figure out why this `uncaughtException` isn't caught, and properly catch it

Went through zombienet code with @pepoviola . The `uncaughtException` is actually a `console.log` as can be seen [here](https://github.com/paritytech/zombienet/blob/main/src/cli.ts#L27) and the "error" was the timeout occurred:
```
Error: Timeout running custom js-script (140)
```

pepoviola:
> > Despite the bugs in smoldot, it would still be great to figure out why this `uncaughtException` isn't caught, and properly catch it
> 
> Went through zombienet code with @pepoviola . The `uncaughtException` is actually a `console.log` as can be seen [here](https://github.com/paritytech/zombienet/blob/main/src/cli.ts#L27) and the "error" was the timeout occurred:
> 
> ```
> Error: Timeout running custom js-script (140)
> ```

Yes, this will be fixed in the next version of zombienet.
Thanks!

PR #915, opened 2022-03-28
fix issue with `chain.remove()`

josepot:


wirednkod (approved):

tomaka:
Please explain what happens and how the PR fixes it? Because I don't understand

packages/connect/src/connector/smoldot-light.test.ts
```diff
@@ -32,19 +32,24 @@ class SdJsonRpcDisabledError extends Error {
 
 const mockSmoldotLightFactory = () => {
   const start = (options: ClientOptions) => {
-    const addChain = (addChainOptions: AddChainOptions) => {
+    const addChain = (
+      addChainOptions: AddChainOptions,
+      isClientTerminated: () => boolean,
```
tomaka, line 37: We pass `isClientTerminated()` to `addChain`?! Why is it not part of the mock to detect when `terminate` has been called?

tomaka:
This is not the correct fix, you're absorbing all errors that `addChain` throws, even the legitimate ones

tomaka:
I don't understand what the bug even is. Did you add a failing test and then fix the test failure, or was this problem actually noticed somewhere in the wild? Because to me everything should work fine: the client is terminated if and only if `clientNumReferences` is set to 0, and since `clientPromise` is set to null we don't keep terminated clients around.

packages/connect/src/connector/smoldot-light.test.ts
josepot: It *is* part of the mock to detect whether `terminate` has been called. This is precisely the point of passing the callback to the mocked chain, so that it can "ask" the mocked client if it has been terminated. It's a pretty common inversion of control technique... I'm sure that there must be a 1000 other ways of accomplishing the same thing, but this does its job, as demoed by the failing tests that it triggered... If you have a better suggestion, though, I'm all ears.

PR #909, opened 2022-03-28
feat: remove-createPolkadotJsScClient

josepot:

PR #922, opened 2022-04-01
Upgrade connect to 0.7.0

wirednkod:


josepot (changes requested):
Let's wait until #909 is merged

projects/demo/package.json
```diff
@@ -22,7 +22,7 @@
   },
   "dependencies": {
     "@polkadot/api": "^7.13.1",
-    "@substrate/connect": "^0.6.5",
```
josepot, line 25: FTR this will no longer be a dependency once #909 is merged, so this PR in its current shape will have conflicts after #909 is merged

projects/burnr/package.json
```diff
@@ -34,7 +34,7 @@
     "@polkadot/types": "^7.13.1",
     "@polkadot/util": "^8.6.1",
     "@polkadot/util-crypto": "^8.6.1",
-    "@substrate/connect": "^0.6.5",
```
josepot, line 37: Same

Issue #739, opened 2022-02-10
Have some kind of bot that automatically updates chain specs with new checkpoints


tomaka:
As part of https://github.com/paritytech/substrate-connect/issues/543, we need an automated way to update the checkpoint found in the chain specifications found in this repo.

I would suggest having a script automatically open pull requests, and maybe a GitHub workflow merging them automatically after verifying that the author is correct and that nothing has been altered apart from the checkpoint.

Considering that the content of the checkpoint is critical for security, this should be done internally within Parity.

tomaka:
https://github.com/paritytech/substrate/pull/10832 being merged, we can wait for 0.9.18

raoulmillais:
Does this mean we can close this issue @tomaka ?

tomaka:
Almost

[It just doesn't really work](https://github.com/paritytech/substrate-connect/pull/962#issuecomment-1098066877).

raoulmillais:
This is now working

tomaka:
Re-reading it, we also need this:

> Additionally, have another GitHub Action that, if a chain spec file is updated, checks whether smoldot can successfully connect to it and is on the same chain as a public JSON-RPC node.

PR #939, opened 2022-04-06
Update chain specs periodically through the CI

tomaka:
cc https://github.com/paritytech/substrate-connect/issues/739


.github/workflows/update-chain-specs.yml
```diff
@@ -0,0 +1,68 @@
+name: chain-specs-periodic-update
+
+on:
+  push:
+    branches: [ tka-update-chain-specs-ci ]   # TODO: remove before merging, I'm just testing
```
tomaka, line 5: When approved, I'll remove this line before merging. It's there just so that I can test if it work.

tomaka:
Actually this is wrong

tomaka:
Ok, this is ready.

Example PR be found here: https://github.com/paritytech/substrate-connect/pull/946

Issue #507, opened 2021-10-08
Smoldot - crash strategy


wirednkod:
We need to have a strategy when a smoldot crash happens.
Probably, when this occurs, we should disconnect the apps from the chain, inform the UI (emit?) of the App and/or the extension and re-start smoldot.

tomaka:
As a heads up, this can be done by catching a `CrashError` when calling any of the methods of smoldot:
https://github.com/paritytech/smoldot/blob/409ea559fa7fddfccd79c4f000def8f0e81bd350/bin/wasm-node/javascript/src/index.d.ts#L43.

I would suggest doing all of this in `index.ts`.

tomaka:
cc https://github.com/paritytech/substrate-connect/issues/429#issuecomment-1059754226

ryanleecode:
#1786

PR #958, opened 2022-04-07
Catch CrashError of smoldot, restart smoldot and add UI indications

wirednkod:
This PR takes care of the possible smoldot panic/error that crashes the extension - and addresses the last bullet on [Pierre's comment](https://github.com/paritytech/substrate-connect/issues/429#issuecomment-1059754226) and issue [#507](https://github.com/paritytech/substrate-connect/issues/507)
When an error happens on smoldot client, the extension catches the error, prints it in the logs <del>and starts a series of extension restarts. There will be 3 restarts of the extension with a 5 seconds delay between each restart.</del>
If <del>at the 3rd retry</del> a CrashError occurs <del>again</del>, then 2 icons appear on the Popup window. 
1 "alert" icon that indicates the issue in the extension and user can hover and read the error. If icon is clicked user is redirected to "paused" logs (that contain ofc the error log last - that lead to the crash)

I did not follow exactly the proposal of Pierre on the 4th bullet of his comment - meaning that if smoldot crashes but recovers the only thing that will happen will be to get disconnected from apps and try to run smoldot again. 

The only way that the _average Joe_ will understand that something really happened is if smoldot crashes 3 consecutive times and then the indication will appear (and the manual "restart extension" button)

EDIT: This covers a potential crash only during initialization of smoldot

image
image

### DIFF
--- a/projects/extension/src/background/index.ts
+++ b/projects/extension/src/background/index.ts
@@ -35,6 +35,12 @@ export interface Background extends Window {
   }
 }
 
+interface ErrorStruct {
+  name: string
+  message: string
+  stack?: string
+}
+
 interface logStructure {
   unix_timestamp: number
   level: number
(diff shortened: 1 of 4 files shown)


tomaka:
This handles the situation where smoldot crashes at initialization, but this is extremely rare.
Handling the crashes while it's running is way more difficult and I've laid down some ground but haven't finished.

wirednkod:
> This handles the situation where smoldot crashes at initialization, but this is extremely rare.
> Handling the crashes while it's running is way more difficult and I've laid down some ground but haven't finished.
> 

That is correct. This covers the crash when smoldot inits. We had plenty of that in past weeks and it would be good to catch them. Will remove the retries and allow just the manual retry

projects/extension/src/containers/Popup.tsx
```diff
@@ -1,23 +1,46 @@
...
+  const [crashError, setCrashError] = useState<string>("")
   const appliedTheme = createTheme(light)
 
   useEffect(() => {
     let isActive = true
     let unsubscribe = () => {}
 
+    chrome.storage.local.get(["crashError"], (res) => {
+      setCrashError(res.crashError)
+    })
```
tomaka, line 43: Does this auto-update if you call `chrome.storage.local.set(["crashError"], ...)`? I guess no?

projects/extension/src/background/index.ts
```diff
@@ -35,6 +35,12 @@ export interface Background extends Window {
   }
 }
 
+interface ErrorStruct {
+  name: string
+  message: string
+  stack?: string
+}
```
tomaka, line 42: Why is this necessary? Isn't this the `Error` interface?

projects/extension/src/containers/Options.tsx
```diff
@@ -184,6 +184,12 @@ const Options: React.FunctionComponent = () => {
   }, [poolingLogs])
 
   useEffect(() => {
+    chrome.storage.local.get(["crashError"], (res) => {
+      if (res.crashError) {
+        setValue(2)
```
tomaka, line 189: What's "value"? Can we have a better name, or if not at the very least a comment?

projects/extension/src/containers/Popup.tsx
wirednkod: This updates every time the popup opens.If popup is open and the localStorage update then no - but based on the nature of the popup thats really an edge scenario

wirednkod:
Due to the major changes that #984 brings, this will close and will re-implemented with Goran's designs

PR #959, opened 2022-04-08
Clean up index.ts

tomaka:
So this PR does a bunch of things.

It renames `manager` to `uiInterface` and `onManagerStateChanged` to `onChainsChanged`. This is because this interface is the communication point between the UI and the rest, and as we've discussed before, the UI doesn't need to know that there's a manager or anything like that.
The list of chains is now passed through a `get chains()` getter, rather than a function parameter. This is more for reasons of consistency with the rest of the interface, but I think it's better for clarity.

It also "flattens" the manager. Before this PR, the manager is constantly kept within a Promise. Unfortunately this becomes a mess when it comes to handling a manager in a crashed state. The manager is now an object that can be in multiple different states.

The crashed state is reported to the UI the same ways as chains: through a getter and a callback.


wirednkod (commented):
Have tested the code and even though code-wise looks good, the extension does not work anymore as before (miscommunication between popup/options and tab). Will investigate more which parts of code needs changes for it to work

tomaka:
The problem, which I've realized afterwards, is that we no longer notify the callback immediately after it has been registered.

wirednkod:
I went through this again and even though refresh seems to work fine for the popup there are some edge scenarios that are not caught from that `refresh` function. One of them is when the options page is open, an app is getting connected to a known chain (e.g. westend), and then the tab is closed.. The `Westend` network still appears after in the Options page after the tab is removed.
That is most likely due to the fact that wellKnown chain remains connected even though Tab is not using it anymore (unfortunately app appears still at the app list of the chain when expanded).
One possible solution would be to use inside the background script the:
```javascript
chrome.tabs.onRemoved.addListener((tabId, removeInfo) => {
    ..... // Remove chains connected from the tabId 
    })
 ```

**Review by wirednkod — APPROVED** [review]
looks good now

File: projects/extension/src/background/index.ts [hunk 1]
```diff
@@ -86,103 +93,165 @@ const logger = (level: number, target: string, message: string) => {
...
-        }
-      }),
-  )
+// Listeners that must be notified when the `get chains()` getter would return a different value.
+const chainsChangedListeners: Set<() => void> = new Set()
+const notifyAllChainsChangedListeners = () => {
+  chainsChangedListeners.forEach((l) => {
+    try {
+      l()
+    } catch (e) {}
```
wirednkod, line 102: Maybe we can add some error log? (using `logger`)

projects/extension/src/background/index.ts
```diff
@@ -86,103 +93,165 @@ const logger = (level: number, target: string, message: string) => {
...
+  })
+}
+// Listeners that must be notified when the `get smoldotCrashError()` getter would return a
+// different value.
+const smoldotCrashErrorChangedListeners: Set<() => void> = new Set()
+const notifyAllSmoldotCrashErrorChangedListeners = () => {
+  smoldotCrashErrorChangedListeners.forEach((l) => {
+    try {
+      l()
+    } catch (e) {}
```
wirednkod, line 112: same as above

PR #967, opened 2022-04-12
API documentation fixes

raoulmillais:
This PR solves a few documentation problems:

* Adds a new dev script which will watch for changes and rebuild all the docs and serve them on http://localhost:8080 so you can iteratively change the docs and check your changes as they will appear when published
* Standardises all the code references so they actually link to the right bit of the docs
* Makes all the protocol documentation actually be published
* Adds a note about chain names for the well known chains


wirednkod:
Im not sure why all these exports, but if its for the sake of documentation, we should probably find some other way to do so. It took a considerable amount of time and work to export only what is needed (plus some of them are already [exported](https://github.com/paritytech/substrate-connect/blob/main/packages/connect-extension-protocol/src/index.ts#L114))

wirednkod (commented):
Im not sure why all these exports, but if its for the sake of documentation, we should probably find some other way to do so. It took a considerable amount of time and work to export only what is needed (plus some of them are already [exported](https://github.com/paritytech/substrate-connect/blob/main/packages/connect-extension-protocol/src/index.ts#L114))

raoulmillais:
> Im not sure why all these exports, but if its for the sake of documentation, we should probably find some other way to do so. It took a considerable amount of time and work to export only what is needed (plus some of them are already [exported](https://github.com/paritytech/substrate-connect/blob/main/packages/connect-extension-protocol/src/index.ts?rgh-link-date=2022-04-12T15%3A39%3A26Z#L114))

Typedoc is annoyingly inflexible unfortunately.  It will only ever include documentation for exported items.  It will also warn / error (depending on config) about any links to those docs.  If you turn it off you turn link validation off for *all documentation* which we don't want because that is helpful.

So the options are:

1. export those things to get complete docs 
2. not export them and neither have docs for them nor link to those docs

Whats the argument for not having those types exported?  It seems to me that it only has upside, we get fuller docs and connect now has the option of explicitly naming those types in its sourcecode if it wants to.

package.json
```diff
@@ -31,6 +31,7 @@
     "lint": "prettier --check . && yarn workspaces -s run lint",
     "build": "yarn workspaces -s run build",
     "test": "yarn workspaces -s run test",
+    "dev:api-docs": "concurrently -n typedoc,http-server 'nodemon -e ts --watch ./packages/connect --watch ./packages/connect-extension-protocol --exec typedoc --entryPoints ./packages/connect/src/index.ts --entryPoints ./packages/connect-extension-protocol/src/index.ts' 'http-server _site/api'",
```
tomaka, line 34: Not that this is needed, but why not

tomaka (approved):

tomaka:
I'm actually very surprised that things work without the interfaces being exported. Is it really allowed that you can access non-public items from other packages?

package.json
raoulmillais: It's not *needed* but it really helps speed up the documentation workflow.  It's much easier to check the docs look right now by using that script.

wirednkod:
> I'm actually very surprised that things work without the interfaces being exported. Is it really allowed that you can access non-public items from other packages?

@tomaka interfaces were exported as needed as I [mentioned](https://github.com/paritytech/substrate-connect/blob/main/packages/connect-extension-protocol/src/index.ts#L114) above. 

I think we should export only what needs to be exported and used from other packages, and not everything for the sake of documentation - at the end of the day - we should document what is exported for the dev - right?

But again - thats just my PoV.

raoulmillais:
> > I'm actually very surprised that things work without the interfaces being exported. Is it really allowed that you can access non-public items from other packages?
> 
> @tomaka interfaces were exported as needed as I [mentioned](https://github.com/paritytech/substrate-connect/blob/main/packages/connect-extension-protocol/src/index.ts?rgh-link-date=2022-04-12T16%3A43%3A34Z#L114) above.
> 
> I think we should export only what needs to be exported and used from other packages, and not everything for the sake of documentation - at the end of the day - we should document what is exported for the dev - right?
> 
> But again - thats just my PoV.

Typescript doesn't need these to be exported unless you actually need to mention those types where they're used.  Connect doesn't do this - it only ever explicitly refers to the `ToApplication` and `ToExtension` types by name the actual type is inferred by the compiler.  The fact its not exported doesn't mattter because typescript is gradually typed.

I can see you might desire not exporting these from a "hygiene" perspective - only export what's really required but I don't see that there's any harm in actually exporting them. It doesn't enable new ways of misusing this library.  It does bring the benefit of being able to publish more complete documentation though.  It's a shame typedoc isn't more flexible in this regard.

PR #987, opened 2022-04-21
Set author and committer in update spec workflow

raoulmillais:
This PR disables credential caching when checking out the repo during a spec update workflow and then always sets the committer and author to the substrate connect user.  When merged this shold stop the `mergify` user being the author when this workflow is triggered by cron.

### DIFF
--- a/.github/workflows/update-chain-specs.yml
+++ b/.github/workflows/update-chain-specs.yml
@@ -24,6 +24,8 @@ jobs:
     steps:
     - uses: actions/checkout@v3
       with:
+        persist-credentials: false
+        fetch-depth: 0
         path: repo
         ref: main
     - run: |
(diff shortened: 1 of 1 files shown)


wirednkod (approved):

.github/workflows/update-chain-specs.yml
```diff
@@ -24,6 +24,8 @@ jobs:
     steps:
     - uses: actions/checkout@v3
       with:
+        persist-credentials: false
+        fetch-depth: 0
```
tomaka, line 28: Why this `fetch-depth: 0`? How does it relate to the problem?

raoulmillais: Only somewhat related .. It was needed for testing so that triggering a manual workflow run could successfully update the PR that was created by the cron job run.  I've removed it but it can be added back on a temporary branch if we need to do that again (triggering an action manually lets you choose the branch)

tomaka (approved):

PR #995, opened 2022-04-21
chore: make dependabot ignore major updates

josepot:
I noticed that dependabot opened a PR for React 18, which surprised me b/c I wasn't expecting dependabot to open PRs for major upgrades since they usually imply breaking changes and therefore the intervention from a developer in order to adjust to the new API. So, I looked into the docs and I found [this section](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#specifying-dependencies-and-versions-to-ignore). If I understood it correctly (please double check :pray:), then I think that this PR should make dependabot to stop sending PRs for major updates.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,3 +8,6 @@ updates:
     schedule:
       interval: daily
     open-pull-requests-limit: 4
+    ignore:
+      - dependency-name: "*"
+        update-types: ["version-update:semver-major"]


wirednkod (approved):
lgtm

tomaka:
The point of dependabot is mostly to notify you about updates. It's completely fine to have to do adjustments to dependabot PRs.

tomaka:
I've accidentally closed, but I'm in favor of this change.

raoulmillais:
I don't know how I feel about dependabot opening PRs for major semver version changes.  I can see downsides either way.  
* No PRs means major version changes could go unnoticed for longer
* Allowing major semver version changes could result in languishing failing PRs or feeling the pressure too prematurely rush into upgrade when there are more pressing things to work on.

I'm erring on the side of letting dependabot open the PRs though because we can ignore them and it should always prompt a discussion about how to tackle it when the upgrade results in a failed build.

josepot:
Yeah, good points! I didn't think of dependabot as an "announcement tool". Closing this, thanks!

PR #1009, opened 2022-04-28
upgrade yarn to latest version

### LINKED ISSUE #972: Upgrade to latest yarn
Most projects in parity are using the newest yarn and so is polkadotjs .. being on the same package manager makes life easier it's one less thing to think about when working with multiple projects locally

wirednkod:
Fixes #972


wirednkod:
> This looks too simple to be true. If it just works though, then awesome! Last time I tried to make this work it was a real pain in the ass... is it really just this `nodeLInker: node-modules` I needed to make it work?

What do you mean by node-linker? I have not used something like that I think.

tomaka:
Why do we have `.yarn/releases` and `.yarn/plugins` now?
Did the yarn people decide that you should put the source code of yarn in every single repo, or something like that?

wirednkod:
> Why do we have `.yarn/releases` and `.yarn/plugins` now? Did the yarn people decide that you should put the source code of yarn in every single repo, or something like that?
Coming from `yarn` documentation, these 2 directories contain the Yarn releases used in the current repo and these should be kept versioned so that potential issues are prevented (e.g. if two engineers use different Yarn versions with different features).

tomaka (approved):
Well, I guess good for me

PR #1020, opened 2022-05-04
log smoldot crash error to UI logs

wirednkod:


tomaka:
This function is a getter. It's just here to return a value. Getters shouldn't have side effects.

Can we not add a field next to the logs but separate from the logs showing the reason for the crash?

PR #1024, opened 2022-05-04
Fix extension log design

wirednkod:
Adjust design of logs based on [Pierre's comment](https://github.com/paritytech/substrate-connect/issues/429#issuecomment-1098769324)

image

### DIFF
--- a/projects/demo/package.json
+++ b/projects/demo/package.json
@@ -13,7 +13,7 @@
     "typecheck:watch": "tsc -b --watch",
     "copy-assets": "cp -R ./src/assets ./dist",
     "build": "yarn clean && yarn copy-assets && yarn typecheck && webpack",
-    "dev": "yarn clean && yarn copy-assets && concurrently \"webpack serve --config webpack.config.js\" \"yarn typecheck:watch\"",
+    "dev": "yarn clean && yarn copy-assets && concurrently \"yarn typecheck:watch\" \"webpack serve --config webpack.config.js\"",
     "test": "exit 0; #No tests",
     "lint": "yarn eslint . --ext .ts,.tsx"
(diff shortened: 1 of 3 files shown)


projects/demo/package.json
```diff
@@ -13,7 +13,7 @@
     "typecheck:watch": "tsc -b --watch",
     "copy-assets": "cp -R ./src/assets ./dist",
     "build": "yarn clean && yarn copy-assets && yarn typecheck && webpack",
-    "dev": "yarn clean && yarn copy-assets && concurrently \"yarn typecheck:watch\" \"webpack serve --config webpack.config.js\"",
+    "dev": "yarn clean && yarn copy-assets && concurrently \"webpack serve --config webpack.config.js\" \"yarn typecheck:watch\"",
```
tomaka, line 16: This is the kind of change that shouldn't be in this PR

projects/extension/src/containers/Options.tsx
```diff
@@ -113,25 +113,23 @@ const Options: React.FunctionComponent = () => {
...
         break
       case 3:
-        color = "#000"
-        desc = "Info"
-        break
       case 4:
-        color = "#5e5e5e"
-        desc = "Debug"
+        color = "#fff"
+        desc = "Info/Debug"
```
tomaka, line 122: Info and Debug shouldn't be merged into one, there's a reason why they're two different levels
Debug logs could have a color of `#eee` for example

projects/demo/package.json
wirednkod: That was a stray change - will revert. Thank you for catching it

tomaka (approved):

PR #1025, opened 2022-05-05
Fix databaseContent bug

wirednkod:
Fix bug that was not saving databaseContent upon first init of extension causing a false 'smoldot client crashed' error

### DIFF
--- a/projects/extension/src/background/ConnectionManager.ts
+++ b/projects/extension/src/background/ConnectionManager.ts
@@ -156,10 +156,10 @@ export class ConnectionManager<SandboxId> {
     chainName: string,
     maxUtf8BytesSize?: number,
   ): Promise<string | undefined> {
+    const chain = this.#wellKnownChains.get(chainName)!.chain
+    if (!chain) return undefined
     try {
-      return await this.#wellKnownChains
-        .get(chainName)!
-        .chain.databaseContent(maxUtf8BytesSize)
+      return await chain.databaseContent(maxUtf8BytesSize)
     } catch (error) {
       // If an exception is thrown, we kill all chains. This can only happen either in case of a
       // crash in smoldot or a bug in substrate-connect.
--- a/projects/extension/src/background/ConnectionManagerWithHealth.ts
+++ b/projects/extension/src/background/ConnectionManagerWithHealth.ts
@@ -149,7 +149,7 @@ export class ConnectionManagerWithHealth<SandboxId> {
     spec: string,
     databaseContent?: string,
   ): Promise<void> {
-    this.#inner.addWellKnownChain(chainName, spec, databaseContent)
+    await this.#inner.addWellKnownChain(chainName, spec, databaseContent)
   }
 
   /**


projects/extension/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- Fix a bug that was not saving the databaseContent of each chain in the extension's localStorage upon first initialization of the extension. ([#1025](https://github.com/paritytech/substrate-connect/pull/1025))
```
tomaka, line 7:
```suggestion
- Fix a bug that was not saving the database content of each chain in the extension's localStorage upon first initialization of the extension. ([#1025](https://github.com/paritytech/substrate-connect/pull/1025))
```

projects/extension/src/background/ConnectionManager.ts
```diff
@@ -156,10 +156,10 @@ export class ConnectionManager<SandboxId> {
     chainName: string,
     maxUtf8BytesSize?: number,
   ): Promise<string | undefined> {
+    const chain = this.#wellKnownChains.get(chainName)!.chain
+    if (!chain) return undefined
```
tomaka, line 160: No no, we want an exception here

tomaka (commented):
Also missing adding the `await` of `saveDbContent` as discussed on Element

raoulmillais (approved):

projects/extension/src/background/ConnectionManager.ts
```diff
@@ -156,10 +156,18 @@ export class ConnectionManager<SandboxId> {
...
-      return await this.#wellKnownChains
-        .get(chainName)!
-        .chain.databaseContent(maxUtf8BytesSize)
+      chain = this.#wellKnownChains.get(chainName)!.chain
+    } catch (error) {
+      // If an exception is thrown it means that the specific chainName is not among the
+      // #wellknownchains (either did not start or was disconnected). This is separated from
+      // the error below as it does not show a crash in smoldot but that a chain does not exist
+      this.#hasCrashed = "Chain does not exist in WellKnownChains"
+      return undefined
```
tomaka, line 167: Do not return undefined if the chain doesn't exist, throw an exception

projects/extension/src/background/ConnectionManager.ts
```diff
@@ -156,10 +156,17 @@ export class ConnectionManager<SandboxId> {
...
+    let chain: SmoldotChain | undefined = undefined
     try {
-      return await this.#wellKnownChains
-        .get(chainName)!
-        .chain.databaseContent(maxUtf8BytesSize)
+      chain = this.#wellKnownChains.get(chainName)!.chain
+    } catch (error) {
+      // If an exception is thrown it means that the specific chainName is not among the
+      // #wellknownchains (either did not start or was disconnected). This is separated from
+      // the error below as it does not show a crash in smoldot but that a chain does not exist
```
tomaka, line 165:
```suggestion
      // If an exception is thrown it means that the specific chainName is not among the
      // #wellknownchains. This is separate from the error below as it does not show a
      // crash in smoldot but that a chain does not exist
```

Well-known chains can't be "disconnected", there's no way to do that

tomaka (changes requested):
Still missing the two missing awaits when calling `saveDbContent`

tomaka (approved):

Issue #965, opened 2022-04-11
Allow users to set the maxLogLevel


raoulmillais:
when users run into trouble with smoldot we have no way of asking them to capture more detailed logs from smoldot.  we need a way to control the smoldot `maxLogLevel` when creating a client.  E.g. #964

tomaka:
This is _not_ about passing a `ClientOptions`, and indeed you cannot pass a `ClientOptions`. This is only about configuring `maxLogLevel`.

The options that you pass to `createSmoldotScClient` are a completely different thing than the options you pass to smoldot. Do as if they were completely unrelated and that the fact that they would both contain `maxLogLevel` should be considered a happy little coincidence.

raoulmillais:
Yeah that's an approach I'd already considered about but it doesn't change anything I've said above .. we still end up with weirdness in the API when `createScClient` is called twice with different options

raoulmillais:
If we do want to allow people to pass in a different `maxLogLevel` I think we also have to tell people not to call `createScClient` more that once (and maybe even remove the reference counting and throw instead for multiple calls)

tomaka:
Users should be able pass a `logCallback` and `maxLogLevel` to `createScClient`.
Then you pass a log callback to smoldot, and this callback dispatches logs to the various log callbacks that have been registered.
If no `logCallback`/`maxLogLevel` is provided when calling `createScClient`, then it goes to `console`

tomaka:
Actually, it's more appropriate to pass these to `addChain` so that the moment when the logs stop corresponds to removing the chain

raoulmillais:
We can't add these to `addChain` the `logCallback` and the `maxLogLevel` can only be specified when the client is created .. this is the problem I'm referring to ... we instantiate the client once and cache it and reference count it.  There's no way to have different options for different chains as things stand

raoulmillais:
Maybe I need to restate things to make it clearer.  At the time `addChain` is called it's too late to say what logging options substrate connect has already initialised the smoldot client with the logging options.  If we modify `createScClient` to allow specifying logging options those logging options can only be honoured the first time that function is called because it caches the instance of the smoldot client.  That IMO is very surprising to the developer because it violates the expected contract - "i gave this functions some options the second time I called it: why were they ignored"

tomaka:
> it might be too late to say what logging options because substrate connect might have already initialised the smoldot client with the logging options for a previous call of addChain.

[...]

PR #1027, opened 2022-05-05
Allow setting maxLoglevel when using the inline client

tomaka:
Close https://github.com/paritytech/substrate-connect/pull/969
cc https://github.com/paritytech/substrate-connect/issues/965

This PR can be reviewed commit by commit.

### DIFF
--- a/packages/connect/src/connector/smoldot-light.ts
+++ b/packages/connect/src/connector/smoldot-light.ts
@@ -22,11 +22,15 @@ const getStart = () => {
   return startPromise
 }
 
-const clientReferences: {}[] = []
+const clientReferences: Set<Config> = new Set()
 let clientPromise: Promise<Client> | null = null
+let clientReferencesMaxLogLevel = 3;
 const getClientAndIncRef = (config: Config): Promise<Client> => {
+  if (config.maxLogLevel > clientReferencesMaxLogLevel)
+    clientReferencesMaxLogLevel = config.maxLogLevel;
+
   if (clientPromise) {
-    clientReferences.push({})
+    clientReferences.add(config)
     return clientPromise
   }
 
@@ -37,7 +41,7 @@ const getClientAndIncRef = (config: Config): Promise<Client> => {
       maxLogLevel: 9999999, // The actual level filtering is done in the logCallback
       cpuRateLimit: 0.5, // Politely limit the CPU usage of the smoldot background worker.
       logCallback: (level, target, message) => {
-        if (level > 3)
+        if (level > clientReferencesMaxLogLevel)
           return;
 
         // The first parameter of the methods of `console` has some printf-like substitution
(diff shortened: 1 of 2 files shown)


tomaka:
Tests are failing because the call to `terminate()` would now be asynchronous.
I don't know whether I should decrease test coverage and increase the code complexity to make them pass.

packages/connect/src/connector/smoldot-light.ts
```diff
@@ -22,26 +22,68 @@ const getStart = () => {
   return startPromise
 }
 
-let clientNumReferences = 0
+const clientReferences: Config[] = [] // Note that this can't be a set, as the same config is added/removed multiple times
 let clientPromise: Promise<Client> | null = null
-const getClientAndIncRef = (): Promise<Client> => {
+let clientReferencesMaxLogLevel = 3
+const getClientAndIncRef = (config: Config): Promise<Client> => {
+  if (config.maxLogLevel && config.maxLogLevel > clientReferencesMaxLogLevel)
```
wirednkod, line 29: I think this works same:
```suggestion
  if (config?.maxLogLevel > clientReferencesMaxLogLevel)
```

packages/connect/src/connector/smoldot-light.ts
```diff
@@ -22,26 +22,68 @@ const getStart = () => {
...
+const decRef = (config: Config) => {
+  const idx = clientReferences.indexOf(config)
+  if (idx === -1) throw new Error("Internal error within smoldot-light")
+  clientReferences.splice(idx, 1)
+
+  // Update `clientReferencesMaxLogLevel`
+  // Note how it is set back to 3 if there is no reference anymore
+  clientReferencesMaxLogLevel = 3
+  for (const cfg of clientReferences.values()) {
+    if (cfg.maxLogLevel && cfg.maxLogLevel > clientReferencesMaxLogLevel)
```
wirednkod, line 77: I think this is the same:
```suggestion
    if (cfg?.maxLogLevel > clientReferencesMaxLogLevel)
```

packages/connect/src/connector/smoldot-light.ts
```diff
@@ -57,21 +99,43 @@ const transformErrors = (thunk: () => void) => {
...
+ */
+export interface Config {
+  /**
+   * The client prints logs in the console. By default, only log levels 1, 2, and 3 (corresponding
+   * respectively to ERROR, WARN, and INFO) are printed.
+   *
+   * In order to more easily debug problems, you can pass 4 (DEBUG) or more.
+   *
+   * This setting is only taken into account between the moment when you use this chain to add a
+   * chain for the first timen, and the moment when all the chains that you have added have been
```
wirednkod, line 113:
```suggestion
   * chain for the first time, and the moment when all the chains that you have added have been
```

wirednkod (approved):
minor things and a typo - overall lgtm

packages/connect/src/connector/smoldot-light.ts
tomaka: I don't really like this change. You're comparing null with a number, which doesn't make sense and should be forbidden by TypeScript if it was me.

wirednkod: That is not a comparison of null with a number. The `if (config?.maxLogLevel > clientReferencesMaxLogLevel)` actually translates to `if ((config === null || config === void 0 ? void 0 : config.maxLogLevel) > clientReferencesMaxLogLevel)` so its actual a comparison between numbers, but its just a detail and doesn't really matter which one we use. Result is same

wirednkod (approved):

Issue #855, opened 2022-03-10
Remove ConnectionManager.addWellKnownChain and use a regular sandbox instead


tomaka:

tomaka:
Changed my mind about this.

The problem with this idea is that the connection manager needs to handle `add-well-known-chain` incoming messages, in which case it looks in its list of well known chains. Using a sandbox for this would require a different weirdness in the API.

PR #1039, opened 2022-05-12
Fix Popup based on Figma files

### LINKED ISSUE #977: Refactor Popup Page UI to to adapt to figma files

wirednkod:
Fixes #977 

image

image


wirednkod:
There are still some things missing - but has to do with the way that smoldot API is implemented.
e.g. The integrated chains do not report peers and/or sync status at the moment (see [here](https://github.com/paritytech/substrate-connect/blob/main/projects/extension/src/background/ConnectionManagerWithHealth.ts#L184)).
I understand that this [will not be fixed](https://github.com/paritytech/substrate-connect/issues/855) - thus we cannot show the pop up (syncing/peers) when extension opens without any tab requesting connection to chains.

Another non-implemented part of the pop-up is the `Ban App` dropdown that - as discussed before - probably will not be implemented.

projects/extension/src/background/index.ts
```diff
@@ -164,6 +169,13 @@ window.uiInterface = {
       return []
     }
   },
+  get integratedChains() {
+    const s: string[] = []
+    for (let value of wellKnownChains.values()) {
+      s.push(value[0])
+    }
+    return s
+  },
```
tomaka, line 178: You don't actually need this.
Right now we intentionally don't report the well-known chains in `chains`, but it's more appropriate to simply report that in `get chains()` rather than add a separate method.

projects/extension/src/containers/Popup.tsx
```diff
@@ -1,15 +1,82 @@
...
+  peers: number
+  isSyncing: boolean
+  chainId: string
+}
 
 const Popup: FunctionComponent = () => {
-  const disconnectTabRef = useRef<(tapId: number) => void>((_: number) => {})
-  const [activeTab, setActiveTab] = useState<TabInterface | undefined>()
-  const [otherTabs, setOtherTabs] = useState<TabInterface[]>([])
+  const disconnectTab = useRef<(tabId: number, chainId?: string) => void>(
```
tomaka, line 31:
```suggestion
  const disconnectTab = useRef<(tabId: number) => void>(
```

tomaka: Also, why is this needed? Does calling `bg.uiInterface.disconnectTab()` not work?

tomaka:
> I understand that this https://github.com/paritytech/substrate-connect/issues/855 - thus we cannot show the pop up (syncing/peers) when extension opens without any tab requesting connection to chains.

I can fix this, just maybe not with the solution in https://github.com/paritytech/substrate-connect/issues/855
We shouldn't change our design/go through intermediary steps just to bypass bugs

projects/extension/src/containers/Popup.tsx
```diff
@@ -1,15 +1,80 @@
...
+import { BiDotsHorizontalRounded } from "react-icons/bi"
 
-import { Logo, ConnectedTab } from "../components"
+import { Accordion, ConnectedTab, Logo } from "../components"
 import { Background } from "../background"
-import { TabInterface } from "../types"
+
+const knownChains = ["polkadot", "kusama", "westend", "rococo"]
+
+interface PoupChains {
```
tomaka, line 17:
```suggestion
interface PopupChains {
```

projects/extension/src/containers/Popup.tsx
```diff
@@ -24,91 +89,114 @@ const Popup: FunctionComponent = () => {
...
+      unsubscribe && unsubscribe()
     }
   }, [])
 
   const goToOptions = (): void => {
     chrome.runtime.openOptionsPage()
   }
 
+  const networkIcon = (network: string) => {
+    const icon = network.replace(new RegExp(" ", "g"), "").toLowerCase()
```
tomaka, line 110: Why this regex? There's no space in any of the well known chain name?

projects/extension/src/containers/Popup.tsx
```diff
@@ -24,91 +89,114 @@ const Popup: FunctionComponent = () => {
...
+    return (
+      <>
+        <div className="icon w-7">
+          {knownChains.includes(icon) ? icon : "?"}
+        </div>
+        <div className="pl-2">{network}</div>
```
tomaka, line 116: I'm not super confident in this. This value is untrusted and entirely decided by the website. A website could have a name that is super long or has tons of weird unicode characters.

projects/extension/src/containers/Popup.tsx
```diff
@@ -24,91 +89,114 @@ const Popup: FunctionComponent = () => {
...
-          })
-        }
-      >
-        <div>About</div>
-        <div>
-          <MdCallMade className="text-base" />
-        </div>
-      </button>
+      {connChains?.map((w) => {
+        console.log("w?.details", w?.details)
```
tomaka, line 136:
```suggestion
```

projects/extension/src/containers/Popup.tsx
```diff
@@ -1,15 +1,80 @@
...
+            allChains[i].details = [
+              {
+                tabId: c.tab?.id,
+                url: c.tab?.url,
+                peers: c.peers,
+                isSyncing: c.isSyncing,
+                chainId: c.chainId,
+              },
+            ]
+          } else if (!details.map((d) => d.tabId).includes(c.tab?.id)) {
```
tomaka, line 64: What about this instead?

```suggestion
          if (!details) details = []
          if (!details.map((d) => d.tabId).includes(c.tab?.id)) {
```

projects/extension/src/containers/Popup.tsx
```diff
@@ -1,15 +1,80 @@
...
-import { Logo, ConnectedTab } from "../components"
+import { Accordion, ConnectedTab, Logo } from "../components"
 import { Background } from "../background"
-import { TabInterface } from "../types"
+
+const knownChains = ["polkadot", "kusama", "westend", "rococo"]
+
+interface PoupChains {
+  chainName: string
+  details?: ChainDetails[]
```
tomaka, line 19:
```suggestion
  details: ChainDetails[]
```

projects/extension/src/containers/Popup.tsx
```diff
@@ -1,15 +1,80 @@
...
+          allChains.push({ chainName: c.chainName })
+          allChains[allChains.length - 1].details = [
+            {
+              tabId: c.tab?.id,
+              url: c.tab?.url,
+              peers: c.peers,
+              isSyncing: c.isSyncing,
+              chainId: c.chainId,
+            },
+          ]
```
tomaka, line 51:
```suggestion
          allChains.push({ chainName: c.chainName, details: [
            {
              tabId: c.tab?.id,
              url: c.tab?.url,
              peers: c.peers,
              isSyncing: c.isSyncing,
              chainId: c.chainId,
            },
          ]})
```

projects/extension/src/containers/Popup.tsx
wirednkod: Im not sure I understand - this part of code is showing the network name. So even if there is a custom chain/parachain that is named: `123AS$@#$SomethingMoreAnd123;'_'ThisCanGoOnUntilItIsVeryLong`, still will just appear as the "network name" and icon as `?`. I'm not sure what the vulnerability is

tomaka: There's no "vulnerability", just someone could completely destroy your CSS with a very long name, unless you made sure that it can't happen

projects/extension/src/containers/Popup.tsx
```diff
@@ -1,15 +1,72 @@
...
+          const details = allChains[i]?.details
+          if (!details?.map((d) => d.tabId).includes(c.tab?.id)) {
+            details?.push({
+              tabId: c.tab?.id,
+              url: c.tab?.url,
+              peers: c.peers,
+              isSyncing: c.isSyncing,
+              chainId: c.chainId,
+            })
+          }
```
tomaka, line 64: You know that `i` is a valid index, so you don't need to handle this weird "null" situation
```suggestion
          const details = allChains[i]!.details
          if (!details.map((d) => d.tabId).includes(c.tab?.id)) {
            details.push({
              tabId: c.tab?.id,
              url: c.tab?.url,
              peers: c.peers,
              isSyncing: c.isSyncing,
              chainId: c.chainId,
            })
          }
```

projects/extension/src/containers/Popup.tsx
```diff
@@ -1,15 +1,72 @@
...
+import { BiDotsHorizontalRounded } from "react-icons/bi"
 
-import { Logo, ConnectedTab } from "../components"
+import { Accordion, ConnectedTab, Logo } from "../components"
 import { Background } from "../background"
-import { TabInterface } from "../types"
+
+const knownChains = ["polkadot", "kusama", "westend", "rococo"]
+
+interface PopupChains {
```
tomaka, line 17:
```suggestion
interface PopupChain {
```

tomaka (approved):

PR #1047, opened 2022-05-16
Move well-known chains outside of sandboxes

tomaka:
Close https://github.com/paritytech/substrate-connect/issues/855

Right now the connection manager has a list of chains and a list of well-known chains.
This PR modifies this by replacing the list of well-known chains with a list of chain specs for well-known chains, while the well-known chains themselves (i.e. the actual chain connections) are now within a sandbox of id equal to `null`.

The database content retrieval is done through additional messages in parallel with the `ToApplication` and `ToExtension`.

This makes it possible for the `ConnectionManagerWithHealth` to use the same code for well-known chains and other chains.

A notable change is that getting the list of chains (in `background/index.ts`) now returns the well-known chains.
To make this work, I've adjusted the API of `ExposedChainConnection`.


projects/extension/src/background/index.ts
```diff
@@ -305,6 +345,8 @@ chrome.runtime.onConnect.addListener((port) => {
 
         if (message.type === "chains-status-changed") {
           notifyAllChainsChangedListeners()
+        } else if (message.type === "database-content") {
+          // We never ask for the database content of a chain added through a port.
         } else {
```
wirednkod, line 350: If there are no actions in `database-content` better remove `else` and just check if the type is NOT `database-content`
```suggestion
        } else if (message.type !== "database-content") {
```

projects/extension/src/background/ConnectionManager.ts
```diff
@@ -342,8 +305,11 @@ export class ConnectionManager<SandboxId> {
         }
 
         // Refuse the chain addition for invalid well-known chain names.
-        if (message.type === "add-well-known-chain") {
-          if (!this.#wellKnownChains.has(message.chainName)) {
+        if (
+          message.type === "add-well-known-chain" ||
+          message.type === "add-well-known-chain-with-db"
+        ) {
+          if (!this.#wellKnownChainSpecs.has(message.chainName)) {
```
wirednkod, line 312: Since [keys of wellKnownChains](https://github.com/paritytech/substrate-connect/blob/main/projects/extension/src/background/index.ts#L21-L26) are all `.id` for better readability and less confusion lets use the chainId ([for well known chains `chainName === chainId` anyways](https://github.com/paritytech/substrate-connect/pull/1047/files#diff-4de63d371461c5b51ea9a173eccc4c3b114374baa9388b564a3033e112f58924R221-R222))
```suggestion
          if (!this.#wellKnownChainSpecs.has(message.chainId)) {
```

projects/extension/src/background/ConnectionManager.ts
```diff
@@ -359,10 +325,15 @@ export class ConnectionManager<SandboxId> {
         const chainSpec =
           message.type === "add-chain"
             ? message.chainSpec
-            : this.#wellKnownChains.get(message.chainName)!.spec
+            : this.#wellKnownChainSpecs.get(message.chainName)!
```
wirednkod, line 328: Same as before
```suggestion
            : this.#wellKnownChainSpecs.get(message.chainId)!
```

wirednkod (approved):
lgtm - some details here and there only...

projects/extension/src/background/ConnectionManager.ts
tomaka: That's wrong. The message can come from the connect package, where the `chainId` is assigned randomly.

projects/extension/src/background/index.ts
wirednkod: > Well, right now it's explained why the `database-content` messages are ignored

Still.. An empty condition is just not nice :)

projects/extension/src/background/ConnectionManager.ts
wirednkod: > That's wrong. The message can come from the connect package, where the `chainId` is assigned randomly.

Found it:
`const chainId = getRandomChainId()`
I think we should alter the [part in connect](https://github.com/paritytech/substrate-connect/blob/main/packages/connect/src/connector/extension.ts#L97-L101) when its a well-known-chain to assign to the chainId the name (as we do in extension as well), or randomise the ID in the extension. Its not about the change I propos but IMO its good to have a common approach in both cases. Opinions?

tomaka: That'd be wrong as well. Multiple libraries based on top of connect and running at the same time need to be able to create each their own connection to a well-known chain, and thus the `chainId` needs to be different to distinguish these multiple connections.

wirednkod: Thank you for the feedback. Then Its good to be merged.. I mean both comments are just "comments for possible improvement" and PR is already approved.

projects/extension/src/background/index.ts
wirednkod: It does not make sense to have empty conditions. It does make the code more self-explanatory and readable - that is true - but it is "nice" not to have empty condition statements and condition check only what have an outcome based on condition.
Again though, that is opinion-based and details only... You can proceed as you find it best

PR #1052, opened 2022-05-17
Print error on spec download failure

tomaka:


.github/workflows/update-chain-specs.yml
```diff
@@ -4,6 +4,7 @@ on:
   schedule:
     - cron: '0 8 * * *'  # every day at 8am
   workflow_dispatch: # allow triggering through the UI
+  push:
```
tomaka, line 7: This is to debug the pull request, should be removed before merging

wirednkod (approved):

Issue #981, opened 2022-04-14
Add bootnodes functionality API and save in extension localstorage


wirednkod:

tomaka:
Modifying a bootnode from the UI should call a method in the `Background` interface in `index.ts`, and it's `index.ts` that handles the logic

wirednkod:
cc #492

wirednkod:
Fixed with #1079

tomaka:
You didn't add methods to add/remove bootnodes

PR #1079, opened 2022-05-31
Add bootnodes in localstorage

wirednkod:
Covers #981


projects/extension/src/background/index.ts
```diff
@@ -12,6 +12,39 @@ import {
...
+  }
+  if (!r) {
+    const { id, bootNodes } = rococo_v2_2
+    localStorage.setItem(id, bootNodes.join(","))
+    r = bootNodes
+  }
+  polkadot.bootNodes = p
+  ksmcc3.bootNodes = k
+  westend2.bootNodes = w
+  rococo_v2_2.bootNodes = r
```
tomaka, line 43: You shouldn't modify the original chain specs. Please return copies.

projects/extension/src/background/index.ts
```diff
@@ -173,6 +212,14 @@ window.uiInterface = {
     if (manager.state === "crashed") return manager.error
     else return undefined
   },
+  get wellKnownChainBootnodes() {
+    return {
+      [polkadot.id]: polkadot.bootNodes,
+      [ksmcc3.id]: ksmcc3.bootNodes,
+      [westend2.id]: westend2.bootNodes,
+      [rococo_v2_2.id]: rococo_v2_2.bootNodes,
```
tomaka, line 220: I don't understand this. Are the keys arrays of the object, or is that some weird JS syntax?

projects/extension/src/background/index.ts
```diff
@@ -12,6 +12,39 @@ import {
   ToExtension,
 } from "@substrate/connect-extension-protocol"
 
+const setupStorageBootnodes = () => {
+  let p = localStorage.getItem(polkadot.id)?.split(",")
```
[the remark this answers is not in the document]
wirednkod: localStorage is the browser's storage (not the extension's). I prefer to avoid using chrome.storage.local for this as  chrome.storage APIs are asynchronous while localStorage is sync

projects/extension/src/background/index.ts
wirednkod: Some [weird TS syntax](https://www.typescriptlang.org/docs/handbook/2/objects.html#index-signatures)

wirednkod: This is due to [index signatures](https://www.typescriptlang.org/docs/handbook/2/objects.html#index-signatures).
Due to the fact that `ChainBootnodes` (which is the return of the [wellKnownChainBootnodes](https://github.com/paritytech/substrate-connect/pull/1079/files#diff-4de63d371461c5b51ea9a173eccc4c3b114374baa9388b564a3033e112f58924R215) has an indexed signature, and since the `ids` are strings they can be used  as indexes.
The following code is equivalent to the one above (just the one above is shorthanded):
```ts
    const obj: ChainBootnodes = {}
    obj[polkadot.id] = polkadot.bootNodes
    obj[ksmcc3.id] = ksmcc3.bootNodes
    obj[westend2.id] = westend2.bootNodes
    obj[rococo_v2_2.id] = rococo_v2_2.bootNodes
    return obj
```

projects/extension/src/background/index.ts
```diff
@@ -12,17 +12,55 @@ import {
   ToExtension,
 } from "@substrate/connect-extension-protocol"
 
+const polkadot_cp = Object.assign({}, polkadot)
+const ksmcc3_cp = Object.assign({}, ksmcc3)
+const westend2_cp = Object.assign({}, westend2)
+const rococo_cp = Object.assign({}, rococo_v2_2)
+
+const setupStorageBootnodes = () => {
```
tomaka, line 20:
```suggestion
const setupStorageBootnodes = () -> [any] => {
```

projects/extension/src/background/index.ts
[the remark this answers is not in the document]
tomaka: What's the difference?

projects/extension/src/background/index.ts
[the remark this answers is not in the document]
wirednkod: I do not understand what the code suggestion above means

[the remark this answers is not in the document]
tomaka: You don't need the `export const wellKnownChains: Map<string, string>` anymore.
The only place where `wellKnownChains` is used is in `whenInitFinished` below, so instead of using `whenInitFinished` in `whenInitFinished` you can just call this function and use what it returns.
This also makes it possible to make this function async.

projects/extension/src/background/index.ts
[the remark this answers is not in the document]
wirednkod: All mentioned [here](https://developer.chrome.com/docs/extensions/reference/storage/#overview). We could for the sake of consistency use the chrome.localstorage API

projects/extension/src/background/index.ts
```diff
@@ -12,18 +12,60 @@ import {
...
+      "bootNode_".concat(ksmcc3_cp.id),
+      "bootNode_".concat(westend2_cp.id),
+      "bootNode_".concat(rococo_cp.id),
+    ],
+    (result) => {
+      let i = "bootNode_".concat(polkadot_cp.id)
+      if (result[i]) {
+        polkadot_cp.bootNodes = result[i]
+      } else {
+        chrome.storage.local.set({ [i]: polkadot_cp.bootNodes })
```
tomaka, line 33:
```suggestion
      } else {
        chrome.storage.local.set({ [i]: polkadot_cp.bootNodes })
```

Well, we don't need to do this, do we?

projects/extension/src/background/index.ts
```diff
@@ -12,18 +12,60 @@ import {
...
+  const ksmcc3_cp = Object.assign({}, ksmcc3)
+  const westend2_cp = Object.assign({}, westend2)
+  const rococo_cp = Object.assign({}, rococo_v2_2)
+
+  chrome.storage.local.get(
+    [
+      "bootNode_".concat(polkadot_cp.id),
+      "bootNode_".concat(ksmcc3_cp.id),
+      "bootNode_".concat(westend2_cp.id),
+      "bootNode_".concat(rococo_cp.id),
```
tomaka, line 26:
```suggestion
      "bootNodes_".concat(polkadot_cp.id),
      "bootNodes_".concat(ksmcc3_cp.id),
      "bootNodes_".concat(westend2_cp.id),
      "bootNodes_".concat(rococo_cp.id),
```

projects/extension/src/background/index.ts
wirednkod: 🤦🏽 Thank you... seriously sometimes I get surprised on these kind of misses from me 🤦🏽

tomaka (approved):

PR #1084, opened 2022-06-03
Add the latest block height to the chain information provided by the manager

tomaka:
This also fixes two issues:

- We didn't properly bump `nextRpcRqId` when restarting the subscription.
- We didn't unpin the blocks that were notified to us.

Can be reviewed commit by commit.

I haven't tested whether it works, sorry.

### DIFF
--- a/projects/extension/src/background/ConnectionManagerWithHealth.ts
+++ b/projects/extension/src/background/ConnectionManagerWithHealth.ts
@@ -495,7 +495,67 @@ interface Chain {
 
 // TODO: this code uses `system_health` at the moment, because there's no alternative, even in the new JSON-RPC API, to get the number of peers
 
-// Converts a block header, as a hexadecimal string, to a block height. Returns undefined
+// Converts a block header, as a hexadecimal string, to a block height.
+//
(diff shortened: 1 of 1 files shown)


projects/extension/src/background/ConnectionManagerWithHealth.ts
```diff
@@ -313,17 +358,72 @@ export class ConnectionManagerWithHealth<SandboxId> {
...
+                  chain.bestBlockHeaderRequestId =
+                    "best-block-header:" + this.#nextRpcRqId
+                  this.#nextRpcRqId += 1
+                  break
+                }
+                case "finalized": {
+                  // When one or more new blocks get finalized, we unpin all blocks except for
+                  // the new current finalized.
+                  let finalized = jsonRpcMessage.params.result
+                    .finalizedBlockHashes as [string]
```
wirednkod, line 404:
```suggestion
                  let finalized = jsonRpcMessage.params.result
                    .finalizedBlockHashes as [string]
                    if (!finalized) break
```

wirednkod (approved):
Tested and it lgtm

PR #1091, opened 2022-06-06
Fix 'invalidated context' bug

wirednkod:
This PR fixes a repeatable error that was appearing in the Extension background script's console:
```
index.ts:103 Uncaught exception in onChainsChanged callback: Error: Extension context invalidated.
    at l (popup.js:2:2485356)
    at index.ts:101:7
    at Set.forEach (<anonymous>)
    at l (index.ts:99:26)
    at index.ts:264:13
    at Generator.next (<anonymous>)
    at t (jsonp chunk loading:1:59)
```

That was due to the way that background was "called" from the popup, which was invalidating the context in each re-render of UI


tomaka:
Can you explain in more details what the problem is and why you solved it this way?

wirednkod:
> Can you explain in more details what the problem is and why you solved it this way?

The problem was created due to [the way that background page is retrieved](https://github.com/paritytech/substrate-connect/blob/main/projects/extension/src/containers/Popup.tsx#L70-L88) in the useEffect - which makes the background to be updated every time there is a rerender of the code (multiple times per minute as can be seen in the screenshot below):

Screenshot 2022-06-06 at 11 34 27 AM

So in each call of refresh there is a new retrieval of the background page that replaces the previous one and thus invalidates the context. The solution that is implemented in this PR, is to retrieve the background page [only once upon constructing the popup](https://github.com/paritytech/substrate-connect/pull/1091/files#diff-b6649eb39ac6b3f307cafb8dd91565c997660d533a1d68e7cb2f020feb098dbfR69-R73) (every time the popup opens) and add it in a state variable - then use that state variable to call all needed functions.

projects/extension/src/containers/Popup.tsx
```diff
@@ -30,68 +30,62 @@ const Popup: FunctionComponent = () => {
...
-
-      chrome.runtime.getBackgroundPage((backgroundPage) => {
-        if (!isActive) return
-        const bg = backgroundPage as Background
-        disconnectTab.current = bg.uiInterface.disconnectTab
-        unsubscribe = bg.uiInterface.onChainsChanged(refresh)
-        refresh()
-      })
-    })()
+    if (!isActive || !bg) return
```
tomaka, line 79: This `isActive` variable seems completely useless

projects/extension/src/containers/Popup.tsx
```diff
@@ -73,16 +73,14 @@ const Popup: FunctionComponent = () => {
   }, [])
 
   useEffect(() => {
-    let isActive = true
     let unsubscribe = () => {}
 
-    if (!isActive || !bg) return
+    if (!bg) return
     disconnectTab.current = bg.uiInterface.disconnectTab
     unsubscribe = bg.uiInterface.onChainsChanged(() => refresh())
```
tomaka, line 80:
```suggestion
    const unsubscribe = bg.uiInterface.onChainsChanged(() => refresh())
```

projects/extension/src/containers/Popup.tsx
```diff
@@ -73,16 +73,14 @@ const Popup: FunctionComponent = () => {
...
     refresh()
 
     return () => {
-      isActive = false
       unsubscribe && unsubscribe()
     }
```
tomaka, line 85: `return unsubscribe`

projects/extension/src/containers/Popup.tsx
```diff
@@ -73,16 +73,12 @@ const Popup: FunctionComponent = () => {
...
     if (!bg) return
     disconnectTab.current = bg.uiInterface.disconnectTab
-    unsubscribe = bg.uiInterface.onChainsChanged(() => refresh())
+    const unsubscribe = bg.uiInterface.onChainsChanged(() => refresh())
     refresh()
 
-    return () => {
-      unsubscribe && unsubscribe()
-    }
+    return unsubscribe && unsubscribe()
```
tomaka, line 81:
```suggestion
    return unsubscribe
```

It can never be null, can it?

wirednkod: It cannot

tomaka: Plus, you don't cant to call `unsubscribe` now, you just want to return it from `useEffect` as a clean up function

tomaka (approved):

PR #1093, opened 2022-06-08
Fix messages from the extension being potentially missed

tomaka:
This PR refactors `extension.ts` in order to guarantee that messages from the extension concerning a chain can never be missed.

Furthermore, we no longer return an error at initialization if the extension sends an unexpected message. Instead, we just print a warning and move on. This is less harmful than introducing a state mismatch between the connect package and the extension.


packages/connect/src/connector/extension.ts
```diff
@@ -57,95 +50,204 @@ export const createScClient = (): ScClient => {
...
-          ...msg,
-          origin: "substrate-connect-client",
-          chainId,
-        },
-        "*",
-      )
-    }
+    type ChainState =
+      | {
+          state: "wait-ack-or-err"
```
wirednkod, line 55: should we name this state: "pending"? I mean - it does not really matter - just sounds better than "wait-ack-or-err"

packages/connect/src/connector/extension.ts
```diff
@@ -57,95 +50,204 @@ export const createScClient = (): ScClient => {
...
+          jsonRpcMessage,
+        })
       },
       remove: () => {
-        if (crashError) throw crashError
-        if (!chains.has(chain)) throw new AlreadyDestroyedError()
-        listeners.delete(chainId)
+        if (chainState.state.state === "dead") {
+          throw chainState.state.error
+        }
```
wirednkod, line 230: This code is multiple times, maybe we should add it in a function that will accept the state as an input?

wirednkod (approved):
Other than couple of nip tucks, lgtm

packages/connect/src/connector/extension.ts
tomaka: I mean, it's 3 lines of code that are present 3 times.

wirednkod: Its not important. Just a proposal

PR #1113, opened 2022-06-16
add dependencies in extension

wirednkod:


projects/extension/package.json
```diff
@@ -7,6 +7,7 @@
   "license": "GPL-3.0-only",
   "type": "module",
   "scripts": {
+    "bundle-source-code": "cp tsconfig.json tsconfig_bkp.json && sed -i '26d' tsconfig.json && sed -i '2d' tsconfig.json && zip -r substrate-connect_source-code.zip . -x \"./node_modules/*\" -x \"./dist/*\" && mv tsconfig_bkp.json tsconfig.json",
```
tomaka, line 10: Uuuhhh, that's a no for me
If there's really no other way than manually editing JSON files with `sed`, then it should at the very least be **thoroughly** explained

tomaka:
I'd rather not do that before workspaces have been removed

wirednkod:
> I'd rather not do that before workspaces have been removed

I dont think that workspaces should be removed. There are dependencies needed that work fine as is.

Issue #982, opened 2022-04-14
create UI for bootnodes editing


wirednkod:

wirednkod:
This is completed

PR #1128, opened 2022-06-21
Add API for updating bootnodes and getting default ones from chainspecs

### LINKED ISSUE #981: Add bootnodes functionality API and save in extension localstorage

wirednkod:
- Add API for updating bootnodes and getting default ones from chainspecs
- Create Settings page with Bootnodes editing content

Fixes #981 #982


projects/extension/src/background/index.ts
```diff
@@ -73,6 +73,12 @@ export interface Background extends Window {
     onChainsChanged: (listener: () => void) => () => void
     onSmoldotCrashErrorChanged: (listener: () => void) => () => void
     disconnectTab: (tabId: number) => void
+    getDefaultBootnodes: (chain: string) => string[]
+    saveBootnodes: (chain_id: string, bootnodes: string[]) => void
+    validateAndAddBootnode: (
```
tomaka, line 78: Why is there a separate `saveBootnodes` section? Can `validateAndAddBootnode` not also save the bootnode?
Having two functions implies that you add the bootnode to the currently running light client, but this bootnode might not be saved for next time.

projects/extension/src/background/index.ts
```diff
@@ -165,6 +176,8 @@ const notifyAllSmoldotCrashErrorChangedListeners = () => {
   })
 }
 
+let extConfigRpcRqId = 0
```
tomaka, line 179: The `ConnectionManagerWithHealth` already has this system in place

projects/extension/src/background/ConnectionManager.ts
```diff
@@ -52,6 +53,14 @@ export interface ToConnectionManagerAddWellKnownChain {
   databaseContent?: string
 }
 
+export interface ToConnectionManagerExtConfig {
+  origin: "trusted-user"
+  type: "ext-config"
+  chainId: string
+  chainName: string
+  jsonRpcMessage: string
+}
```
tomaka, line 62: This is exactly the same thing as the regular RPC message.
The `ConnectionManagerWithHealth` already has a system in place to send RPC requests. There's no need to add *another* mechanism to the `ConnectionManager`. Only the `ConnectionManagerWithHealth` needs to be changed.

projects/extension/src/components/Settings.tsx
```diff
@@ -0,0 +1,277 @@
...
+      <Title>Network Bootnodes</Title>
+      <Title titleType="small">Default</Title>
+      <div className="mb-8">
+        {defaultWellKnownChainBn?.map((bn) => (
+          <div className="leading-4 flex items-center mb-2">
+            <CheckBox
+              bootnode={bn}
+              alterBootnodes={alterBootnodes}
+              defaultBootnode={true}
+              isChecked={defaultBn.map((d) => d.bootnode).includes(bn)}
```
tomaka, line 192: IMO the bootnode should be "validated" before being added to the list. Checking if a bootnode address is valid takes in the order of milliseconds.
It doesn't make sense for a bootnode to be in the list and invalid. This would be plain confusing.

projects/extension/src/background/index.ts
tomaka, line 179: The `ConnectionManagerWithHealth` already has this system in place

tomaka (changes requested):

projects/extension/src/background/index.ts
wirednkod: It does but its a private method - so I thought to keep it separated. I can always create a `get` to retrieve the private variable if that sounds ok

projects/extension/src/background/index.ts
wirednkod: `saveBootnodes` is actually saving all the given `string[]` into the localStorage while `validateAndAddBootnode` is only checking the validity and not adding in the localstorage. 

> Having two functions implies that you add the bootnode to the currently running light client, but this bootnode might not be saved for next time.

That is correct. User may want to check the validity of the bootnode but decide not to finally have it in the list of the saved bootnodes. Localstorage is updated only when `Save` is pressed, no matter of the action that the user is doing since last save.

projects/extension/src/components/Settings.tsx
wirednkod: In order for a bootnode to be added to the list, it needs first to be validated. There are 2 lists, the "default bootnodes" one that contain all bootnodes coming from the chainspecs (already validated by us), and an empty "custom bootnodes" one. In the latter, bootnodes can be added only through the "Add" functionality/button which calls the `validateAndAddBootnode` function. This (as per name) validates the bootnode and add it to the custom list. Once user hits "save" then it is added in the localstorage with the "Custom bootnodes" list.

TLDR; All bottnodes in lists with checkboxes are valid - to add one it first gets validated with `unstabe_p2p_discover`

projects/extension/src/background/index.ts
tomaka: > Localstorage is updated only when Save is pressed, no matter of the action that the user is doing since last save.

But why? Can't it just save whenever you add or remove a bootnode?

wirednkod: I mean it can - but that makes `Save` obsolete.. 
I can make that change, but that means I won't be following designs and flow logic given on figma.

projects/extension/src/background/index.ts
```diff
@@ -191,6 +202,70 @@ window.uiInterface = {
...
+          params: [bootnode],
+        }),
+      })
+
+      while (true) {
+        const msg = await manager.manager.nextSandboxMessage(null)
+        if (
+          msg.type === "extension-config" &&
+          msg.origin === "connection-manager"
+        ) {
```
tomaka, line 253: This is wrong. If there are any other messages pending, you're just silently discarding them.
The logic of that whole thing should be in `ConnectionManagerWithHealth`.

projects/extension/src/background/index.ts
```diff
@@ -191,6 +193,55 @@ window.uiInterface = {
...
+        chainId: chain,
+        type: "rpc",
+        origin: "substrate-connect-client",
+        jsonRpcMessage: JSON.stringify({
+          id: "extension:" + nextId,
+          jsonrpc: "2.0",
+          method: "sudo_unstable_p2pDiscover",
+          params: [bootnode],
+        }),
+      })
```
tomaka, line 242: So you're using the local storage as if it was a variable, basically.
What if the extension shuts down while the request is in progress. The next time it loads you'll still have the item in there, except that now it doesn't correspond to any request.

Please instead just add an asynchronous method to the `ConnectionManagerWithHealth` that sends a request, and returns the response to the request.

projects/extension/src/background/index.ts
```diff
@@ -138,6 +143,8 @@ const logger = (level: number, target: string, message: string) => {
   all.push(incLog)
 }
 
+const bootnodeRequest = new Map<number, { chain: string; bootnode: string }>()
```
tomaka, line 146: This should be a field `ConnectionManagerWithHealth`

projects/extension/src/background/ConnectionManagerWithHealth.ts
```diff
@@ -301,6 +335,19 @@ export class ConnectionManagerWithHealth<SandboxId> {
...
+              // created Map and save to the localStorage
+            } else if (jsonRpcMessageId.startsWith("extension:")) {
+              jsonRpcMessage.id = JSON.parse(
+                (jsonRpcMessage.id as string).slice("extension:".length),
+              )
+              return {
+                origin: "connection-manager",
+                type: "extension-config",
+                jsonRpcMessage: JSON.stringify(jsonRpcMessage),
+              }
```
tomaka, line 350: ...and then here you use the `Map`, instead of returning a message.

wirednkod: Im not sure what you mean. This is needed, in order to "inform" `index` that rpc was returned and returning `type: "extension-config"` is also needed in order to identify that this is not any other RPC and initiate the `verifyBootnode` listeners.. Id and error that are returning from the rpc are needed for mapping the outcome thus jsonRpcMessage must be returned to `index` (error message should appear to the UI).

tomaka: Right now you call `validateBootnode` you have the guarantee to later receive back a JSON-RPC response of type `extension-config` (why "extension-config"?) containing an id that you must then pass to `getBootnode`.
I don't know if that makes sense to you, but I consider this a spaghetti mess.

What I'm saying is that you should get rid of this idea of returning an `extension-config` message and of `getBootnode`, and instead pass some kind of callback to `validateBootnode` that is called at this point in the code (where we're commenting right now).

tomaka: Right now you call validateBootnode, and you then have the guarantee to later receive a JSON-RPC response of type extension-config (why "extension-config"?) containing an id that you must then pass to getBootnode.
I don't know if that makes sense to you, probably yes since you've design this, but I consider this a spaghetti mess.

What I'm saying is that you should get rid of this idea of returning an extension-config message and of getBootnode, and instead put some kind of callback inside the `Map`, that is called at this point in the code (where we're commenting right now).

Then you can change `validateBootnode`. The callback that is put in the map could be provided to `validateBootnode`, but a `Promise` would IMO be even better.

wirednkod:
Closein favor of #1294

PR #1143, opened 2022-07-08
Add CJS support to connect-extension-protocol

wirednkod:
This PR adds CJS support to `connect-extension-protocol`.
It is needed for adding CJS support to `@substrate/connect` 

This needs to merge prior to #1142 

This also needs to be merged AFTER [#2487](https://github.com/paritytech/smoldot/pull/2487)


packages/connect-extension-protocol/fix-package-type
```diff
@@ -0,0 +1,16 @@
...
+{
+    "type": "commonjs"
+}
+!EOF
+
+cat >dist/mjs/package.json <<!EOF
+{
+    "type": "module"
+}
+!EOF
```
tomaka, line 16: Why is this needed? There's already a `package.json`, why do we need two additional ones?

wirednkod: Its not exactly "two additional ones", rather "two  complementary ones". These `package.json` files define the default package type for each of the `dist/*` sub-directories. Main `package.json` does not have a type property (if it did - hybrid solution would not work), and with this script it is pushed down to the `package.json` files under `dist/`.

tomaka: It doesn't answer my question: _why_ do we need them?

wirednkod: When `type` is set to `module` it clarifies that this is an ESM package - we could avoid that by changing the `js` extension to `mjs` but I find it unnecessary as a simple `"type" : "module"` would do the same.

packages/connect-extension-protocol/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
 # Changelog
 
 ## [Unreleased]
+
+### Added
+
+- Add CJS support to @substrate/connect-extension-protocol ([#1143](https://github.com/paritytech/substrate-connect/pull/1143))
```
tomaka, line 7:
```suggestion
- Add CommonJS support to @substrate/connect-extension-protocol ([#1143](https://github.com/paritytech/substrate-connect/pull/1143))

```

packages/connect/tsconfig.json
```diff
@@ -22,5 +22,5 @@
     "src/connector/specs/*.json"
```
tomaka, line 22: Can you remove `outDir`, `module` and `target` from `tsconfig.json`?

packages/connect/tsconfig.json
```diff
@@ -22,5 +22,5 @@
...
   ],
   "include": ["src/**/*", "src/**/*.json"],
-  "references": [{ "path": "../connect-extension-protocol" }]
+  "references": [{ "path": "../connect-extension-protocol/tsconfig-mjs.json" }]
```
tomaka, line 25: Why is that needed?

packages/connect-extension-protocol/tsconfig-mjs.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "module": "esnext",
+    "outDir": "./dist/mjs",
+    "target": "esnext"
```
tomaka, line 6:
```suggestion
    "target": "es2018"
```

Let's keep the values that were in `tsconfig.json`

packages/connect-extension-protocol/tsconfig-mjs.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "module": "esnext",
```
tomaka, line 4:
```suggestion
    "module": "es2020",
```

packages/connect-extension-protocol/fix-package-type
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
tomaka, line 1: Can you add a `.sh` extension to this file name?

tomaka (changes requested):

tomaka (approved):

PR #1164, opened 2022-07-19
Fix deploy script

wirednkod:
Make deploy script working again


wirednkod:
`typedoc` update to latest versions does not allow deploy to pass. 
As mentioned in TS docs `downlevelIteration is used for transpiling to an older version of JavaScript`. Since we now are transpiling also to CommonJS there are some errors, I guess

tomaka (approved):
This answered my question:
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-3.html#new---downleveliteration
https://mariusschulz.com/blog/downlevel-iteration-for-es3-es5-in-typescript#reducing-code-size-with-importhelpers-and-tslib

TL;DR: `downLevelIteration` isn't on by default because it bloats the generated code a bit. But we don't really care about that.

PR #1167, opened 2022-07-20
Bump babel to 7.18.*

wirednkod:


tomaka:
What's the reason why the `yarn.lock` file has only additions? Do we indirectly depend on the old version too?

wirednkod:
> What's the reason why the `yarn.lock` file has only additions? Do we indirectly depend on the old version too?

Im not sure what you mean by that. This is just a bulk bump of versions of babel in order to bump rest of dependencies next

tomaka (approved):

PR #1171, opened 2022-07-26
Add documentation concerning supported bundlers

wirednkod:


tomaka:
Sorry but I don't think we should be linking to your private repositories. Even if they were moved to `paritytech`, they can very quickly become obsolete. Even the information that we support a certain bundler can become obsolete, unless we checked compatibility on the CI.

wirednkod:
> checked compatibility on the CI.

Then we should do this

PR #1180, opened 2022-07-29
Create Modal for Brave Browser notification

### LINKED ISSUE #1178: Implement UI for `Restrict WebSockets pool` on Brave Browser
As per #1016 we should implement something similar like Talisman extension's popup informing users for websocket restrictions that Brave browser has implemented.

We should create 2 Different Modals, that will appear if the browser is identified as `Brave`:
- 1 inside the popup and
- 1 inside the Options page.

This Popup should be able to be dismissed (either Dismiss button or X) and should be able to navigate user to the browser settings with a "Settings button". We can also have a "Learn more" button that will lead user to a [section in Learning page](#1179) - where we can add a "Note" concerning Brave browser (similar to [this](https://docs.talisman.xyz/talisman/help-and-support/troubleshooting/balances-on-brave-not-showing))

More information about the restriction can be found [here](https://docs.talisman.xyz/talisman/help-and-support/troubleshooting/balances-on-brave-not-showing).

This is a solution until [WebRTC](https://github.com/paritytech/substrate-connect/issues/1016#issuecomment-1115757054) is integrated which will probably resolve the case.

Talisman notifications:

Popup:
Screenshot 2022-07-29 at 12 17 15 PM

Options page:
Screenshot 2022-07-29 at 12 17 27 PM

wirednkod:
This PR creates the notification inside Popup and Options (only in Brave Browser) about the socket restriction that was introduced in Brave Browser in [version 1.36.109](https://github.com/brave/brave-browser/issues/19990).

Final design details are still pending.
Fixes #1178 

A new PR will open to adjust the design when it is provided.

Popup outcome:
Screenshot 2022-07-29 at 6 25 27 PM

Options outcome:
Screenshot 2022-07-29 at 6 25 09 PM


wirednkod:
Some extra info: `Dismiss` button is adding an entry in localStorage for the note to stop appearing.
Learn more should lead to Landing Page section that will explain what happened with Brave Browser

tomaka (approved):
I'm not a fan of using `localStorage` directly from within UI components. This is spaghetti-ish. But I don't want to go back and forth with review comments.

wirednkod:
> I'm not a fan of using `localStorage` directly from within UI components. This is spaghetti-ish. But I don't want to go back and forth with review comments.

I see what you mean - will transfer it to the background. Thank you

tomaka:
Your last commit is actually IMO worse than what there was before.
The problem of using `localStorage` directly has nothing to do with using `localStorage` **directly**, it has to do with **using `localStorage`**.
The code of the UI component shouldn't contain any reference to the word "local storage". It should query the state of this modal popup, and it's `background.ts` that decides that this state is stored in the local storage.

wirednkod:
> Your last commit is actually IMO worse than what there was before. The problem of "using `localStorage` directly" has nothing to do with using `localStorage` **directly**, it has to do with **using `localStorage`**. The code of the UI component shouldn't contain any reference to the word "local storage". It should query the state of this modal popup, and it's `background.ts` that decides that this state is stored in the local storage.

The wording in your comment, sounded like the UI components should not access chrome.storage at all, rather than `"querying" the state of this modal popup, and it's 'background.ts' that decides that this state is stored in the local storage`.  Comment above makes clearer what you actually meant. Will alter in a new PR

PR #1214, opened 2022-08-15
Pin version of Rober19/compare-npm-versions-ci GHA to specific commit

sergejparity:
Addresses https://gitlab.parity.io/parity/vulnerability-response/-/issues/50
Quick fix is to pin GHA to specific commit so further possible changes may not affect our setup.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -15,7 +15,7 @@ jobs:
       - run: yarn install --immutable
       - run: yarn build
       - name: Compare versions of npm with local
-        uses: Rober19/compare-npm-versions-ci@master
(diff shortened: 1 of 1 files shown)


radupopa2010 (approved):

alvicsam (approved):

sergejparity:
> > Quick fix is to pin GHA to specific commit so further possible changes may not affect our setup.
> 
> If there is this case I prefer to remove this npm action and replace with a more secure one rather than using a specific commit

Fully agree with you. This is something like "urgent" fix to freeze any possible changes till moment when appropriate solution will be found and tested

wirednkod (approved):

Issue #1225, opened 2022-08-22
Extensions crashes after sometime using it - smoldot error message


filippoweb3:
On multiple occasions and on multiple devices the extension crashed. The only way of restoring it is to delete it from the browser and re-install it. While crashing there is also a smoldot error message on the UI.

<img width="364" alt="Screenshot 2022-08-22 at 15 50 57" src="https://user-images.githubusercontent.com/110459737/185937604-8d158e24-0a94-4202-b160-aaf000472fc2.png">

wirednkod:
An error message has been added with #1228

PR #1228, opened 2022-08-29
Add error message in case Smoldot panics

wirednkod:
Added an error message in Popup and Options page in case all networks stop running in the extension.
Partially addresses #1225 

image

image


projects/extension/src/containers/Options.tsx
```diff
@@ -186,6 +189,8 @@ const Options: React.FunctionComponent = () => {
                     />
                   )
                 })
+              ) : bg?.uiInterface.chains?.length === 0 ? (
```
tomaka, line 192:
```suggestion
              ) : bg?.uiInterface.smoldotCrashError ? (
```

Please use the `smoldotCrashError` and `onSmoldotCrashErrorChanged` functions, which were specifically designed to report about crashes.

tomaka:
I believe you need to use `useState`, and register a callback with `onSmoldotCrashErrorChanged` to modify the state.
We want the error to show up as soon as the error appears.

wirednkod:
> I believe you need to use `useState`, and register a callback with `onSmoldotCrashErrorChanged` to modify the state.
> We want the error to show up as soon as the error appears.

It is not needed. As soon as Smoldot crashes, the onChainsChanged is called, and the error is retrieved at once from the `smoldotCrashError`.

tomaka:
> As soon as Smoldot crashes, the onChainsChanged is called

No. This is your assumption. It might not be true.

projects/extension/src/containers/Options.tsx
```diff
@@ -26,6 +26,8 @@ const Options: React.FunctionComponent = () => {
   const [showModal, setShowModal] = useState<boolean>(false)
   const [bg, setBg] = useState<Background | undefined>()
 
+  const [clientError, setClientError] = useState<string>("")
```
tomaka, line 29:
```suggestion
  const [clientError, setClientError] = useState<string | undefined>(undefined)
```

Why not leave it like it is? `undefined` means "no error". A string, even if empty, means "there's an error".

projects/extension/src/containers/Options.tsx
```diff
@@ -116,10 +121,14 @@ const Options: React.FunctionComponent = () => {
       setNetworks([...networks.values()])
     }
 
-    cb = bg.uiInterface.onChainsChanged(refresh)
+    const cb = bg.uiInterface.onChainsChanged(refresh)
+    const errCb = bg.uiInterface.onSmoldotCrashErrorChanged(errorOccured)
```
tomaka, line 125:
```suggestion
    const errCb = bg.uiInterface.onSmoldotCrashErrorChanged(() => setClientError(bg.uiInterface.smoldotCrashError))
```

projects/extension/src/containers/Options.tsx
```diff
@@ -175,8 +184,10 @@ const Options: React.FunctionComponent = () => {
           {/** Networks section */}
           <TabsContent activeTab={activeTab}>
             <section>
-              {bg?.uiInterface?.smoldotCrashError ? (
-                <ClientError error={bg?.uiInterface?.smoldotCrashError} />
+              {clientError || bg?.uiInterface.smoldotCrashError ? (
+                <ClientError
+                  error={clientError || bg?.uiInterface.smoldotCrashError || ""}
+                />
```
tomaka, line 190:
```suggestion
              {clientError !== undefined ? (
                <ClientError
                  error={clientError}
                />
```

projects/extension/src/containers/Options.tsx
```diff
@@ -175,8 +184,10 @@ const Options: React.FunctionComponent = () => {
...
               ) : networks.length ? (
```
tomaka, line 191: I also don't think it makes sense to hide the list of chains if there is an error.
Again, you're assuming that "client error" is equal to "no chain", but I don't see a reason to put this assumption into the code.

tomaka (changes requested):

projects/extension/src/components/ClientError.tsx
```diff
@@ -0,0 +1,28 @@
...
+  return (
+    <div className="break-words block my-3 mx-6 justify-between pt-1.5 leading-4">
+      Smoldot light client has panicked with error:{" "}
+      <span
+        onClick={() => navigator.clipboard.writeText(error)}
+        className="cursor-pointer inline text-red-500 font-bold"
+      >
+        "{error}"
+      </span>
+      .Please copy the error message by clicking it or at the link below, and
```
tomaka, line 17:
```suggestion
      . Please copy the error message by clicking it or at the link below, and
```

projects/extension/src/components/ClientError.tsx
```diff
@@ -0,0 +1,28 @@
...
+      open a new bug issue at the
```
tomaka, line 18:
```suggestion
      open a new bug issue in the
```

projects/extension/src/components/ClientError.tsx
```diff
@@ -0,0 +1,28 @@
...
+      <a
+        className="underline inline pl-2 font-bold"
+        href="https://github.com/paritytech/smoldot/issues/new"
+        target="_blank"
+      >
+        Smoldot light client repository.
```
tomaka, line 24:
```suggestion
        smoldot repository.
```

projects/extension/src/containers/Options.tsx
wirednkod: This will lead to the error appearing ONLY if the popup or options page is open at the moment that the error takes place. If the error happens and after the user opens the pages then no error message will appear.

projects/extension/src/containers/Options.tsx
wirednkod: Im not sure why you call an "assumption" the case when the Smoldot panics (aka crashes). AFAI there are no networks running on smoldot light client when it crashes.

tomaka: My point is that it is not _necessarily_ correct.
The UI should not make any assumption about what happens in the background. If the background says that there is a crash error then the UI should display the crash error, if the background says there are networks then the UI should display the networks, and if the background says there are both then the UI should display both. It's as simple as that.

projects/extension/src/components/ClientError.tsx
```diff
@@ -0,0 +1,28 @@
...
+  return (
+    <div className="break-words block my-3 mx-6 justify-between pt-1.5 leading-4">
+      Smoldot light client has panicked with error:{" "}
+      <span
+        onClick={() => error && navigator.clipboard.writeText(error)}
+        className="cursor-pointer inline text-red-500 font-bold"
+      >
+        "{error}"
+      </span>
+      . Please copy the error message by clicking it or at the link below, and
```
tomaka, line 17: I don't really understand what you're trying to say with "clicking it or at the link below"

projects/extension/src/containers/Options.tsx
```diff
@@ -57,6 +60,7 @@ const Options: React.FunctionComponent = () => {
 
   useEffect(() => {
     if (!bg) return
+    setClientError(bg?.uiInterface.smoldotCrashError)
```
tomaka, line 63: This should happen after you call `onSmoldotCrashErrorChanged`, otherwise, if there's an `await` somewhere in between, the crash might happen during this `await`.

projects/extension/src/containers/Popup.tsx
```diff
@@ -76,6 +79,7 @@ const Popup: FunctionComponent = () => {
 
   useEffect(() => {
     if (!bg) return
+    setClientError(bg?.uiInterface.smoldotCrashError)
```
tomaka, line 82: Needs to happen after `onSmoldotCrashErrorChanged`

projects/extension/src/containers/Options.tsx
wirednkod: Maybe I was not clear before. There are 2 cases that we need to cover in order to show the error in the extension's UI.
The first case is when the UI is open and the error occurs (which is covered [by the onSmoldotCrashErrorChanged](https://github.com/paritytech/substrate-connect/pull/1228/files#diff-a091fcf55d424bd3b3ee40fced674890335fb4753e0fe0f88722e7b986cdac43R121-R123)), and when the error occurs while the Popup or Options are not open. 
The `setClientError` in the code here (or the way it was prior to the previous change), runs as soon as the page inits, and sets the outcome of the `smoldotCrashError` to the `clientError` state variable. If this code is not there, then there may be a crashed Client prior to opening any extension page/popup, but the user will never realize as the error will not appear

tomaka: I understand all what you're saying. This is not related to my comment.
If `bg.uiInterface.smoldotCrashError` is `undefined` when you call `setClientError`, then becomes defined, and then only you call `onSmoldotCrashErrorChanged`, you won't detect that the value has been defined.

tomaka (approved):

tomaka:
Also needs a CHANGELOG entry

PR #1238, opened 2022-09-05
Bump zombienet

wirednkod:
Rename zombienet poc to zombienet;
Bump zombienet and polkadot downad versions to latest;

### DIFF
--- a/.github/workflows/zombienet.yml
+++ b/.github/workflows/zombienet.yml
@@ -16,11 +16,11 @@ jobs:
           yarn build
       - name: Get zombienet
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/zombienet.yml
```diff
@@ -12,15 +12,15 @@ jobs:
       - uses: actions/checkout@v2
       - name: build
         run: |
-          yarn install --frozen-lockfile
```
tomaka, line 15: What's the difference?

tomaka (approved):

.github/workflows/zombienet.yml
wirednkod: There was notice while using it tha frozen lockfile is now deprecated and immutable should be used instead

PR #1254, opened 2022-09-20
Add block height in the popup

### LINKED ISSUE #1125: Transfer info from Options to Popup
We should proceed with the following 2 redesigns. 
- [x] **Show the latest block number in the popup**: The latest block number should appear inside the popup instead of the options page. This way the user will be aware that something is moving and the extension is alive even though there may be no apps connected.
- [ ] **Remove list of chains from the Options page**: There is no actual reason why the list of chains appear in the Options page. The information that appear in chain status (e.g. "synchronizing") is a hack and should not allow the user to rely on that (plus its not really useful). That could probably lead to a larger popup-screen with more information per chain instead of the Options Tab.

wirednkod:
Fixes #1125 

image


projects/extension/src/containers/Popup.tsx
```diff
@@ -142,10 +146,20 @@ const Popup: FunctionComponent = () => {
...
               return (
                 <>
-                  <div key={w.chainName} className="pl-6 flex text-lg">
-                    {networkIcon(w.chainName)}
+                  <div className="block mt-4">
+                    <div key={w.chainName} className="pl-6 flex text-lg">
+                      {networkIcon(w.chainName)}
+                    </div>
+                    <div className="pl-[4.5rem] text-sm flex pt-2">
+                      <span className="text-[#323232]">Last block</span>
```
tomaka, line 154:
```suggestion
                      <span className="text-[#323232]">Latest block</span>
```

projects/extension/src/containers/Popup.tsx
```diff
@@ -179,13 +193,21 @@ const Popup: FunctionComponent = () => {
...
-                  <div className="flex justify-between items-center w-full">
+                  <div className="block mt-4">
                     <div className="pl-4 flex text-lg justify-start">
                       {networkIcon(w.chainName)}
                       <span className="pl-2 text-[#616161]">
                         ({contents.length})
                       </span>
                     </div>
+                    <div className="pl-16 flex pt-2">
+                      <span className="text-[#323232]">Last block</span>
```
tomaka, line 204:
```suggestion
                      <span className="text-[#323232]">Latest block</span>
```

tomaka (approved):
I'm not completely sure about the esthetics, but I think that adding the current best is extremely important and is worth merging even if we change the design later.

Needs a CHANGELOG entry as well.

wirednkod:
> I'm not completely sure about the esthetics

Thank you.
There is a [Figma file split in issues (issue 1125)](https://www.figma.com/file/s8KFS3Anf7Wfk2axp8UdUY/Substrate-Connect-UI-v-0.1.4.-and-newer?node-id=801%3A7827) that contains the designs, and these are the ones I follow, after @Goranch3 provides them. Please add comments on the designs there, for Goran to follow accordingly.

Issue #882, opened 2022-03-21
Run smoldot in the content script instead of the extension


tomaka:
This issue covers one option that we could go for in order to fix https://github.com/paritytech/substrate-connect/issues/549

As the title says, the basic idea is to run smoldot in the content script of each individual web page rather than in the extension.
The extension would only be used as a "coordinator" between all these instances.

The design is a bit complicated, but it would work like this:

- When the extension starts, it starts its own instance of smoldot with the four well-known chains, and once these four chains are synced their database is extracted and stored in the local storage. It might be that this instance gets killed before it has managed to sync, but that's only an optimization so it's ok.
- When needed, the content script starts an instance of smoldot, and reports this to the extension. The extension assigns an id to this content script. It then sends back the ids of all the other existing content scripts that have a smoldot instance, and notifies all the other content scripts that have a smoldot instance about this new id. For better debuggability, this id could include parts of the URL of the page with the content script, but of course this must be done carefully.
- When the extension reports to a content script that another instance exists, the content script injects in smoldot the id of the other or all the other content scripts that have an instance of smoldot. This is done by using `/memory` multiaddresses (for example: `/memory/1`) and the [sudo_unstable_p2pDiscover](https://paritytech.github.io/json-rpc-interface-spec/api/sudo_unstable_p2pDiscover.html#sudo_unstable_p2pdiscover) JSON-RPC call.
    - Should smoldot refuse `sudo_unstable_p2pDiscover` JSON-RPC requests that contain a `/memory` address that don't come from the extension? It can't be used as an attack vector, but we could do it anyway just in case.
- When smoldot tries to connect to a `/memory/xxx` address, we send a message to the extension which either accepts or refuses the connection depending on whether the content script with that id is still alive.
    - After a connection to a `/memory/xxx` address has been established, when smoldot sends a message to that connection, it is sent the extension, which in turn sends the message to the proper content script.
    - If a content script gets killed, the extension notifies the other smoldot instances of the connection closure.
- When smoldot tries to connect to a non-`/memory` address, the content script first asks the extension for the authorization to connect. The extension can say yes or no. The actual connection is still done in the content script. The extension limits all the smoldots combined to 3 external connections maximum. Any further will be refused.
[...]

PR #1272, opened 2022-09-27
Run smoldot in the content script instead of the extension

tomaka:
cc https://github.com/paritytech/substrate-connect/issues/882

This PR is the first and main step towards converting the extension to manifest v3.

This PR considerably strips down the background script.
Instead of running smoldot in the background script and exchanging RPC requests between the web page, content script, and background script, we now run smoldot in the content script of each individual page.

This means that each web page that starts using the extension will have its own instance of smoldot.

The logs of the smoldot in the content script are directly shown in the web page's console, which makes debugging issues way easier.

### How it works

The messages between the web page and "the extension" are still the same. In other words, this PR is entirely backwards compatible from the point of view of applications.
However, these messages used to be simply transferred from the content script to the background script, while now they are directly handled by the content script.

The content script now talks with the background script in a different way. I've added a `protocol.ts` file in the `background` directory describing the messages that the content script and background script exchange with each other.

The main message is `'get-well-known-chain'`, which asks the background script to send back the chain spec and database of a well-known chain (i.e. Polkadot, Kusama, Westend, Rococo).

The content script then sends information to the extension about the list of chains that are currently running in the tab.

For well-known chains, the content script also periodically sends the database content in the extension (with a `'database-content'` message) so that the extension stores it. If there are multiple tabs running the same well-known chain, these database contents will override each other, but that's not a big deal.

### What remains to do

After this PR, remains to do:

- Remove the logs tab from the UI.
- Remove the "smoldot has crashed" UI component.
- Remove the button to force-disconnect a tab.
- Convert the background script to a service worker, and actually switch to manifest v3.


wirednkod:
This is not working with manifest v3 atm. I have not dag down to why just yet, but when adding the manifest v3 imple the service worker remains inactive and the apps do not connect anymore (as it happens in v2 with this impl).

Manifest v3 changes for file `manifest.json`:
```json
{
  "description": "Browser extension to manage substrate-based blockchain light clients",
  "homepage_url": "https://github.com/paritytech/substrate-connect",
  "name": "Substrate Connect Extension",
  "short_name": "substrate-connect",
  "version": "0.1.11",
  "manifest_version": 3,
  "permissions": ["notifications", "storage", "tabs", "alarms"],
  "background": {
    "service_worker": "background.js"
  },
  "action": {
    "default_title": "Substrate Connect",
    "default_popup": "popup.html"
  },
  "options_ui": {
    "page": "options.html",
    "open_in_tab": true
  },
  "content_scripts": [
    {
      "js": ["content.js"],
      "matches": ["http://*/*", "https://*/*"],
      "run_at": "document_start"
    }
  ],
  "icons": {
    "16": "./icons/icon-16.png",
    "32": "./icons/icon-32.png",
    "48": "./icons/icon-48.png",
    "128": "./icons/icon-128.png"
  },
  "content_security_policy": {
    "extension_pages": "script-src 'self' 'wasm-unsafe-eval'; object-src 'self'"
  }
}
```

tomaka:
I know that this doesn't work with manifest v3. Switching to service worker requires more changes in how events are handled in the service worker.
This PR is only the necessary preliminary change to make it work.

wirednkod (approved):

Polkadot-Forum:
This pull request has been mentioned on **Polkadot Forum**. There might be relevant details there:

https://forum.polkadot.network/t/peer-to-peer-connectivity-current-situation-and-roadmap/476/12

PR #1280, opened 2022-09-30
Switch to using chrome.runtime.sendMessage/onMessage

tomaka:
cc https://github.com/paritytech/substrate-connect/issues/1278

Next step towards switching to manifest v3.

We now no longer maintain `MessagePort`s open. Instead, we just send messages and wait for answers.


projects/extension/src/background/protocol.ts
```diff
@@ -52,7 +52,6 @@ export type ToContentScript = ToContentScriptWellKnownChain
 // Response to a {ToExtensionGetWellKnownChain}
 export interface ToContentScriptWellKnownChain {
   type: "get-well-known-chain"
-  chainName: string
```
tomaka, line 55: It was necessary in order to match the proper response with the proper request. Removed as no longer necessary.

wirednkod (approved):

PR #1285, opened 2022-10-05
Fix source map warnings

sesh92:
These changes were made to generate `sourcesContent` in your source map files, it will allow avoiding the warnings which you tried to fix [here](https://github.com/paritytech/substrate-connect/issues/1196).

Docs:
[ts config inlineSources ](https://www.typescriptlang.org/tsconfig#inlineSources)

### DIFF
--- a/tsconfig.base.json
+++ b/tsconfig.base.json
@@ -4,7 +4,8 @@
     "baseUrl": ".",
     "composite": true,
     "declaration": true,
-    "inlineSourceMap": true,
+    "inlineSources": true,
+    "sourceMap": true,
     "paths": {
       "@substrate/connect-extension-protocol": [
         "./packages/connect-extension-protocol"


wirednkod (approved):

wirednkod:
I have tried what you propose in the PR but I can see that the created map files (and this was the initial source of the issue back then) are pointing to the `.ts` files, which will again lead to missing file error:
e.g. `index.js.map`:
```
{"version":3,"file":"index.js","sourceRoot":"","sources":["../../src/index.ts"],"names":[],"mappings":";AAAA;;;;;;;;;;;;;;
```
(see the `.....src/index.ts` filepath in the sources array)

wirednkod:
Will [remove source maps](https://github.com/paritytech/substrate-connect/pull/1288) in order to avoid these warnings

MOZGIII:
> I have tried what you propose in the PR but I can see that the created map files (and this was the initial source of the issue back then) are pointing to the `.ts` files, which will again lead to missing file error: e.g. `index.js.map`:
> 
> ```
> {"version":3,"file":"index.js","sourceRoot":"","sources":["../../src/index.ts"],"names":[],"mappings":";AAAA;;;;;;;;;;;;;;
> ```
> 
> (see the `.....src/index.ts` filepath in the sources array)

You misinterpret the error.

If is perfectly fine for a `.map` file to point to `.ts` file (chack almost any of the packages at npm).

The problem was, in fact, caused by the absence of the `sourcesContent` field at the `.map` file, which was leading to the issues with the source map resolution. This is expected, and the solution is to enable `inlineSources` (not `inlineSourcesMap` though, that's a different thing!) to add the `sourcesContent` to the `.map`s.

`inlineSourcesMap` in turn will just embed the source into the `.js` file, and that's irrelevant fo the issue. `inlineSources` adds the source to the `.map` file, but keeps the `.map` files separate. This allows proper bundle optimization and makes source maps work.

For completeness, enabling both `inlineSources` and `inlineSourcesMap` is a kind of edge case where the source map will be embedded into the minimized `.js` as a comment, and it will contain the source code of the original `.ts` within itself. Doesn't make a lot of sense really.

PR #1288, opened 2022-10-07
Remove source maps

wirednkod:


tomaka (approved):

MOZGIII:
Please see an explainer at https://github.com/paritytech/substrate-connect/pull/1285#issuecomment-1271588986.
Proper source maps support is right there. We have already tested that solution and it works flawlessly.
Please don't make things worse.

wirednkod:
> Please see an explainer at [#1285 (comment)](https://github.com/paritytech/substrate-connect/pull/1285#issuecomment-1271588986). Proper source maps support is right there. We have already tested that solution and it works flawlessly. Please don't make things worse, let's just merge that code.

I think that removing the source makes things better rather than worse for the production code, and I think its preferable to keep the amount of files and size of it to minimum (removing it, reduces amount of files from 62 -> 48 and save some hundreds of KB). 

The sample application you provided for supporting your comment is a CRA app. I have tested these changes in non-CRA apps/demos and were working without warnings. 

There is an explanation [here](https://github.com/mswjs/msw/issues/1030) about CRA 5 and how the integrated in CRA webpack handles the sourcemaps. There is proposed a workaround that you can run in order to avoid warnings but in my opinion having source maps in production is kind-of-redundant and adds files and size to the final `dist`.

MOZGIII:
> I think that removing the source makes things better rather than worse for the production code, and I think its preferable to keep the amount of files and size of it to minimum (removing it, reduces amount of files from 62 -> 48 and save some hundreds of KB).

You are wrong. The production code is shipped in bundles, and the way this works is libraries provide source maps, and whoever is using the library decides whether they want or don't want to use source maps.

> The sample application you provided for supporting your comment is a CRA app. I have tested these changes in non-CRA apps/demos and were working without warnings.

I don't know what you're talking about, I didn't provide anything, this is the first time I've chimed in on this issue.

I'm also not just talking from a perspective on coming up with a hack, I just know how these things are supposed to work currently. In CRA or non-CRA, the source maps with properly filled source content will work.

> There is an explanation [here](https://github.com/mswjs/msw/issues/1030) about CRA 5 and how the integrated in CRA webpack handles the sourcemaps. There is proposed a workaround that you can run in order to avoid warnings but in my opinion having source maps in production is kind-of-redundant and adds files and size to the final `dist`.

You know, CRA is not really intended for the experts, and sometimes (quite often really) the workarounds they provide are misused by people when a really good solution is available. Properly supporting source maps on the source level is a good solution. Removing them is a bad solution. How we can tell it's bad? We invented them out of nothing and made a whole system to make them work across the cross-cutting packages and build systems. This took a lot of work, but it has still happened. Nullifying this whole effort just cause you don't fully understand how it all works is pathetic. Especially when the working solution has been contributed by the community.

wirednkod:
> I don't know what you're talking about, I didn't provide anything, this is the first time I've chimed in on this issue.

Apologies about this. I was referring to an example app provided by the person who opened #1285. Got a bit confused there.

> Nullifying this whole effort just cause you don't fully understand how it all works is pathetic. Especially when the working solution has been contributed by the community.

I have removed _for now_ the source maps in order to cover [a specific case](https://github.com/paritytech/polkadot-staking-dashboard/issues/260) concerning warnings, that was opened. 
My intention, is to investigate further the case of source maps, based on the solution provided in #1285 in order to "understand how it all works". 

I think I should mention at this point, that is really not polite and not called for, the accusations of "nullifying efforts" and "being pathetic". Especially when the intention (as said above) is to investigate the possible solutions (including the one proposed).

MOZGIII:
> I think I should mention at this point, that is really not polite and not called for, the accusations of "nullifying efforts" and "being pathetic". Especially when the intention (as said above) is to investigate the possible solutions (including the one proposed).

Wouldn't be this way if were more transparent about the intentions. I apologize too.

> I have removed _for now_ the source maps in order to cover [a specific case](https://github.com/paritytech/polkadot-staking-dashboard/issues/260) concerning warnings, that was opened.

Yep, and I'm more confused about this PR, because https://github.com/paritytech/substrate-connect/pull/1285 would solve that one as well.

> My intention, is to investigate further the case of source maps, based on the solution provided in #1285 in order to "understand how it all works".

Inf you plan to put in more work here - it's all good.

PR #1292, opened 2022-10-10
Minor fixes that were returning error

wirednkod:


tomaka:
No, we don't want to make things optional.

The reason why you're having these errors is the 4th line of the changelog here: https://github.com/paritytech/smoldot/blob/main/bin/wasm-node/CHANGELOG.md#fixed

In other words, you're seeing these errors only because you have smoldot 0.7.1 installed. This fix should be part of the "Update smoldot to 0.7.1" PR.

wirednkod:
Thank you - though I identified these errors initially when I was testing version 0.7.1 🤔

PR #1315, opened 2022-10-19
Adding Astar Shiden and GM icons

gluneau:
- Added astar shiden and gm glyphs to fonts using https://fontforge.org/
- Added all 3 liga entries in the fonts.
- Added the 3 parachain names to the array in IconWeb3.tsx

Tested by loading an unpacked extension:
image

image

PR #1318, opened 2022-10-21
Only show icon for well-known chains

tomaka:
This PR makes it so the extension knows what is a well-known chain and what isn't.

This way, we can show the icon of a chain only if it is well-known.

This is important for security. Right now we just do `if icon.get(chainName)`, but given that `chainName` is untrusted user input, anyone can pretend that they're connecting to for example Polkadot while in fact connecting to a chain that isn't Polkadot.

@wirednkod Could you push on this PR and make it so that the `IconWeb3` also shows `?` if `isWellKnown` is `false`? I don't know how to pass values around with React.

### DIFF
--- a/projects/extension/src/background/index.ts
+++ b/projects/extension/src/background/index.ts
@@ -106,6 +106,7 @@ chrome.runtime.onMessage.addListener(
 
             chains.push({
               chainId: message.chainId,
+              isWellKnown: message.isWellKnown,
               chainName: message.chainSpecChainName,
               isSyncing: false,
               peers: 0,
--- a/projects/extension/src/background/protocol.ts
+++ b/projects/extension/src/background/protocol.ts
@@ -44,6 +44,7 @@ export interface ToExtensionReset {
 export interface ToExtensionAddChain {
   type: "add-chain"
   chainId: string
+  isWellKnown: boolean
   chainSpecChainName: string
 }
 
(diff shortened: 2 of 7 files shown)


wirednkod (approved):

tomaka:
Thank you

projects/extension/src/components/IconWeb3.tsx
```diff
@@ -10,10 +11,13 @@ interface Props {
 const hasGlyph = (string: string) =>
   ["kusama", "polkadot", "westend", "rococo"].indexOf(string) > -1
```
gluneau, line 12: Will I be allowed to parachains like Astar, Shiden & GM?

My previous PR was reverted.
https://github.com/paritytech/substrate-connect/pull/1315

tomaka: No, only these 4 relay chains have an icon because we have explicitly integrated them into substrate-connect.
We don't want to curate a list of chains like PolkadotJS does.

The reason why these 4 relay chains are hardcoded in the extension is because, specifically for relay chains, it makes syncing them much faster. And because this is a Parity product, we're also skewed towards putting the Parity chains in the code. This optimization, however, is irrelevant for parachains. Hardcoding parachains in the code brings absolutely no technical benefit.
So basically it's these 4 chains and no other.

PR #1389, opened 2023-02-17
@substrate/smoldot-light@0.7.9 -> smoldot@0.7.11

josepot:


wirednkod (approved):
The PR is fine, though I'm wondering why the package changed from @substrate/smoldot to smoldot. Is rhere a reason?

josepot:
> The PR is fine, though I'm wondering why the package changed from @substrate/smoldot to smoldot. Is rhere a reason?

The reason being that now the fixes and new development are being published on the `smoldot` npm package. We will have to make a decision of what we do now with `@substrate/smoldot-light`. I think that there are only 2 options:

1) We set something up so that it mirrors the `smoldot` package
2) We use [`npm-deprecate`](https://docs.npmjs.com/cli/v8/commands/npm-deprecate) on `@substrate/smoldot-light` with a warning message that points users to the `smoldot` package.

Option 2 seems the simplest.

PR #1400, opened 2023-03-21
fix: remove dynamic imports

josepot:
Apparently not all bundlers (vite, parcel, etc) are able to gracefully handle dynamic imports on package's code OOTB. So, rather than creating particular solutions for each bundler, I think that's simpler to get rid of the dynamic imports all together. Will this increase the bundle size on production builds? Yep, a bit. However, we started doing this back in the days when the spec files were huge, and thankfully that's no longer the case, so this is IMO the lesser evil.


packages/connect/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Changed
+
+- Stop using dynamic imports in order to import the "well-known" specs, due to the fact that not all bundlers are able to handle dynamic imports gracefully on package's code.
+
```
wirednkod, line 8: This is no needed information for the users. I think we can remove it

wirednkod (changes requested):

packages/connect/CHANGELOG.md
josepot: Sorry, I'm not following. What information is not needed? What is that you think that it should be removed? Could you please propose an alternative text?

wirednkod, line 8: Sure thing. I do not think we need to add this at all - as its internal things of SC that are not need to be known to the user;
```suggestion
```

wirednkod: No worries - Lets just keep that as is 👍🏽

wirednkod (approved):

Issue #1405, opened 2023-03-24
Document how Substrate Connect Extension works


wirednkod:
Following [this comment](https://github.com/paritytech/substrate-connect/issues/1404#issuecomment-1482932013), a sentence should be added, to the following places in order to explain in a couple of sentences how Substrate Connect Extension shall work:

- [x] README of the repo [#1](https://github.com/paritytech/substrate-connect/blob/main/README.md), [#2](https://github.com/paritytech/substrate-connect/blob/main/projects/extension/readme.md), - [PR](https://github.com/paritytech/substrate-connect/pull/1407);
- [x]  [Chrome store's extension](https://chrome.google.com/webstore/detail/substrate-connect-extensi/khccbhhbocaaklceanjginbdheafklai?hl=en-GB);
- [x]  [Mozilla addons' extension](https://addons.mozilla.org/en-US/firefox/addon/substrate-connect/);
- [x]  Substrate Documentation [#1](https://docs.substrate.io/fundamentals/light-clients-in-substrate-connect/) ([PR](https://github.com/substrate-developer-hub/substrate-docs/pull/1918)), [#2](https://substrate.io/developers/substrate-connect/) ([PR](https://github.com/paritytech/substrate-website/pull/328));

PR #1407, opened 2023-03-27
Add extra info in documentation about Substrate Connect Extension

wirednkod:
Concerning #1405


README.md
```diff
@@ -84,3 +84,17 @@ yarn dev:burnr
 
 
 ## [Deployments and releases](./DEPLOY-RELEASE.md)
+
+
+## Substrate Connect Extension
+
+A Browser extension that keeps the latest state of well known substrate-based chains' specs and bootnodes (Polkadot, Kusama, Rococo, Westend) synced across tabs - using Substrate Connect and Smoldot light client;
```
josepot, line 91: > A Browser extension that keeps the latest state of well known substrate-based chains' specs and bootnodes (Polkadot, Kusama, Rococo, Westend) synced across tabs

I don't think that this is true anymore,. Unfortunately, the extension no-longer keeps things in sync across tabs. The extension does let the user know how the light-client is being used across the different tabs by displaying open connections and the tabs that are using the light-client, etc. However, it's not accurate to say that the extension keeps the latest state of well known substrate-based chains synced across tabs. That used to be the case, but unfortunately it no longer is.

josepot (approved):

PR #1413, opened 2023-04-10
Switch to DNS name based bootnodes for Rococo

bakhtin:
Update chainspec for Rococo as we migrated from ephemeral IP addresses to a more stable DNS name based addressing scheme.


wirednkod:
Thank you @bakhtin 
Quick question - shouldn't there be also `wss` instead of just `ws` though?

wirednkod:
@bakhtin Any feedback on this? ☝🏽

wirednkod (approved):
lgtm.
I have talked with @PierreBesson and the `wss` should be added in this list as well

PR #1471, opened 2023-07-11
Create tests in zombienet for testing the chainspecs

wirednkod:
Create a zombienet "dump" test that allows us to test light client against all the chainSpecs.
This way every time the chainspec states are updated (daily) this test will ensure that they are correct and are not failing against the actual network


josepot (approved):
Thanks!

PR #1512, opened 2023-08-16
feat: run smoldot in the background

kratico:
Closes https://github.com/paritytech/capi/issues/58

- run smoldot in the background service worker script
- use `substrate-connect` in the background instead of plain Smoldot
- add `potentialRelayChains` and `databaseContent` to `AddChain` and `AddWellKnownChain`
- move UI (options and popup) related state updates to `trackChains.ts` module
- add keep alive messages from the content script to background script
- updated message passing APIs to [`chrome.runtime.connect`](https://developer.chrome.com/docs/extensions/reference/runtime#method-connect)


packages/connect/src/connector/smoldot-light.ts
```diff
@@ -238,6 +243,8 @@ export const createScClient = (config?: Config): ScClient => {
   const addWellKnownChain: AddWellKnownChain = async (
     supposedChain: WellKnownChain,
     jsonRpcCallback?: (msg: string) => void,
+    potentialRelayChains?: Chain[],
```
josepot, line 246: since "well known chains" are all relay-chains, why would we need this? 🤔

projects/extension/src/background/trackChains.ts
```diff
@@ -0,0 +1,250 @@
...
+
+  const chainInfo: ChainInfo = {
+    isSyncing: true,
+    peers: 0,
+  }
+
+  const chain = await scClient.addChain(
+    chainSpec,
+    (rawMessage) => {
+      // const channel = chainOptions.channel(channelId, (rawMessage) => {
```
ryanleecode, line 53: Extra comment?

projects/extension/src/background/enqueueAsyncFn.ts
```diff
@@ -0,0 +1,4 @@
+let guard = Promise.resolve()
+export const enqueueAsyncFn = (fn: () => Promise<void>) => {
+  return (guard = new Promise((resolve) => guard.then(fn).finally(resolve)))
```
josepot, line 3:
```suggestion
  return (guard = new Promise((resolve) => guard.then(fn, (e) => {
    console.log("An unhandled execption happened on 'enqueueAsyncFn'")
    console.error(e);
  }).finally(resolve)))
```

packages/connect/src/connector/smoldot-light.ts
kratico: will remove because this is not needed for wellKnownChains

josepot (approved):
Thanks a lot for this @kratico !!

PR #1525, opened 2023-09-05
feat: add @substrate/connect/worker

kratico:
Closes https://github.com/paritytech/capi/issues/47

Add `./worker` subpath to `@substrate/connect`.
Then, `@substrate/connect/worker` is a WebWorker that runs smoldot (see [smoldot usage with a web worker](https://smol-dot.github.io/smoldot/doc-javascript/#md:usage-with-a-worker))

Test it with `vite` [WebWorkers](https://v3.vitejs.dev/guide/features.html#web-workers)

```ts
import { createScClient } from "@substrate/connect"
import ScWorker from "@substrate/connect/worker?worker"

createScClient({
  embeddedNodeConfig: {
    workerFactory: () => new ScWorker(),
  },
})
  .addWellKnownChain("polkadot", (response) =>
    console.log({ response }),
  )
  .then((chain) => {
    chain.sendJsonRpc(
      JSON.stringify({
        jsonrpc: "2.0",
        id: "1",
        method: "chainHead_unstable_follow",
        params: [true],
      }),
    )
  })
```

image

Test it with Node

```js
// main.mjs
import { createScClient, WellKnownChain } from "@substrate/connect"
import { Worker } from "node:worker_threads"

createScClient({
  embeddedNodeConfig: {
    workerFactory: () => new Worker("./worker.mjs"),
  },
})
  .addWellKnownChain(WellKnownChain.polkadot, (response) =>
    console.log({ response }),
  )
  .then((chain) => {
    chain.sendJsonRpc(
      JSON.stringify({
        jsonrpc: "2.0",
        id: "1",
        method: "chainHead_unstable_follow",
        params: [true],
      }),
    )
  })

```

```js
// worker.mjs
import "@substrate/connect/worker"
```

### DIFF
--- a/packages/connect/src/connector/smoldot-light.ts
+++ b/packages/connect/src/connector/smoldot-light.ts
@@ -36,9 +36,15 @@ const getClientAndIncRef = (config: Config): Promise<Client> => {
     else return Promise.resolve(clientPromise)
   }
 
-  const newClientPromise = getStart().then((start) =>
-    start({
-      portToWorker: config.portToWorkerFactory?.(),
+  const newClientPromise = getStart().then((start) => {
+    let portToWorker: MessagePort | undefined = undefined
+    if (config.workerFactory) {
+      const { port1, port2 } = new MessageChannel()
+      config.workerFactory().postMessage(port1, [port1])
+      portToWorker = port2
+    }
+    return start({
+      portToWorker,
       forbidTcp: true, // In order to avoid confusing inconsistencies between browsers and NodeJS, TCP connections are always disabled.
       forbidNonLocalWs: true, // Prevents browsers from emitting warnings if smoldot tried to establish non-secure WebSocket connections
       maxLogLevel: 9999999, // The actual level filtering is done in the logCallback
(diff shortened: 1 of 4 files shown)


packages/connect/src/connector/smoldot-light.ts
```diff
@@ -134,6 +135,8 @@ export interface Config {
    * value will be used.
    */
   maxLogLevel?: number
+
+  portToWorkerFactory?: () => MessagePort
```
josepot, line 139: I was thinking that perhaps it would be slightly better to have something like:
```ts
workerFactory?: () => Worker
```

and then on the `start` method we could do something like:

```ts
const newClientPromise = getStart().then((start) => {
  let portToWorker: MessagePort | undefined = undefined
  if (config.workerFactory) {
    const { port1, port2 } = new MessageChannel()
    config.workerFactory().postMessage(port1, [port1])
    portToWorker = port2
  }
  start({
    portToWorker,
    ...otherStuff
  });
})
```

That way the public API would be a lot simpler. Using the example of the description of the PR, it would look like this:

```ts
createScClient({
  embeddedNodeConfig: {
    workerFactory: () => new ScWorker()
  },
})
```

WDYT? 🤔

kratico: yes this is a better public API

josepot (approved):

PR #1529, opened 2023-09-07
feat: upgrade to smoldot@2.0.0

kratico:


josepot (approved):
Thanks @kratico !

tomaka:
This paragraph should have been removed IMO: https://github.com/paritytech/substrate-connect/blob/b7c777cac5af447f3a4d77f26eb178c095562477/packages/connect-extension-protocol/src/index.ts#L282-L284

projects/extension/src/background/index.ts
```diff
@@ -259,21 +259,16 @@ chrome.runtime.onConnect.addListener((port) => {
           try {
             chain.sendJsonRpc(msg.jsonRpcMessage)
           } catch (error) {
-            // As documented in the protocol, malformed JSON-RPC requests are silently ignored.
```
tomaka, line 262: This was a subtle hint that the behavior in this code matches the protocol, and so if you change the behavior it doesn't match the protocol anymore

PR #1532, opened 2023-09-07
chore: use typescript for zombie net tests

ryanleecode:
As per title, this updates the zombie net tests to use typescript. For some reason the zombienet cli does not transpile `utils.ts`, even though when using `tsc` on any of the test files it would automatically transpile it; as a result, i manually run tsc on `utils.ts` in the workflow.


.github/workflows/zombienet.yml
```diff
@@ -29,6 +29,7 @@ jobs:
       - name: Run test
         run: |
           export PATH=$(pwd):$PATH
+          tsc .zombienet-tests/utils.ts || true
```
kratico, line 32: Interesting finding on zombienet.

kratico (approved):
LGTM

PR #1539, opened 2023-09-11
feat: simplify tooling extension with vite

kratico:
Use vite to bundle the extension scripts.

By default, Vite targets browsers which support the [native ES Modules](https://caniuse.com/es6-module), [native ESM dynamic import](https://caniuse.com/es6-module-dynamic-import), and [import.meta](https://caniuse.com/mdn-javascript_operators_import_meta), this presents a challenge because
- `import` statement not supported in content scripts
- `import(...)` or dynamic imports are not supported in the ServiceWorker

Therefore, the extension scripts/UIs are build with 3 vite bundles
- `vite.ui.config.js`, which bundles the popup and options UI
- `INPUT=content vite build --config vite.script.config.js`, which bundles the content script in a single file
- `INPUT=background vite build --config vite.script.config.js`, which bundles the background script in a single file
Note: for content and background scripts, dynamic imports are inlined and because the output is a single file without code-splitting, no `import` declarations are needed.

This PR also removes any references and dependencies related to `webpack` and `babel`.


josepot (approved):
Thanks a lot for this!

PR #1560, opened 2023-09-26
feat: improve extension background script bundle size

kratico:
Reduce extension background script bundle size by
- loading chainspecs from extension assets
- importing `@substrate/connect` from a different chunk
- updating `@substrate/connect` to use dynamic imports for well known chainspecs

Note: extension [service worker does not support dynamic imports](https://github.com/w3c/webextensions/issues/212), but `@substrate/connect` chainspecs are not used by the extension, so the dynamic import path will not cause a runtime error in the extension service worker.

Bundle before this update
image

Bundle after this update
image

Future improvements 
- `@substrate/connect` `createScClient` dynamically imports `smoldot` or `smoldot/no-auto-bytecode` by doing a check at runtime so it can not be statically removed from the bundle. Creating and using a function that only dynamically imports `smoldot/no-auto-bytecode` will tree shake and remove smoldot byte code from the bundle.


josepot (approved):

wirednkod:
Nice one @kratico . The point is if the final package contains ANY file above 4 MB. If so then mozilla will deny it (meaning its not just the bg script but any file above 4 MB will fail)

wirednkod (approved):

PR #1563, opened 2023-09-26
Replace chain Ids with chain Names for well known chains in Popup

wirednkod:
At the moment (extension v0.2.11) the Popup shows the `id` field of the well known chain instead of the `name` of the chain (`ksmcc3` instead of `Kusama`).

image

This PR introduces an easy way, using a map for mapping the `id` to the `name` of each chain;
A more "complicated" (not really - just bit more than this one) - is to parse the Chainspec of the well known chain once more and retrieve the chain name.

Let me know if this solution is enough or you prefer to go for the JSON.parse method;

The outcome that this PR will deliver is:
image

Also fixes the icons and Network names in Options page
From:
image

To:
image


kratico:
It would be better to fix it here

https://github.com/paritytech/substrate-connect/blob/47f4acfb0d903c98eef49d05d19b9f992acb97fd/projects/extension/src/background/index.ts#L218-L220

something like

```
export const wellKnownChainNames: Record<string,string> = {
  westend2: "Westend",
  polkadot: "Polkadot",
  ksmcc3: "Kusama",
  rococo_v2_2: "Rococo",
}

// ...

 chainName: isWellKnown 
   ? wellKnownChainNames[msg.chainName] ?? msg.chainName 
   : (JSON.parse(msg.chainSpec).name as string), 
```

projects/extension/src/background/index.ts
```diff
@@ -6,7 +6,7 @@ import { enqueueAsyncFn } from "./enqueueAsyncFn"
 import { trackChains } from "./trackChains"
 
 import * as environment from "../environment"
-import { PORTS } from "../shared"
+import { PORTS, wellKnownChainNames } from "../shared"
```
kratico, line 9: Given that `wellKnownChainNames` is not shared by other contexts (content script or extension pages), let's remove it from `shared` and keep it local to this module

kratico (approved):

PR #1605, opened 2023-11-11
fix: code splitting problem

josepot:
The dynamic import of `smoldot` used on the `smoldot-light.ts` file was pointless due to the fact that the class `QueueFullError` was not being dynamically imported. As a result of this,  all users of `@substrate/connect` end up with the smoldot binary included in the main-bundle of their projects, this is specially annoying for those ones who have the extension installed.

This PR is a pragmatic fix to this issue.


kratico (approved):
LGTM

PR #1609, opened 2023-11-13
fix(extension): @substrate/connect dynamic import

kratico:
Removes `@substrate/connect` lib from the extension because it uses a dynamic import to create the `smoldot` client.
Dynamic imports are not supported in the extension background script ServiceWorker. 

Note: downgraded `smoldot` to `2.0.6` until this [issue](https://github.com/smol-dot/smoldot/issues/1330) is fixed.

### DIFF
--- a/projects/extension/src/background/addChain.ts
+++ b/projects/extension/src/background/addChain.ts
@@ -1,0 +1,150 @@
+import type { Chain, Client } from "smoldot"
+import {
+  start,
+  QueueFullError,
+  JsonRpcDisabledError,
+  AlreadyDestroyedError,
+  CrashError,
+} from "smoldot"
+
+let clientReferences: number = 0
+let client: Client | null = null
+const getClientAndIncRef = () => {
+  if (client) {
+    clientReferences++
+    return client
+  }
+
+  client = start({
+    forbidTcp: true, // In order to avoid confusing inconsistencies between browsers and NodeJS, TCP connections are always disabled.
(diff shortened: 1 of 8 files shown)


projects/extension/src/background/addChain.ts
kratico, line 1: This file is a slimmed version of [packages/connect/src/connector/smoldot-light.ts](https://github.com/paritytech/substrate-connect/blob/a03b832827e4fb1252df316307bc3c13fc691450/packages/connect/src/connector/smoldot-light.ts)

projects/extension/vite.script.config.js
```diff
@@ -14,12 +13,7 @@ export default defineConfig(({ mode }) => ({
       },
       output: {
         entryFileNames: (chunk) => `${chunk.name}.js`,
-        inlineDynamicImports: !isBackground,
-        manualChunks: isBackground
-          ? {
-              "substrate-connect": ["@substrate/connect"],
-            }
-          : undefined,
```
kratico, line 22: The background script size is 2.8Mb (< 4Mb JS file size limit required for any extension script), so there is no need to do `manualChunks`

josepot (approved):
This is a pragmatical solution, so let's get it in. For sure!

However, I would like to eventually improve the bundling of substrate-connect, so that we can use a flag that hints the bundler whether to use a dynamic import or not. That would allow us to publish a version of substrate-connect without dynamic imports in a specific subpath export. So, that we don't have to always remember to copy/paste the changes into the background of the extension.

Anyways, we definitely have bigger fish to fry ATM, so this is perfectly fine for now. Thanks @kratico !

PR #1645, opened 2023-11-29
feat: add @substrate/connect-known-chains

kratico:
Move well known chain specifications to `@substrate/connect-known-chains`

Using the main entry, unused chainspecs are removed, for example with `vite` the following is bundled
image


josepot (approved):
Thanks @kratico !

PR #1647, opened 2023-11-29
feat(connect): use `@substrate/connect-known-chains`

kratico:
Use `@substrate/connect-known-chains` in `@substrate/connect`

Note: to support sub-path imports in `@substrate/connect`
- the `tsconfig.json` was updated with `"moduleResolution": "Bundler"`
- and, the transpiling and bundling has been updated to use `tsup` (instead of `tsc`)


packages/connect/package.json
```diff
@@ -7,21 +7,22 @@
   "publishConfig": {
     "access": "public"
   },
-  "main": "dist/cjs/index.js",
-  "module": "dist/mjs/index.js",
-  "types": "dist/mjs/index.d.ts",
+  "type": "module",
```
josepot, line 10: ufff, I'm a bit scared to make this change, TBH...

PR #1717, opened 2024-01-04
feat: use pnpm

kratico:


josepot (commented):
There seems to be an issue with zoombienet... could you please look into that?

PR #1835, opened 2024-02-13
feat: secure storage

ryanleecode:


marshacb (approved):
lgtm!

packages/secure-storage/src/secure-storage.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export type SecureLocalStorage = {
+  setItem: (typeof localStorage)["setItem"]
+  getItem: (typeof localStorage)["getItem"]
+  removeItem: (typeof localStorage)["removeItem"]
+  clear: (typeof localStorage)["clear"]
+}
+
+export const createSecureLocalStorage = (
+  password: Uint8Array,
```
kratico, line 16: it would be better to pass the `cipher`

kratico (commented):
This package should be moved inside the substate-connect extension.

The secure storage is meant to store an encrypted seed with a password using the extension storage not the browser `localStorage`.

Issue #1894, opened 2024-02-27
Scaffold Basic Extension UI and RPC to submit a transaction


kratico:
```[tasklist]
### Tasks
- [x] fetch selected account balance
- [x] handle form fields and submission with react
- [x] fetch accounts from extension
- [ ] validate destination address
- [x] use PAPI to encode the transaction calldata
- [x] transfer should trigger an extension popup that signs the transaction
- [x] extract transaction submission into a hook
- [x] follow transaction submission events until it is finalized
- [x] share observable client across hooks/components
- [ ] https://github.com/paritytech/substrate-connect/pull/1914
- [ ] add block number
- [ ] transaction hash
- [ ] switch between chains
- [ ] add transaction fee
```

PR #1923, opened 2024-03-01
feat(lc-dapp): transfers

ryanleecode:
Does a transfer and reports the transaction status in the UI.

#1894


kratico (changes requested):

kratico (approved):
I'll do a 2nd pass on a different PR to check if the RxJS observables can be simplified.

PR #1941, opened 2024-03-05
feat: release please

ryanleecode:


kratico (approved):
Approving because it's not possible to test a new workflow if it's not on `main`

PR #1971, opened 2024-03-12
Bump @substrate/connect to 0.8.8

kratico:
- Bump @substrate/connect to 0.8.8
- Bump @substrate/connect-known-chains to 1.1.1
- Bump @substrate/light-client-extension-helpers to 0.0.4
- Bump Substrate connect extension to 0.4.2


.ensure-resolution.mjs
kratico, line 1: PNPM workspaces already handle this, so it's redundant to have this check

PR #1984, opened 2024-03-14
feat: add password encryption

kratico:
- add password encryption using a Keystore following [EIP-2335](https://eips.ethereum.org/EIPS/eip-2335)
- add welcome screen that is open when the extension is installed
  (Note: in future PRs the welcome screen should also be open from the browser action icon)
- add typed storage to save the password keystore

An example of a stored password
image

The basic welcome screen
image


projects/wallet-template/src/background/createBackgroundRpc.ts
```diff
@@ -37,23 +41,50 @@ const keyset = {
...
+        ),
+      )
+
+      // TODO: re-encrypt accounts with new password
     },
-    isLocked() {
-      return !currentPassword
+    async setup(password: string) {
+      assert(!(await getKeystore()), "keyring is already setup")
+      await setKeystore(keystoreV4.create(password, randomBytes(32)))
```
kratico, line 82: @ryanleecode `randomBytes(32)` is the encrypted secret.
With the password it can be decrypted and used as a key to encrypt seed phrases or private keys.

WDYT?

ryanleecode: better to use the key incase the cipher changes in the future

kratico: for the above alternatives, when there is a password change
- keeping the key does not need account secret re-encryption because the key hasn't changed
- using cipher, anything that was encrypted with the cipher needs to be re-encrypted

ryanleecode:
i saw in that EIP, there is a sdk for the keystore made by chainsafe. chainsafe is typically pretty reputable i think. should we use their keystore or roll our own?

https://github.com/ChainSafe/bls-keystore

kratico:
> i saw in that EIP, there is a sdk for the keystore made by chainsafe. chainsafe is typically pretty reputable i think. should we use their keystore or roll our own?
> 
> https://github.com/ChainSafe/bls-keystore

I checked that and these are some things that motivate a separate implementation
- it's specific to ethereum and to encrypt account private keys
- to store an arbitrary secret, let's say the seed for a keyset, some fields like `path`, `publicKey` are not needed
- the `cipher` does not support `xsalsa20-poly1305`
- to minimize extra deps, it can be implemented with the audited noble-hashes and noble-ciphers that we already use

To store the wallet accounts/keysets, the keystore can be extended with a `meta` prop that describes derivations paths with their public keys, chain ids, etc.

kratico:
BTW, [this](https://github.com/ethers-io/ethers.js/blob/ad5f1c5fc7b73cfb20d9012b669e9dcb9122d244/src.ts/wallet/json-keystore.ts#L298-L319) is another implementation for the previous keystore version from ether.js

ryanleecode (approved):

kratico:
Related to https://github.com/paritytech/substrate-connect/pull/1984

Polkadot-Forum:
This pull request has been mentioned on **Polkadot Forum**. There might be relevant details there:

https://forum.polkadot.network/t/parity-tech-update-for-march/7226/1

PR #1994, opened 2024-03-15
chore: add type checking

ryanleecode:


examples/light-client-extension-helpers-extension/tests/fixtures.ts
```diff
@@ -30,7 +30,7 @@ export const test = base.extend<{
     let [background] = context.serviceWorkers()
     if (!background) background = await context.waitForEvent("serviceworker")
 
-    const extensionId = background.url().split("/")[2]
+    const extensionId = background.url().split("/")[2] ?? ""
```
kratico, line 33: `background.url().split("/")[2]!`

projects/extension/src/components/Accordion.tsx
```diff
@@ -115,7 +115,7 @@ const AccordionItem = ({
 }
 
 export const Accordion = ({
-  origin = "popup",
+  origin: _origin = "popup",
```
kratico, line 118: it would be better to delete if it's not used

projects/wallet-template/src/components/Accordion.tsx
```diff
@@ -115,7 +115,7 @@ const AccordionItem = ({
 }
 
 export const Accordion = ({
-  origin = "popup",
+  origin: _origin = "popup",
```
kratico, line 118: it would be better to delete if it's not used

kratico (approved):
Thanks 🚀 !!

PR #2046, opened 2024-04-01
feat(wallet-template): import private keys and mnemonics

ryanleecode:
Closes: https://github.com/paritytech/substrate-connect/issues/2040

Adds support for:

- Importing a private key. Note the private key must by a 64 byte expanded private key such as `d60fac497af78b5045b3a7785841b49809a850f027376a2aad3a9533a0cd71514994bddfdeafd5e636256d98ffafcf70698f518f652fb9a0f13af03510e8fab4`
- Import a mnemonic. This can be generated via subkey: `subkey generate`
- 
https://www.loom.com/share/f24bae5b0a8645069310239c0ae0675e?sid=92827df9-440a-413d-ad27-1e63b055dc3f


projects/wallet-template/src/background/keyring.ts
```diff
@@ -4,36 +4,150 @@ import {
...
+        return keystoreMeta.derivationPaths.map((d) => ({
+          ...d,
+          _type: "DerivationPath",
+        }))
+      case "PrivateKeyKeystore":
+        return [
+          {
+            _type: "Keypair",
+            publicKey: privateKeyToPublicKey(
+              keystoreMeta.privateKey,
```
kratico, line 51: `privateKey` should be encrypted in the secrets and not unencrypted in the meta

image

projects/wallet-template/src/background/types.ts
```diff
@@ -7,16 +7,20 @@ export type Account = {
   address: string
 }
 
-type DerivationPath = {
+export type KeysetAccount =
+  | ({ _type: "DerivationPath" } & DerivationPath)
+  | { _type: "Keypair"; publicKey: string }
```
kratico, line 12: use `type` instead of `_type`

`type` as a prop is not a reserved word

projects/wallet-template/src/background/keyring.ts
```diff
@@ -164,31 +232,70 @@ export const createKeyring = () => {
...
+            sign: (msg) => {
+              switch (keyset.scheme) {
+                case "Sr25519":
+                  return sr25519.sign(privateKey, msg)
+                case "Ed25519":
+                  return ed25519.sign(privateKey, msg)
+                case "Ecdsa":
+                  return ecdsa.sign(privateKey, msg)
+              }
+            },
```
kratico, line 293: This should be simplified so the `publicKey` and the `sign` use the same cryptographic `Curve`.

For example
```ts
import {
  sr25519,
  ed25519,
  ecdsa,
  KeyPair,
  Curve,
} from "@polkadot-labs/hdkd-helpers"

const curveFnMap = {
  Sr25519: sr25519,
  Ed25519: ed25519,
  Ecdsa: ecdsa,
} as Record<string, Curve>

const createKeyPair = (privateKey: string, scheme: string): KeyPair => {
  const curve = curveFnMap[scheme]
  if (!scheme) throw new Error("unsupported signature scheme")

  return {
    publicKey: curve.getPublicKey(privateKey),
    sign(message) {
      return curve.sign(message, privateKey)
    },
  }
}

File: projects/wallet-template/src/background/keystore/types.ts [hunk 4]
```diff
@@ -1,10 +1,28 @@
...
+}
+
+export type DerivationPathKeystore = BaseKeystore & {
+  _type: "DerivationPathKeystore"
+  derivationPaths: DerivationPath[]
+}
+
+export type PrivateKeyKeystore = BaseKeystore & {
+  _type: "PrivateKeyKeystore"
+  privateKey: string
```
kratico, line 22: see previous comment, the `privateKey` should be encrypted and removed from the keystore meta.

And the `publicKey` should be precomputed and stored in the keystore meta

projects/wallet-template/src/containers/WalletPopup/pages/Accounts/ImportAccounts.tsx
```diff
@@ -0,0 +1,349 @@
...
+  }
+
+  const onSubmit: SubmitHandler<FormFields> = async (data) => {
+    switch (activeTab) {
+      case "private": {
+        await rpc.client.insertKeyset({
+          _type: "PrivateKey",
+          name: data.keysetName,
+          scheme: data.scheme,
+          miniSecret: bytesToHex(entropyToMiniSecret(fromHex(data.key!))),
```
kratico, line 77: I believe this is not right.
The miniSecret can not be derived from the private key

projects/wallet-template/src/containers/WalletPopup/pages/Accounts/ImportAccounts.tsx
```diff
@@ -0,0 +1,349 @@
...
+  mnemonicInput?: string
+  key?: string
+  scheme: "Sr25519" | "Ed25519" | "Ecdsa"
+  networks: {
+    polkadot: boolean
+    westend: boolean
+    kusama: boolean
+  }
+  keysetName: string
+}
```
kratico, line 28: Improve, use 2 set of form fields so `mnemonic` and `key` are not optional

projects/wallet-template/src/containers/WalletPopup/pages/Accounts/ImportAccounts.tsx
```diff
@@ -0,0 +1,349 @@
...
+  mnemonicToEntropy,
+} from "@polkadot-labs/hdkd-helpers"
+import { fromHex, toHex } from "@polkadot-api/utils"
+import { bytesToHex } from "@noble/ciphers/utils"
+import { sr25519CreateDerive } from "@polkadot-labs/hdkd"
+import { networks } from "./networks"
+
+type FormFields = {
+  mnemonicInput?: string
+  key?: string
```
kratico, line 20: rename to `privateKey`

projects/wallet-template/src/containers/WalletPopup/pages/Accounts/ImportAccounts.tsx
```diff
@@ -0,0 +1,349 @@
...
+          privatekey: data.key!,
+          createdAt: Date.now(),
+        })
+        break
+      }
+      case "mnemonic": {
+        const entropy = mnemonicToEntropy(data.mnemonicInput!)
+        const miniSecret = entropyToMiniSecret(entropy)
+        // TODO: Add support for Ed25519 and Ecdsa
+        const derive = sr25519CreateDerive(miniSecret)
```
kratico, line 87: re-use the signature scheme dropdown to support the 3 signature schemes

projects/wallet-template/src/containers/WalletPopup/pages/Accounts/ImportAccounts.tsx
```diff
@@ -0,0 +1,349 @@
...
+  const validateMnemonic = (value: string | undefined) => {
+    if (!value) return "Mnemonic is required"
+
+    try {
+      mnemonicToEntropy(value)
+      return true
+    } catch (_) {
+      return "Invalid mnemonic"
+    }
+  }
```
kratico, line 149: `validateMnemonic` should be outside this component fn so it's not re-created on every render

projects/wallet-template/src/containers/WalletPopup/pages/Accounts/ImportAccounts.tsx
```diff
@@ -0,0 +1,349 @@
...
+          break
+        case "Sr25519":
+          sr25519.getPublicKey(bytes)
+          break
+      }
+      return true
+    } catch (_) {
+      return "Invalid private format"
+    }
+  }
```
kratico, line 138: `validatePrivateKey` should be outside this component fn so it's not re-created on every render

kratico (changes requested):

projects/wallet-template/src/containers/WalletPopup/pages/Accounts/ImportAccounts.tsx
```diff
@@ -0,0 +1,399 @@
...
+  Sr25519: sr25519CreateDerive,
+  Ed25519: ed25519CreateDerive,
+  Ecdsa: ecdsaCreateDerive,
+}
+
+type Scheme = "Sr25519" | "Ed25519" | "Ecdsa"
+
+type Tab =
+  | {
+      _type: "mnemonic"
```
ryanleecode, line 32: `type` is reserved in react hook form

kratico: However, react-hook-form should not dictate the API for an extension RPC endpoint.
Therefore, there could be a TS type for a react-hook-form and another type for the RPC endpoint.
Then, before submitting the form fields the `_type` could be transformed to `type`

projects/wallet-template/src/background/createBackgroundRpc.ts
```diff
@@ -171,11 +171,12 @@ export const createBackgroundRpc = (
     async createPassword([password]) {
       return keyring.setup(password)
     },
-    async insertKeyset([keyset, miniSecret]) {
-      const existingKeyset = await keyring.getKeyset(keyset.name)
+    async insertKeyset([args]) {
```
kratico, line 174: Anything that was previously named `keyset` needs a better name because now you can insert a keyset or keypair.
Methods like `insertKeyset`/`removeKeyset` need to be renamed.

ryanleecode: Going with "cryptographic key"

---

The umbrella term for "keyset" and "keypair" could be:

Cryptographic Key Components

  Both "keyset" and "keypair" are related to cryptographic keys used in secure communication and authentication 
  systems. A keyset is a collection of cryptographic keys, while a keypair consists of a public key and its 
  corresponding private key, which are used together in asymmetric cryptography.

kratico: Nice

`CryptoKey` could be shorterd

kratico (approved):

Issue #2078, opened 2024-04-11
Chainspec CI is broken


ryanleecode:
https://github.com/paritytech/substrate-connect/actions/runs/8643429253/job/23696459031

ryanleecode:
https://github.com/paritytech/substrate-connect/actions/runs/8648620056/job/23712876731

ryanleecode:
the commits arent even signed so i dont see why we have secret in the first place

https://github.com/paritytech/substrate-connect/pull/2080/commits

ryanleecode:
The token was created here: https://github.com/paritytech/substrate-connect/pull/939/files

PR #2079, opened 2024-04-11
fix: chainspec ci

ryanleecode:
Changes the github token from `GH_TOKEN` to `GITHUB_TOKEN` for deploying change specs. The old token expired and was likely somebody's personal access token. We need a org level solution with signed commits to do this, but for now changing it back to `GITHUB_TOKEN` is sufficient enough to get this working again.

closes: #2078


ryanleecode:
Changing the github token worked, so the old one is probably expired: https://github.com/paritytech/substrate-connect/actions/runs/8648620056/job/23712876731

kratico (approved):

PR #2116, opened 2024-04-16
feat(wallet-template): account details

ryanleecode:
closes: #2117

Keep it basic for now, but this lets us get the wallet address from the exntension without having to go into the dapp to copy it.

image


projects/wallet-template/src/containers/WalletPopup/pages/Accounts/AccountDetails.tsx
```diff
@@ -0,0 +1,67 @@
...
+    return null
+  }
+
+  const onCopyToClipboard = async () => {
+    setCopied(true)
+    setTimeout(() => {
+      setCopied(false)
+    }, 2000)
+    await copy(accountId)
+  }
```
kratico, line 22: the `setTimeout` needs to be removed when the component is unmounted.
If not, `setCopied` could be called in an unmounted component and there will be a React warning in the console

kratico:
Visual bug, letter case should not be transformed

image

ryanleecode:
Check this... https://zagjs.com/components/react/clipboard

way better

projects/wallet-template/src/containers/WalletPopup/pages/Accounts/AccountDetails.tsx
```diff
@@ -0,0 +1,60 @@
...
+      id: useId(),
+      value: accountId,
+    }),
+  )
+
+  const api = clipboard.connect(state, send, normalizeProps)
+
+  if (!accountId) {
+    return null
+  }
```
kratico, line 21: this seems to be redundant because the react-router will not match a route with an empty `accountId`

kratico (approved):

PR #2124, opened 2024-04-18
feat(wallet-template): list, add, remove chainspecs

ryanleecode:
UI+rpc method to list/add/remove chainspecs to the wallet. Its in the options page for now but this is with in mind the options page will be merged into wallet popup.

#2123

---

https://www.loom.com/share/25480f0011994bf6999085e972ddf862?sid=d07b8faa-82e8-4ab5-a36b-43bcf5cf816f


kratico (approved):
LGTM 🎉 

UX suggestion, when the chains are listed, I think it may be better to show parachains under their relay chain.

PR #2133, opened 2024-04-22
feat(wallet-template): cascade delete parachains

ryanleecode:
When you delete a relaychain from the allowlist chainspec UI. It should also delete their associated parachains.

### DIFF
--- a/projects/wallet-template/src/background/rpc/chainspec.ts
+++ b/projects/wallet-template/src/background/rpc/chainspec.ts
@@ -2,7 +2,10 @@ import { RpcMethodHandlers } from "@substrate/light-client-extension-helpers/uti
 import { z } from "zod"
 
 import { BackgroundRpcSpec } from "../types"
-import { wellKnownGenesisHashByChainId } from "../../constants"
+import {
+  wellKnownChainIdByGenesisHash,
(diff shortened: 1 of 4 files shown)


projects/wallet-template/src/background/rpc/chainspec.ts
```diff
@@ -65,5 +68,18 @@ export const removeChainSpecHandler: RpcMethodHandlers<
...
+  const chains = await lightClientPageHelper.getChains()
+  const parachains = chains
+    .filter((chain) => chain.relayChainGenesisHash === genesisHash)
+    .map((chain) => chain.genesisHash)
+
+  await Promise.all(
+    [...parachains, genesisHash].map((hash) =>
+      lightClientPageHelper.deleteChain(hash),
+    ),
+  )
```
kratico, line 84: I think this is covered in https://github.com/paritytech/substrate-connect/blob/fa0692522d4d06dba089b0907c278cde66b49fd1/packages/light-client-extension-helpers/src/background/background-helper.ts#L140-L161

ryanleecode: Oops good catch! Less work for the wallet :smiley:

PR #2137, opened 2024-04-24
feat(lc-dapp): disconnect provider when chain id is changed

ryanleecode:
Updates the LC Dapp to disconnect the provider when the chain is changed. This is to ensure we don't show the accounts from the old chain.


marshacb (approved):
LGTM!

Issue #1786, opened 2024-01-23
The extension gets "stuck" in the past


josepot:
### Issue Description

After a period of inactivity, then the extension stops announcing new blocks and nothing seems to work until you manually reset it.

![image](https://github.com/paritytech/substrate-connect/assets/8620144/f9ed1085-2649-4d16-b676-f1550214cc19)

### Steps to reproduce the issue

Steps to reproduce the issue
1. Use the extension in a DApp
2. Put your laptop to sleep for a few minutes
3. Try using the extension again

### Describe the results you received

The extension stops working, no new blocks get announced and the DApps can't communicate with smoldot

### Describe the results you expected

I expected for that not to happen

### Substrate-Connect version

extension: 0.3.1 npm-package: 0.8.5

### Provider

Chrome

### Browser version

As I laready mentioned before: 0.3.1

### Additional environment details

(This issue template is a huge PITA)

### Additional information

Additional information like issue happens only occasionally or issue happens with a particular architecture or on a particular setting

ryanleecode:
im able to further reproduce with the following ..

1. load the demo app
2. connect to a vpn (breaks ur current internet connect)
3. reload the demo app

the app will nuke with smoldot panicing

```
{code: -32603, message: 'Error: panicked at /__w/smoldot/smoldot/light-base…5:35:\ncalled `Option::unwrap()` on a `None` value'}
code
: 
-32603
message
: 
"Error: panicked at /__w/smoldot/smoldot/light-base/src/runtime_service.rs:1845:35:\ncalled `Option::unwrap()` on a `None` value"
[[Prototype]]
: 
Object
```

ryanleecode:
marking as blocked for now

https://github.com/smol-dot/smoldot/issues/1712

ryanleecode:
I haven't crashed or froze since the latest update. Will close this for now. Plz re-open if it happens again.

ryanleecode:
nope still goes to sleep. reopened

josepot:
According to the comment on #2151 :

> This doesn't fix https://github.com/paritytech/substrate-connect/issues/1786 but reduces the likelihood of the extension getting stuck in the past

it seems that this issue was closed by accident.

ryanleecode:
> According to the comment on #2151 :
> 
> > This doesn't fix https://github.com/paritytech/substrate-connect/issues/1786 but reduces the likelihood of the extension getting stuck in the past
> 
> it seems that this issue was closed by accident.

[...]

PR #2157, opened 2024-04-29
feat(extension): supervised smoldot

ryanleecode:
closes: #1786 

Adds a supervisor to detect when smoldot crashes/unresponsive by monitoring each chain that is added and polling it by getting the genesis hash. When a crash is detected, smoldot is restarted and this means you won't need to turn the extension on and off again.

However, the webpage/dapp will most likely require a refresh.

**In terms of functionality, when you leave a dapp the chains it is connected to no longer are "unconnnected". they are watched forever until a smoldot restart happens.** This is the compromise made to get this to work, but I think having it like this is better.

--- 

Avoiding any type of chainhead follow polling due to (somehow) getting chainhead follow subscription limit reached

image

---

Demo of smoldot crashing and it being restarted in the extension.

image

https://www.loom.com/share/59ecf161491f4a5884b0551be7a36698?t=111&sid=0d9e68c1-9833-47ce-848d-1de004cc87ba

### DIFF
--- a/packages/light-client-extension-helpers/src/background/background-helper.ts
+++ b/packages/light-client-extension-helpers/src/background/background-helper.ts
@@ -212,12 +212,14 @@ export const register = ({
           bootNodes:
             (await storage.get({ type: "bootNodes", genesisHash })) ??
             (JSON.parse(chain.chainSpec).bootNodes as string[]),
-          provider: await createSmoldotProvider({
-            smoldotClient,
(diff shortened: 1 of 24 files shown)


ryanleecode:
@jsdw tagging you for a review here since this issue is a bit trickey

ryanleecode:
superseding this with #2161 and #2162 and more to come...., since this approach isnt really working all that well

PR #2162, opened 2024-05-04
fix: recover from smoldot crashes

ryanleecode:
Fixes: #1786 (github will probably auto-close this issue but it shouldnt closed until the extension is updated on the chrome store).

Adds smoldot recovery to `light-client-extension-helpers`.

This change adds a wrapper to the smoldot client to add a `restart` method and change the `AddChainOptions` to accept `AddChainOptions` for potentialRelayChains instead of `Chain` objects, because `Chain` objects cannot persist between smoldot restarts.

This change also creates a custom JSON RPC Provider that will automatically re-connect to the chain it was connected to post-smoldot restart with the exact same `AddChainOptions`.

To actually perform restarts, there is now a `supervise` function that gets invoked during `register` that repeatedly calls `addChain` with an empty chain spec. The moment that there is a `CrashError` from smoldot, smoldot gets restarted.

## Testing

Follow the instructions in the repo README.md to run the extension and open the demo page. Then repeatedly disconnect and re-connect from VPN and watch the extension console for smoldot crashes.

## Demo

https://www.loom.com/share/c00c5348596d447a8ff371d043a3a818?sid=f3967a38-b725-460a-8c84-10975befbc99

---

<details>
  <summary>Old Description</summary>

This PR addresses [issue 1786](https://github.com/paritytech/substrate-connect/issues/1786), which prevents Substrate Connect from recovering when Smoldot crashes. The current fix attempt in [PR 2157](https://github.com/paritytech/substrate-connect/pull/2157) has issues with recovery, chain handle management, and observability.

This PR introduces a new library using [Effect](https://effect.website/docs/introduction) to manage Smoldot restarts and JSON RPC provider connections. It ensures that each chain's lifecycle is tied to a single Smoldot client. When a Smoldot client is killed, all chains automatically disconnect and reconnect to the new Smoldot client in the JSON RPC Provider implementation. This approach improves recovery and simplifies chain management.

However, this change requires additional feature work in future PRs before it can fully replace the background folder in the [light client extension helpers](https://github.com/paritytech/substrate-connect/tree/main/packages/light-client-extension-helpers/src/background). Future PRs will need to adapt the existing codebase to utilize the new Smoldot management and chain handling capabilities.

The key benefits are:

- **Improved recovery**: Substrate Connect can now properly handle Smoldot crashes without becoming unusable.
- **Better resource management**: The new library streamlines the creation and destruction of Smoldot clients and the addition and removal of chains.
- **Simplified chain handling**: Each chain is now directly linked to a single Smoldot client, reducing complexity.

**The most important file you'll want to be reviewing is the JSON RPC Provider implementation**: https://github.com/paritytech/substrate-connect/pull/2162/files#diff-00e4cc89f4df6dfb8719cbbe32fd288d8a4bb1e26a1932817814734d3bf556b1

---

## Use it yourself!

```sh
1. Clone this PR
2. corepack pnpm i
3. cd packages/light-client-experimental
4. corepack pnpm turbo build
5. pnpm dlx tsx ./scripts/testing.ts
```

If you want to have some fun you can also adjust the log levels in the testing script. 

image

image
</details>


packages/light-client-experimental/package.json
```diff
@@ -0,0 +1,43 @@
...
+    "dev": "pnpm build --watch",
+    "lint": "prettier --check README.md \"src/**/*.{js,jsx,ts,tsx,json,md}\""
+  },
+  "prettier": {
+    "printWidth": 80,
+    "semi": false,
+    "trailingComma": "all"
+  },
+  "devDependencies": {
+    "@effect/experimental": "^0.15.18",
```
jsdw, line 29: DQ (I'll have lots of these): Do we need this "experimental" package (always a word I'd rather avoid seeing :D), and what does it give over "effect"?

packages/light-client-experimental/package.json
```diff
@@ -0,0 +1,43 @@
...
+    "@polkadot-api/json-rpc-provider": "^0.0.1",
+    "@polkadot-api/json-rpc-provider-proxy": "^0.1.0",
+    "@polkadot-api/substrate-client": "^0.1.0",
+    "@substrate/connect-known-chains": "workspace:^",
+    "effect": "^3.1.2",
+    "effect-log": "^0.31.1",
+    "smoldot": "2.0.25"
+  },
+  "peerDependencies": {
+    "effect": "^3.1",
```
jsdw, line 40: DQ: Why is this a peer dependency?

packages/light-client-experimental/scripts/testing.ts
```diff
@@ -0,0 +1,116 @@
+import {
```
jsdw, line 1: If we want to commit this file as is, we should also make it clear how to actually run it (eg in the README or comment at the top of this file) ie the `pnpm dlx tsx ./scripts/testing.ts` step

packages/light-client-experimental/src/index.ts
jsdw, line 1: Should this be exposing some entrypoint to use this package?

packages/light-client-experimental/scripts/testing.ts
```diff
@@ -0,0 +1,116 @@
...
+  Cause,
+  Duration,
+  Effect,
+  Logger,
+  LogLevel,
+  pipe as $,
```
jsdw, line 7: Is this an Effect convention? Offhand I found $ confusing when looking below because it's been so overloaded in the past in JS libs!

packages/light-client-experimental/src/smoldot/data.ts
jsdw, line 1: This file just contains errors by the looks of it so maybe call it `errors.ts` for clairty?

packages/light-client-experimental/src/smoldot/resources/chain.ts
```diff
@@ -0,0 +1,31 @@
...
+import * as smoldot from "smoldot"
+
+import { Cause, Effect, pipe as $ } from "effect"
+import { AddChainError, AlreadyDestroyedError, CrashError } from "../data"
+
+export type AcquireOptions = smoldot.AddChainOptions & {
+  client: bindings.client.Client
+}
+
+export const acquire = (
```
jsdw, line 11: I'd like to see more comments around this code, eg what is the intended purpose of these `acquire` and `release` functions?

packages/light-client-experimental/src/smoldot/resources/client.ts
```diff
@@ -0,0 +1,31 @@
...
+    Effect.andThen((runSync) =>
+      Effect.sync(() =>
+        smoldot.start({
+          ...options,
+          logCallback: (...args) => options.logCallback(...args).pipe(runSync),
+        }),
+      ),
+    ),
+    Effect.map(bindings.client.make),
+  )
```
jsdw, line 26: I wonder what the standard version of this would look like, because it can be a bit hard following when there are no intermediate variable names etc. I geuss something like this but not really sure:

```ts
const runtime = Effect.runtime<never>();
const res = runtime.runSync(() => {
    smoldot.start({ 
        ...options.
        logCallback: (...args) => Effect.runSync(options.logCallback(...args))
});
bindings.client.make(res)
```

In general I wonder whether we can reduce the amount of `Effect` related code and use it to tie up things where it provides a bunch of value while writing the "lower level" bits in more plain JS so that it's more readable; it's important that others can understand and review these things :)

jsdw (commented):
I find this very hard to review personally. I had a look into the `Effect` library and in theory I think it's a good idea, but it's also used very heavily throughout thie package, and I feel like it'd require a somewhat deeper dive into Effect to really understand it all for me.

Things that I think would help me:
- Many more comments explaining what you're trying to do in various places (especially where any less standard `Effect` things are used; people not proficient in the library need to be able to follow this code).
- Less `Effect` usage, ie write more basic TS anywhere that doesn't gain so much value from `Effect` and then wrap it into `Effect` stuff when combining it or whatever. (Not sure how possible this is or not)
- Example usage (though I appreciate there is one such usage in the testing file)
- A proper code walkthrough to help talk through things (this might also let me give better feedback on the above things too) :)

I'd suggest we schedule a chat for a code review with you, myself and at least one of the other TS engs as a next step and that should help a bunch!

packages/light-client-experimental/scripts/testing.ts
ryanleecode: Nope, usually you'd do `some_effect.pipe(...whatever)`, but prettier make it look ugly as hell, so i prefer this

ryanleecode:
Nuked the experimental package and cut down on a lot of stuff. The new changes are now a part of the `light-client-extension-helpers` package and its now actually working inside of substrate connect.

Still need to add some docs/comments and do a little bit more cleanup, and it'll be ready for another review.

ryanleecode:
Also realized we don't even need a registry at all, just make the `potentialRelayChains` an array of `addChainOptions` instead of `Chain`

ryanleecode:
@jsdw @TarikGul @marshacb This is ready for another review :)

packages/light-client-extension-helpers/src/background/background-helper.ts
```diff
@@ -40,6 +41,8 @@ export const register = ({
   if (isRegistered) throw new Error("helper already registered")
   isRegistered = true
 
+  supervise(smoldotClient)
```
TarikGul, line 44: To get a better understanding, why doesn't `supervise` need an `await`?

packages/light-client-extension-helpers/src/background/json-rpc-provider.ts
```diff
@@ -0,0 +1,205 @@
...
+): Effect.Effect<JsonRpcProvider, Cause.UnknownException, never> => {
+  return Effect.gen(function* () {
+    const runtime = yield* Effect.runtime<never>()
+    const runCallback = Runtime.runCallback(runtime)
+    const runPromise = Runtime.runPromise(runtime)
+
+    /**
+     * This is a special single-element queue that signals when the daemon is
+     * ready. It is cleared every time the daemon restarts. While a `Deferred`
+     *  would typically be used in this situation, the ability of the daemon
```
TarikGul, line 46:
```suggestion
     * would typically be used in this situation, the ability of the daemon
```

packages/light-client-extension-helpers/src/background/json-rpc-provider.ts
```diff
@@ -0,0 +1,205 @@
...
+     * An error pubsub for the daemon to publish errors to.
+     * It is the responsibility of the JSON RPC provider to consume these
+     *  errors and propagate them with the onError callback.
+     */
+    const errorPubSub = yield* PubSub.sliding<void>(1)
+
+    /**
+     * A synchronized reference to the current chain. This reference is updated
+     * every time the daemon restarts.
+     *
```
TarikGul, line 68:
```suggestion
```

packages/light-client-extension-helpers/src/background/background-helper.ts
ryanleecode: register isnt marked as async (maybe it should be)? so it is indeed just a dangling promise here

ryanleecode:
Theres still the issue of chainhead subscriptions being reached because we are going over 2. i think this can be fixed in a subsequent PR. And we won't do another release of the extension until thats fixed.

packages/light-client-extension-helpers/src/background/json-rpc-provider.ts
```diff
@@ -0,0 +1,205 @@
...
+     * A message pubsub for the daemon to publish JSON RPC messages to.
+     * It is the responsibility of the JSON RPC provider to consume these
+     * messages and propagate them with the onMessage callback.
+     */
+    const messagePubSub = yield* PubSub.unbounded<string>()
+
+    /**
+     * An error pubsub for the daemon to publish errors to.
+     * It is the responsibility of the JSON RPC provider to consume these
+     *  errors and propagate them with the onError callback.
```
marshacb, line 61:
```suggestion
     * errors and propagate them with the onError callback.
```

packages/light-client-extension-helpers/src/smoldot/tasks.ts
```diff
@@ -0,0 +1,68 @@
...
+ * will be restarted.
+ *
+ * @param client - The smoldot client instance to be supervised.
+ * @param options - Options for customizing the supervision behavior.
+ * @param options.repeatSchedule - The frequency at which to invoke `addChain`.
+ * Defaults to {@link DEFAULT_SUPERVISE_REPEAT_SCHEDULE}.
+ * @param options.retrySchedule - The frequency at which to attempt restarting
+ * the client if needed. Defaults to {@link DEFAULT_SUPERVISE_RETRY_SCHEDULE}.
+ */
+export const supervise = async (
```
jsdw, line 26: Nit: this file is called `tasks.ts` but the main export is this supervise stuff; maybe call it `supervise.ts` or similar?

packages/light-client-extension-helpers/src/smoldot/tasks.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    const stop = () =>
+      Fiber.interrupt(daemon)
+        .pipe(Effect.andThen(() => Effect.void))
+        .pipe(runPromise)
+
+    return {
+      stop,
+    }
+  }).pipe(Effect.runPromise)
```
jsdw, line 68: When does something actually get the `stop` function back? ie does `.pipe(Effect.runPromise)` just run the gen thing forever?

It feels like you could kick off the daemon and return the stop method synchronously (esp because the only user of this function is not async anyways), but I don't understand this `Effect` stuff well enough to know what it's actually doing!

packages/light-client-extension-helpers/src/background/json-rpc-provider.ts
```diff
@@ -0,0 +1,205 @@
...
+        Effect.tap(Effect.logDebug),
+        Effect.tap((message) => Queue.offer(messagePubSub, message)),
+      ).pipe(Effect.forever)
+    }).pipe(
+      // by scoping here, we ensure the finalizers are run before the next retry
+      Effect.scoped,
+      Effect.retry(Schedule.spaced("1 second")),
+      Effect.forkDaemon,
+      Effect.interruptible,
+    )
```
jsdw, line 117: I still find this effect stuff all very hard to parse; eg just here we have `pipe`, `tryPromise`, `tap`, `Queue.offer`, `scoped`, `forkDaemon` and `interruptible` which are all random effect related things that one needs to read up on to understand what's going on here. And in general, there are a _ton_ of new functions and things just in this file that need to be understood to follow the code, eg `addFInalizer`, `runCallback`, `runPromise`, the whole concept of `Effect.gen` and yielding effect things in a monadic sort of style...

I generally think that we shouldn't use these sorts of libraries so heavily without first getting agreement/buyin from the team, because others need to be willing to learn and udnerstand all of this new stuff to be able to actually review and commit to the code.

I'm not really sure what to suggest, because I don't want to push to rewrite everything. The best "easy" thing that I think could be done to help is to add comments describing exactly what's happening on top of all of these really Effect heavy blocks. If you're keen on Effect, then perhaps it's also worth doing an internal team presentation to highlight some of the reasons it's useful and what this sort of code does?

In general, please do keep in mind going forwards that whatever code you write needs to be readable and understandable by others in the team, else it is unmaintainable.

packages/light-client-extension-helpers/src/background/json-rpc-provider.ts
```diff
@@ -0,0 +1,205 @@
...
+              Effect.try(() => onMessage(message)),
+              Effect.catchAll(() => Effect.void),
+            )
+
+          const invokeOnErrorCallback = $(
+            Effect.try(() => onError()),
+            Effect.catchAll(() => Effect.void),
+          )
+
+          runCallback(
```
jsdw, line 139: why do we get hold of a runtime and call `runCallback` like this, instead of something like `Effect.run(..)` here? also below :)

packages/light-client-extension-helpers/src/background/json-rpc-provider.ts
```diff
@@ -0,0 +1,205 @@
...
+     * This daemon process listens for JSON RPC messages, publishing them to
+     * the message pubsub. It also detects errors and publishes those to the
+     * error pubsub.
+     *
+     * If an error occurs with `chain.nextJsonRpcResponse()`, the daemon
+     * restarts. This indicates that the chain is no longer functional.
+     * In response, the daemon reconnects to smoldot using the same
+     * `AddChainParameters` to obtain a new `Chain` object.
+     */
+    const daemon = yield* Effect.gen(function* () {
```
jsdw, line 85: I wonder if breaking a couple of bits into separate functions with suitable names would help eg

```ts
function streamMessagesToPubSub(client, options, messagePubSub): Fiber.RuntimeFiber<void, Cause.UnknownException> {

}
```

(I think that's what this does right? It just calls addChain and then streams `chain.nextJsonRpcResponse()` to a message queue, restarting the whole thing in case of error)

packages/light-client-extension-helpers/src/background/json-rpc-provider.ts
```diff
@@ -0,0 +1,205 @@
...
+          Effect.tap(() => Effect.log("JSON RPC provider initialized.")),
+          runPromise,
+        )
+
+        return (onMessage, onError) => {
+          const invokeOnMessageCallback = (message: string) =>
+            $(
+              Effect.try(() => onMessage(message)),
+              Effect.catchAll(() => Effect.void),
+            )
```
jsdw, line 132: This is basically just the following, right?

```ts
try {
    onMessage(message)
} catch {
    // Ignore error calling onMessage
}
```

packages/light-client-extension-helpers/src/background/json-rpc-provider.ts
```diff
@@ -0,0 +1,205 @@
...
+
+          const invokeOnErrorCallback = $(
+            Effect.try(() => onError()),
+            Effect.catchAll(() => Effect.void),
+          )
```
jsdw, line 137: We could also just have:

```ts
const callOnError = () => {
    try { onError() } catch {}
}
```

And then below, just call `callOnError`.

packages/light-client-extension-helpers/src/background/json-rpc-provider.ts
```diff
@@ -0,0 +1,205 @@
...
+            }).pipe(Effect.scoped, Effect.forkDaemon),
+          )
+
+          runCallback(
+            // Daemon process that waits for the first error and then invokes
+            // the onError callback.
+            PubSub.subscribe(errorPubSub).pipe(
+              Effect.andThen((errorMessages) => Queue.take(errorMessages)),
+              Effect.andThen(() => invokeOnErrorCallback),
+              Effect.scoped,
```
jsdw, line 164: What does `scoped` do here?

packages/light-client-extension-helpers/src/smoldot/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from "./client"
+export * from "./types"
+export * from "./tasks"
```
jsdw, line 3: I know that this pattern exists throughout the library, but IMO I think it'd be better practice to be explicit about the exports from a package rather than use * exports (just so the library interface is obvious just looking at this root file and not needing to traverse other stuff :)

packages/light-client-extension-helpers/src/background/json-rpc-provider.ts
ryanleecode: Sets the boundary for where the subscription will automatically be cleaned up / unsubscribed.

jsdw: Is there a more explicit way to do this? It's not very obvious to me :D

Otherwise, perhaps worth a comment?

packages/light-client-extension-helpers/src/smoldot/tasks.ts
ryanleecode: it would be on `const { stop } = await supervise(client)`, but you're right that it can be sychronous. updated it to use `Effect.runSync` instead. 

We don't actually use the stop method but I think its just good to have.

packages/light-client-extension-helpers/src/background/json-rpc-provider.ts
ryanleecode: I think this works and is more concise

```ts
return (_onMessage, _onError) => {
  const onMessage = (message: string) =>
    Effect.try({
      try: () => _onMessage(message),
      catch: () => {},
    })

  const onError = Effect.try({ try: _onError, catch: () => {} })
```

packages/light-client-extension-helpers/src/background/json-rpc-provider.ts
ryanleecode: Its to ensure everything runs in the same context. If the application blew up somewhere (similar to a rust panic), whatever is running inside of `runCallback` & `runPromise` would immediately be cancelled. Whereas if you just ran it with `Effect.runPromise`, `Effect.runCallback`, it runs in its own context. 

Additionally, it perserves whatever you chose to bake into the runtime. So if I chose to set a custom logger, the effects that run inside of the `runPromise`, `runCallback` extracted from the runtime would use that logger instead of the default Effect or Console logger.

packages/light-client-extension-helpers/src/background/json-rpc-provider.ts
ryanleecode: For the record I am trying to experiment with something super basic, but this doesn't work and I have no idea why.

```ts
export const make = async (client: Client, options: AddChainOptions) => {
  let chain = await client.addChain(options)

  const provider = getSyncProvider(async () => {
    try {
      chain.remove()
    } catch (_) {}
    chain = await client.addChain(options)

    return (onMessage, onError) => {
      let connected = true
      ;(async () => {
        while (connected) {
          try {
            const message = await chain.nextJsonRpcResponse()
            onMessage(message)
          } catch (_) {
            onError()
            break
          }
        }
      })()

      const send: JsonRpcConnection["send"] = (message) => {
        try {
          chain.sendJsonRpc(message)
        } catch (_) {}
      }

      const disconnect: JsonRpcConnection["disconnect"] = () => {
        connected = false
        try {
          chain.remove()
        } catch (_) {}
      }

      return {
        send,
        disconnect,
      }
    }
  })

  return provider
}
```

ryanleecode: Ok we can forget all that, since this just worked. I suppose it was just overengineered.

```ts
export const make = async (client: Client, options: AddChainOptions) => {
  let chain = await client.addChain(options)
  chain.remove()

  const provider = getSyncProvider(async () => {
    chain = await client.addChain(options)

    return (onMessage, onError) => {
      let connected = true
      ;(async () => {
        while (connected) {
          try {
            const message = await chain.nextJsonRpcResponse()
            onMessage(message)
          } catch (_) {
            connected = false
            onError()
            break
          }
        }
      })()

      const send: JsonRpcConnection["send"] = (message) => {
        try {
          chain.sendJsonRpc(message)
        } catch (_) {}
      }

      const disconnect: JsonRpcConnection["disconnect"] = () => {
        connected = false
        try {
          chain.remove()
        } catch (_) {}
      }

      return {
        send,
        disconnect,
      }
    }
  })

  return provider
}
```

packages/light-client-extension-helpers/src/smoldot/tasks.ts
ryanleecode: Yeah thats possible but I dislike the Java-esque style where 1 file is 1 class or 1 function. If i treat it as a module then this is a task module and it has a supervise function and the module can be extended in the future to have more tasks.

packages/light-client-extension-helpers/src/smoldot/index.ts
ryanleecode: Yeah but I find its really easy to forget to add a new type that should get re-exported and then it gets missed during code review because people don't really pay attention to that. 

For anything that should be kept 100% internal I like to use the `@internal` jsdoc tag and then with `stripeInternal` set `true` in your tsconfig, these definitions won't be included in the d.ts`

https://www.typescriptlang.org/tsconfig/#stripInternal

packages/light-client-extension-helpers/src/background/json-rpc-provider.ts
```diff
@@ -0,0 +1,82 @@
...
+  try {
+    chain.remove()
+  } catch (_) {}
+}
+
+/**
+ * Creates a new JSON RPC provider using a client and options.
+ *
+ * This provider will automatically re-connect to the chain if the
+ * connection is lost.
```
jsdw, line 29: Ah so actually `getSyncProvider` is reconnecting behind the scenes by the looks of it already, so all this needs to do is use that and things mostly Just Work, right?

This looks way cleaner/simpler than the Effect stuff; good job simplifying!

packages/light-client-extension-helpers/src/background/json-rpc-provider.ts
```diff
@@ -0,0 +1,82 @@
...
+    const chain = await client.addChain(options)
+
+    return (onMessage, onError) => {
+      let connected = true
+      ;(async () => {
+        while (connected) {
+          try {
+            const message = await chain.nextJsonRpcResponse()
+            onMessage(message)
+          } catch (_) {
```
jsdw, line 51: Nit (and below?) :)

```suggestion
          } catch {
```

packages/light-client-extension-helpers/src/smoldot/tasks.ts
```diff
@@ -0,0 +1,68 @@
+import { Client, AddChainError } from "."
```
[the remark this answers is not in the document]
jsdw: Re exporting the Schedule stuff in the API?

Personally I'd prob just ask for times as ms `number`s because that's pretty standard(eg for setTimeout, setInternal etc) and then keep Schedule 100% an implementation detail :)

packages/light-client-extension-helpers/src/smoldot/index.ts
jsdw: Yup totally true; we forget to export types in Rust a fair bit, but personally I much prefer to forget exporting a type than to accidentally export a type I didn't mean to; it's easier to go one way than the other!

TIL about `@internal`; that's a cool approach if used thoroughly :)

packages/light-client-extension-helpers/src/smoldot/tasks.ts
```diff
@@ -0,0 +1,68 @@
...
+ * will be restarted.
+ *
+ * @param client - The smoldot client instance to be supervised.
+ * @param options - Options for customizing the supervision behavior.
+ * @param options.repeatSchedule - The frequency at which to invoke `addChain`.
+ * Defaults to {@link DEFAULT_SUPERVISE_REPEAT_SCHEDULE}.
+ * @param options.retrySchedule - The frequency at which to attempt restarting
+ * the client if needed. Defaults to {@link DEFAULT_SUPERVISE_RETRY_SCHEDULE}.
+ */
+export const supervise = (
```
jsdw, line 26: FWIW I thought I'd see what this looked like without any `Effect` stuff too and came up with this:

```ts
export type SuperviseOptions2 = {
  repeatSchedule: number, 
  retrySchedule: number,
}

export const supervise2 = async (
  client: Client,
  { repeatSchedule, retrySchedule }: SuperviseOptions2
) => {
  let stopped = false

  async function throwIfError(): Promise<boolean> {
    return client
      .addChain({ chainSpec: "" })
      .then(() => true)
      .catch((err) => {
        if (err instanceof AddChainError) {
          return true
        }
        throw err
      })
  }

  ;(async () => {
    while (!stopped) {
      try {
        if (await throwIfError()) {
          // Things are ok; wait a little and loop
          await sleep(repeatSchedule)
        }
      } catch (err) {
        // Things aren't ok; wait a little and restart.
        console.error(err)
        await sleep(retrySchedule)
        await client.restart()
      }
    }
  })()

  return {
    stop: () => {
      stopped = true
    }
  }
}

function sleep(ms: number): Promise<void> {
  return new Promise((resolve) => {
    setTimeout(resolve, ms)
  })
}
```

ryanleecode: restart on the client needs to be supressed in try/catch but otherwise looks good :+1:

ryanleecode:
I think this is ready to ship. Made a slight tweak to the superivse method to accept an abort signal instead of returning a stop function.

ryanleecode:
Lets try to get 2 approvals on this so James + either one of @TarikGul / @marshacb

packages/light-client-extension-helpers/src/smoldot/index.ts
ryanleecode: Gonna opt for the YOLO approach and just use @internal.

But if you want to actually see what the publicly export types are u can run `corepack pnpm turbo build` then open the `dist` folder and each of the the subpath exports will have a `d.ts` file in it showing the public api.

image

packages/light-client-extension-helpers/src/smoldot/tasks.ts
```diff
@@ -0,0 +1,83 @@
...
+      }
+    }
+  })()
+
+  if (options.abortSignal) {
+    options.abortSignal.addEventListener("abort", () => {
+      stopped = true
+    })
+  }
+}
```
jsdw, line 77: Nit: could move this next to the other `options.abortSignal` code so it's clear that we either bail immediately if stopped or add event to listen for stop if not already :)

jsdw (approved):
LGTM from what I understand, and much more readable now :)

marshacb (approved):
Solid PR! LGTM!

PR #2168, opened 2024-05-09
Update readme with latest dev steps

jsdw:
The current steps didn't work, so @ryanleecode gave me some new ones which I stuck in the README. I also removed a little cruft. (We should probably give the rest of the README and the DEPLOY-RELEASE.md file a proper go over when possible) 

I tested these all locally and they seem to work!


ryanleecode (approved):
Perfect, thanks!

jsdw:
It looks like you updated the README too in the meantime (I prob missed that), so up to you @ryanleecode what you prefer; it's all good so long as the steps work :)

PR #2196, opened 2024-05-26
chore(wallet-template): add crypto key ui revamp

ryanleecode:
image


TarikGul (approved):
Lgtm, just one zombienet test failing. Not sure how involved it is with the code changes.

PR #2200, opened 2024-05-29
feat: substrate discovery packages

ryanleecode:


ryanleecode:
@jsdw I took what you posted in element and made a few tweaks. this should be good. 

instead of opting for a unique `getExtensions` call for each package i chose just to export the filter method, since that way you won't be invoking the base `getExtensions` method mutliple times.

packages/discovery/src/index.ts
```diff
@@ -0,0 +1,56 @@
+export type ProviderDetail = {
+  _tag: string
```
jsdw, line 2: Is there a good reason for `_tag` (ie the underscore prefix) and not `tag` or `kind`? (Personally I prefer kind because this thing literally designates what _kind_ of interface it is)

packages/discovery/src/index.ts
```diff
@@ -0,0 +1,56 @@
...
+  info: ProviderInfo
+  provider: Promise<unknown>
```
jsdw, line 4: Does this have to be a promise? I would have thought that, by the time the extension responds to the event, it already can provide a well defined interface and doesn't need to give a Promise back?

packages/discovery/src/index.ts
```diff
@@ -0,0 +1,56 @@
...
+      detail: {
+        onProvider(detail) {
+          providers.push(detail)
+        },
+      },
+    }),
+  )
+
+  // slice the array to prevent further "asynchronous" updates. Providers
+  // that did not respond synchronously will be dropped.
```
jsdw, line 35: THankyou for this comment :D

packages/discovery/README.md
```diff
@@ -0,0 +1 @@
+# @substrate/discovery
```
jsdw, line 1: Maybe populate this with something small to begin with (we can improve later) like:

This package implements the discovery protocol that browsers use to find compliant browser extensions. 

The main export is a function called `getProviders`. This function dispatches an event on the `window` object that compliant browser extensions (or similar) may respond to by providing back an interface of the correct shape. An array of all such interfaces that we get back will be given back to the caller of `getProviders`.

packages/smoldot-discovery/package.json
```diff
@@ -0,0 +1,30 @@
...
+  "version": "0.0.6",
+  "author": "Parity Technologies (https://github.com/paritytech)",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/paritytech/substrate-connect.git"
+  },
+  "license": "MIT",
+  "exports": {
+    "./api": "./dist/api/index.mjs",
+    "./smoldot": "./dist/smoldot/index.mjs"
```
jsdw, line 12: To check my understanding, will this return something that you call like so:

```ts
import { .. } from "@substrate/smoldot-discoery/api"
import { .. } from "@substrate/smoldot-discoery/smoldot"
```

I think that as it stands, you'd need to include both this package and "@substrate/discovery" to use this, and then do something like:

```ts
import { getProviders } from "@substrate/discovery"
import { isSmoldotExtension, SmoldotExtensionProviderDetail } from "@substrate/smoldot-discovery/api"

// Not actually certain if this ends up with the provider typing or not:
let smoldotProviders: SmoldotExtensionProviderDetail[] = getProviders().filter(isSmoldotExtension)
```

Personally I think, to make this as simple as possible to use, we'd want to hand something back like:

```ts
// Most dApps will just import this (and maybe the types):
import { getSmoldotProviders, SmoldotProvider } from "@substrate/smoldot-discovery"

let smoldotProviders: SmoldotProvider[] = getSmoldotProviders()
```

And then I geuss extensions would just want to import the types to use so might do something like:

```ts
// Extensions might import just the types to make sure they expose a compliant interface:
import { SmoldotProvider, WellKnownChain .. } from "@substrate/smoldot-discovery/types"
```

packages/discovery/src/index.ts
ryanleecode: I think so because there is no top level await.

Currently we create the provider by invoking `getLightClientProvider` which returns a Promise. And this promise would resolve well before `requestProvider` is invoked, but theres no way to unwrap that promise since the callback is synchronous. 

https://github.com/paritytech/substrate-connect/blob/main/projects/wallet-template/src/inpage/index.ts#L43

packages/discovery/src/index.ts
ryanleecode: Sure lets go with `kind`

packages/discovery/src/index.ts
ryanleecode: Alternatively the extension could create a wrapper and for each method do `providerPromise.then(provider) => provider.invokeInterfaceMethod`, assuming all the methods are async.

jsdw: Mmm that was my thinking! I don't know why that thing is async offhand, and fundamentally I don't really see why the extension wouldn't be able to synchronously provide an interface back (even if it is an interface containing async methods).

So I guess I think that, all else aside, we shouldn't _force_ the interface to be `Promise<..>` and instead let the specific things decide what they return

zombienet-tests/src/0002-parachain.zndsl
```diff
@@ -6,4 +6,4 @@ alice: is up
 bob: is up
 alice: parachain 100 is registered within 225 seconds
 alice: parachain 100 block height is at least 3 within 200 seconds
-bob: ts-script ./0002-checkSync.ts within 140 seconds
```
ryanleecode, line 9: not sure why i had to change this but it _seems_ to just work.

Times out randomly:
https://github.com/paritytech/substrate-connect/actions/runs/9306397328/job/25615572039?pr=2200

ryanleecode:
Ok this should ready for another review

packages/discovery/README.md
```diff
@@ -0,0 +1,61 @@
...
+      return rpc.client.getAccounts(chainId)
+    },
+    async createTx(chainId: string, from: string, callData: string) {
+      return rpc.client.createTx(chainId, from, callData)
+    },
+  }),
+)
+
+window.addEventListener(
+  "unstableWallet:requestProvider",
```
marshacb, line 52: I see we now have the `substrateDiscovery:requestProvider` event. Is the `unstableWallet:requestProvider` still valid?

packages/discovery/README.md
```diff
@@ -0,0 +1,61 @@
...
+  ({ detail: { onProvider } }) => onProvider(detail),
+)
+
+window.dispatchEvent(
+  new CustomEvent("unstableWallet:announceProvider", {
```
marshacb, line 57: same question as above

packages/discovery/README.md
ryanleecode: nope just a copy and paste error on my part :)

ryanleecode:
I've also deleted the old docs since they aren't necessary anymore:

https://github.com/paritytech/substrate-connect/blob/main/docs/multi-injected-wallet-provider-discovery.md

zombienet-tests/src/0002-parachain.zndsl
ryanleecode: https://github.com/paritytech/substrate-connect/pull/2212

marshacb (approved):
Looks good to me!

PR #2222, opened 2024-06-06
fix: discovery protocol - providers not being async

ryanleecode:
While working on #2221, I discovered its actually infeasible for the provider in `ProviderDetail`l to be a not declared as a promise.

```diff
export type ProviderDetail = {
  kind: string
  info: ProviderInfo
-  provider: unknown
+  provider: Promise<unknown>
}
```

This is because the provider is created in the inpage/content-script of the extension which runs at the same time when the dapp loads. So if your only able to create the provider asynchronously, you're out of luck because discovery protocol function `getProviders` will run when the page loads, leading to a race condition.

Hence, I'm changing this back to `Promise<unknown>`, which is what we orignally had before. 

## Example

In the wallet-template we invoke `getLightClientProvider` to create the provider which always returns a promise.

We could wait for that promise to resolve and register the provider 

```ts
getLightClientProvider(DOM_ELEMENT_ID).then((provider) => {
  const detail: LightClientProviderDetail = Object.freeze({
    info: PROVIDER_INFO,
    provider,
  })

  window.addEventListener(
    "substrateDiscovery:requestProvider",
    ({ detail: { onProvider } }) => onProvider(detail),
  )
  
  window.dispatchEvent(
    new CustomEvent("substrateDiscovery:announceProvider", {
      detail,
    }),
  )
})
```

But this doesn't work because its running at the same time, the dapp is calling `getProviders`. Meaning the eventListener may not be registered __BEFORE__ getProviders is called.

### DIFF
--- a/packages/connect-discovery/src/index.ts
+++ b/packages/connect-discovery/src/index.ts
@@ -65,7 +65,7 @@ export namespace Unstable {
   export type SubstrateConnectProviderDetail = {
     kind: typeof Kind
     info: ProviderInfo
-    provider: Provider
+    provider: Promise<Provider>
   }
 
   export const isSubstrateConnectExtension = (
--- a/packages/discovery/src/index.ts
+++ b/packages/discovery/src/index.ts
@@ -1,7 +1,7 @@
 export type ProviderDetail = {
   kind: string
   info: ProviderInfo
-  provider: unknown
+  provider: Promise<unknown>
 }
 
 export type OnProvider = {
--- a/packages/discovery/src/index.ts
+++ b/packages/discovery/src/index.ts
@@ -1,7 +1,7 @@
 export type ProviderDetail = {
   kind: string
   info: ProviderInfo
-  provider: Promise<unknown>
+  provider: unknown
 }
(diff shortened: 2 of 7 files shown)


marshacb (approved):

jsdw:
> This is because the provider is created in the inpage/content-script of the extension which runs at the same time when the dapp loads. So if your only able to create the provider asynchronously, you're out of luck because discovery protocol function getProviders will run when the page loads, leading to a race condition.

So you're saying that if you need to create a provider async, then you're screwed because `getProviders` might run before you've had a chance to?

But if you wanted, you could immediately return something to the `onProvider` which is like `{ async addChain() { ... }, async addWellKnownChain() { ... } }` or whatever, and those async fns can wait.

But all of that said: if it's likely that providers will need to do something async anyway before they can provide anything sensible, then it probably makes more sense to just return a `Promise<unknown>`. We could also be flexible and return just `unknown` in the general case, but then for eg `smoldot-v1` specify that we expect back a `Promise<SmoldotProvider>` or whatever

PR #2226, opened 2024-06-07
chore(wallet-template): connect networks page

ryanleecode:
Note the networks page is accessible even when the wallet is locked.

---

## NEW

image
image

## OLD

image
image


marshacb (approved):
lgtm!

PR #2247, opened 2024-06-20
chore(wallet-template): update playwright tests

ryanleecode:
Update playwright tests for wallet template.

https://www.loom.com/share/56300629ff0744ba89ba9a281fa06105?sid=3bbd8b72-3e28-45e5-8192-5791dfe828ac

### DIFF
--- a/projects/wallet-template/tests/fixtures.ts
+++ b/projects/wallet-template/tests/fixtures.ts
@@ -1,5 +1,4 @@
 import { test as base, chromium, type BrowserContext } from "@playwright/test"
-import path from "path"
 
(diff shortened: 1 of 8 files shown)


projects/wallet-template/tests/fixtures.ts
```diff
@@ -19,14 +18,6 @@ export const test = base.extend<{
     await context.close()
   },
   extensionId: async ({ context }, use) => {
-    /*
```
ryanleecode, line 22: manifest v2 is dead ... so this should be removed

marshacb (approved):

